PR #3, opened 2018-07-04
Adding Haralick Texture features

HugoFRN:
One example of the proposed features extraction.
I think the usage is self explanatory but please tell me if you think I need to change things in the way the algorithm is implemented or documented so I will adapt for the next ones.


AleksMat:
Hi Hugo,

thank you for your contribution. I checked the code, made a few code style changes and optimizations. I also added unit tests. Then I merged your branch to develop branch. 

You can see changes I made here: https://github.com/sentinel-hub/eo-learn/commit/537eeda7259b021627cd8a8ed7d4d8d958b49ac4
and current version: https://github.com/sentinel-hub/eo-learn/blob/develop/features/eolearn/features/compute_haralick.py
Let us know if you agree with the changes and if everything still works as it should.

At the moment this task is still very slow because of loops over all elements of array. I think there has to be a way speed things up running GLCM function on vectorized data. I will let you know if we find some improvement.

PR #12, opened 2018-09-20
Adding HOG, LBP and blob featires

HugoFRN:
I added Histogram of Gradient, Local Binary Pattern and blob features to eolearn.features.
I also wrote unittest for it.


devisperessutti:
Hi Hugo,

Thank you very much for your contributions. 

I have mainly refactored the `BlobTask` to create three sub-tasks to ease the initialisation. The remaining changes are minor fixes and docs updates.

Bests,

Devis

PR #13, opened 2018-09-27
Added GaussianProcess missing data imputation

jO-Osko:
Added GaussianProcess(Kriging) missing data imputation and corresponding unittests, will generalize it to enable arbitrary regression like imputation of missing values


features/eolearn/features/interpolation.py
```diff
@@ -263,6 +264,78 @@ def __init__(self, feature, **kwargs):
         super().__init__(feature, interpolate.Akima1DInterpolator, **kwargs)
 
 
+class PixelWiseInterpolationTask(InterpolationTask):
+
```
azupanc: Missing documentation about PixelWiseInterpolationTask. Please add it.

features/eolearn/features/interpolation.py
```diff
@@ -263,6 +264,78 @@ def __init__(self, feature, **kwargs):
...
+        :rtype: numpy.ndarray
+        """
+        # get size of 2d array t x nobs
+        _, nobs = data.shape
+
+        # initialise array of interpolated values
+        new_data = data if self.resample_range is None else np.full((len(resampled_times), nobs),
+                                                                    np.nan, dtype=data.dtype)
+
+        # Interpolate for each pixel, could be easily parallelized
```
azupanc: If this is easily parallelized, could it be done?

azupanc:
This comment is related to connection between `PixelWiseInterpolationTask` (added in this PR) and `InterpolationTask` (already exists). 

To my understanding both tasks aim to do the same thing: replace missing values of a feature (or resample to fixed set of dates) in a time series of a single pixel. This means that interpolated values of a feature for a pixel depend solely on values of that feature and pixel. Values from other features or other pixels have no impact on interpolation object for that feature/pixel.

The above description holds for `PixelWiseInterpolationTask` but not for `InterpolationTask`. In `PixelWiseInterpolationTask` a new interpolation object is created/fitted for each pixel-feature. In `InterpolationTask` a single interpolation object is created/fitted for all pixel-features. Some measures are however taken in order to minimize the influence of other pixel-features on interpolation of a given pixel-feature. It was checked that these _inter_ pixel-feature influence is 0 for linear interpolation and negligible for cubic. A single interpolation object is created/fitted for all pixel-features in order to speed-up the execution time.

Before accepting this PR I would therefore like to make sure that we don't have two tasks that do almost the same thing. @jO-Osko could you please confirm that the possibility of other pixel-features influencing the interpolation of a given pixel with `GaussianProcessRegressor` made you implementing `PixelWiseInterpolationTask`? If answer is yes, could you share you're method of parallelization (speed-up) the task? Perhaps this could also be used in `InterpolationTask` and hence remove the need to have both.

jO-Osko:
Your understanding is correct. Both tasks aim to do the same thing with `PixelWiseInterpolationTask` imputing missing data for each pixel separately. The time overhead is quite high (construction of new imputation object for each pixel).

In my understanding, using single interpolation object for a whole range of pixels works because linear and cubic interpolation are relatively local (linear interpolation is fully local, while cubic, as you have noticed, incurs only a negligible error), while most regression methods use dependencies among all data points (in multi-pixel case, among all pixels), which is not desirable (to be fair, I have not tested direct influence in case of `GaussianProcessRegressor`). 

Another problem is the use of such models. `GaussianProcessRegressor` for example builds covariance matrix of all input data points, which can be really huge (of order `N^2` in the number of pixels), which almost always triggers instantaneous `MemoryError`, Splitting dataset into smaller patches *might* work if inter-pixel dependencies do not cause severe errors. While splitting data into smaller patches might work in this case, `PixelWiseInterpolationTask` seems really necessary when considering other types of data imputation, specifically using doubly logistic curve (which is currently being implemented), which has to process each pixel independently.

Regarding speed and parallelization issue since each model is trained independently of others, one way is to compute those interpolations in parallel. Due to GIL, currently, only `multiprocessing` library seems viable e.g. something like`multiprocessing.pool.Pool.map`, but I haven't tested the performance improvements yet.

azupanc:
Thanks for clarifying it. Since both tasks aim to do the same thing -- the only difference is whether or not single interpolation object trick can be applied or not -- I'm in favor of _combining_ these two tasks to avoid confusion. Perhaps by having an argument that selects between the two. What do you think?

In anyway, I would be very much interested to see what performance improvements you would get my using `multiprocessing` library.

jO-Osko:
This seems reasonable, I will add an argument to select between two implementations in eo task and additionally tested the usability of multiprocessing library.

azupanc:
Thanks for making the requested change. Could you please take a look at the failed travis build and fix it? Then the PR is ready to be merged.

PR #16, opened 2018-10-04
Interpolation update

mlubej:
Changes introduced:

- One can now provide a list of custom dates to the interpolation
- Updated unit tests for interpolation (no mask used before)
- Created a real-life example of a test patch, all existing tests should now be updated to use it, potentially it can be updated if there is some missing info
- Fixed some bugs in the interpolation if two dates are too close together


mlubej:
I'll clean this up and rebase on a clean develop

PR #21, opened 2018-10-16
Add class S1IWWCSInput() and class S1IWWMSInput()

wouellette:
It might be premature to add it now considering no examples based on S1 GRD data are provided on readthedocs, but it still took me a little too long to figure out how to query S1 IW data.

I don't think it hurts to have it available already for testing purposes.


batic:
Thanks, @wouellette, appreciated. Keep the pull requests coming :)

PR #29, opened 2018-11-06
Bug fix in sentinelhub_service

eerzin:


AleksMat:
Thanks for fixing this. :+1:

PR #37, opened 2018-11-14
Erosion Task PR

mlubej:
The erosion part of the Sampling task was split into a separate Erosion Task. The unit test for this was also upgraded and updated to use the test EOPatch.


mlubej:
I also fixed the unit tests in commit 8f4b3b3. The numpy arrays should've been copied prior to changing.

PR #41, opened 2018-12-13
Squeeze first dimension of feature_data if len=1

wouellette:
Some eopatches fail the interpolation when their first dimension is of length 1, because they no longer match the dimensions of the mask. This happens when the interpolated eopatch contains masks which do not match the FEATURES or BANDS in first dimension (time) length.

In the following example, this patch could not be interpolated before the script change:

`EOPatch(
  data: {
    FEATURES: <class 'numpy.ndarray'>, shape=(10, 1339, 1368, 9), dtype=float32
  }
  mask: {
    IS_VALID: <class 'numpy.ndarray'>, shape=(1, 1339, 1368, 1), dtype=bool
  }
  scalar: {}
  label: {}
  vector: {}
  data_timeless: {
    PRED: <class 'numpy.ndarray'>, shape=(1339, 1368, 1), dtype=uint8
  }
  mask_timeless: {}
  scalar_timeless: {}
  label_timeless: {}
  vector_timeless: {}
  meta_info: {}
  bbox: BBox(((486812.800062168, 2789900.2618496483), (500487.89285265905, 2803295.097495106)), crs=EPSG:32639)
  timestamp: <class 'list'>, length=10
)
`

### DIFF
--- a/features/eolearn/features/interpolation.py
+++ b/features/eolearn/features/interpolation.py
@@ -292,6 +292,8 @@ class InterpolationTask(EOTask):
         # Prepare mask of valid data
         if self.mask_feature:
             mask_type, mask_name = next(self.mask_feature(eopatch))
+            if feature_data.shape[0] == 1:
+                feature_data = np.squeeze(feature_data,axis=0)
             feature_data[~eopatch[mask_type][mask_name].squeeze(), :] = np.nan
 
         # Flatten array


wouellette:
The problem seems to be related to patches which have fewer IS_VALID pixels than the number of interpolated FEATURES. 

In my case, I interpolate at intervals of 20 days, giving me 10 images in the time domain, whereas the VALID_COUNT_SH indicates that a lot of pixel locations have less than 10 valid pixels.

Therefore, interpolating with fewer values than the output interpolated FEATURES seems difficult. 

This is typically the case in desert areas with really bright sand being detected as clouds (taking a look into your sentinelhub's `S2PixelCloudDetector` method now to mitigate this problem) consistently over many dates. This results in some areas having very few valid pixels (i.e. <10 over a period of 8 months) and other areas not covered in sand such as impervious man-made surfaces having full data availability (i.e. 32 out of 32 pixels in the specified time interval, because desert areas are not often cloudy).

How could the `interpolation` method handle those cases of extremely low valid pixel availability? 
In the meantime, I will see if I can fine tune the `S2PixelCloudDetector` to prevent false cloud detection.

AleksMat:
Hi William,

I checked the code and I agree there is an issue with dimensions when applying the mask. The code assumes that time dimensions of the feature, about to be interpolated, and mask will be the same. That is actually something we plan to add in the future - `EOPatch` will require that all time dependent features have the same size of time component. So it won't be possible to have a `FeatureType.DATA` feature of shape `(10, 1339, 1368, 9)` and a `FeatureType.MASK` feature of shape `(1, 1339, 1368, 1)` in the same patch. Instead such mask feature will have to be moved into `FeatureType.MASK_TIMELESS` as it is really time independent feature.
(Before we make such changes we will probably realese a version of `eo-learn` which will just show a `DeprecationWarning` if time dimensions won't match.)

About not enough valid pixel for interpolation - if there are not enough valid (non-masked) observations the interpolation will perform worse. In desert areas it is known that `s2cloudless` can make too many false positive detections. Therefore one option would be not to apply cloud mask on interpolation, instead you can just use a mask specifying where there is satellite data (we usually name it `IS_DATA` mask). Another option would be to produce a cloud mask from cloud probability mask with higher threshold than default and therefore reducing the number of false positives.

At the moment I'm adding some more checks when applying a mask on interpolated feature. I'll let you know more once I'm done.

AleksMat:
I implemented quite a few changes and created a new pull requests from this one: #43 (details are described there)
Therefore I am closing this PR.

PR #42, opened 2018-12-17
Enable ExportToTiff for FeatureType.DATA or FeatureType.MASK

wouellette:
Class `ExportToTiff` can now cater for both the extraction of `TIMELESS` and non-`TIMELESS` `FeatureType`.

An additional argument was added to specify the number of dates to export in the same fashion as one can specify the number of bands to export. 

For features which are both multi-date and multi-bands, it stacks the bands and the dates in order.
For instance, a feature with two `timestamp` and 3 bands would be stacked in the following order in the geotiff:
T1B1 T1B2 T1B3 T2B1 T2B2 T2B3


io/eolearn/io/local_io.py
```diff
@@ -39,18 +42,39 @@ def execute(self, eopatch, *, filename):
         feature_type, feature_name = next(self.feature(eopatch))
         array = eopatch[feature_type][feature_name]
 
-        if self.band_count == 1:
-            array = array[..., 0]
-
         dst_shape = array.shape
-        dst_transform = rasterio.transform.from_bounds(*eopatch.bbox, width=dst_shape[1], height=dst_shape[0])
+        filename_list = []
+        print(self.date_count)
```
AleksMat: Please remove prints from the code. Instead we use `logging` (on debug level) throughout `eo-learn` package. I think it might be a good idea to use it somewhere in this task.

io/eolearn/io/local_io.py
```diff
@@ -39,18 +42,39 @@ def execute(self, eopatch, *, filename):
...
+
+        if str(feature_type) in ['FeatureType.DATA','FeatureType.MASK']:
```
AleksMat: Instead of this you can simply check:
```Python
if feature_type in [FeatureType.DATA, FeatureType.MASK]:
```
(Note that `FeatureType` is an `Enum` class therefore you don't have to cast it into string)

io/eolearn/io/local_io.py
```diff
@@ -39,18 +42,39 @@ def execute(self, eopatch, *, filename):
...
         # Write it out to a file.
         with rasterio.open(os.path.join(self.folder, filename), 'w', driver='GTiff',
-                           width=dst_shape[1], height=dst_shape[0],
-                           count=self.band_count, dtype=self.image_dtype, nodata=self.no_data_value,
+                           width=width, height=height,
+                           count=self.date_count*self.band_count,
+                           dtype=self.image_dtype, nodata=self.no_data_value,
                            transform=dst_transform, crs=dst_crs) as dst:
-            dst.write(array.astype(self.image_dtype), indexes=self.band_count)
+            for id in range(self.date_count*self.band_count):
```
AleksMat: All iterations above and this iteration could be handled better. Instead of splitting numpy array into `filename_list` you could do the following:
- Swap array's axis and reshape it into `(height, width, self.date_count * self.band_count)`.
- Rasterize it without this for loop. The code from before wouldn't change much:
```Python
with rasterio.open(os.path.join(self.folder, filename), 'w', driver='GTiff',
                   width=dst_shape[1], height=dst_shape[0],
                   count=self.date_count * self.band_count, dtype=self.image_dtype, nodata=self.no_data_value,
                   transform=dst_transform, crs=dst_crs) as dst:
    dst.write(array.astype(self.image_dtype), indexes=self.date_count * self.band_count)
```
That should also work faster.

io/eolearn/io/local_io.py
```diff
@@ -20,17 +20,20 @@ class ExportToTiff(SaveToDisk):
...
+    :param date_count: Number of dates to be added to tiff image
+    :type date_count: int
     :param image_dtype: Type of data to be saved into tiff image
     :type image_dtype: numpy.dtype
     :param no_data_value: Value of pixels of tiff image with no data in EOPatch
     :type no_data_value: int or float
     """
 
-    def __init__(self, feature, folder='.', *, band_count=1, image_dtype=np.uint8, no_data_value=0):
+    def __init__(self, feature, folder='.', *, band_count=1, date_count=1, image_dtype=np.uint8, no_data_value=0):
```
AleksMat: I have 2 suggestions about `date_count` parameter:
- Instead of word `date` in the name it might be better to use word `time` as timestamps are not only dates but also datetimes.
- Instead of just a count this parameter could maybe also accept a time interval? Then it could select all times which fall into given time interval. But then the name would have to change to something more general...

wouellette: I agree on this point, but then you should allow the setting of intervals for `band_count` as well. Currently, you are basically only able to retrieve the `n` first bands, whereas you should be able to retrieve the `n` last bands too, or any other single band or intervals of bands. So maybe an argument in form of an index like `[n]` , `[-n]` or `[n:n+3]` would make the most sense.

io/eolearn/io/local_io.py
wouellette: Would swapping array's axis and reshaping it into `(height, width, self.date_count * self.band_count)` guarantee that the `band` and `time` dimensions are ordered in the correct way (e.g. T1B1 T1B2 T1B3 T2B1 T2B2 T2B3) if `band_count` and `date_count ` are larger than 1?

wouellette:
I shall be using pylint before pull requests from now on. Sorry for the inconvenience.

io/eolearn/io/local_io.py
AleksMat: Yes, the `band_count` parameter could be improved as well. Maybe the following options could be accepted:
- A single integer `n` - it takes first `n` bands
- A tuple of 2 integers `(a, b)` - it takes bands `a, a+1, ..., b-1`
- A list of indices, e.g. `[0, 3, -4, -1]` - it takes bands with exactly these indices.
And something similar could be done with the time parameter. 

I hope I am not complicating too much. So far we were using this task only for exporting single channel timeless data and masks. But you never know when something more complicated will be required.

wouellette: I would say that for `time_count`, giving a `time_interval` makes most sense because users typically know roughly which date they need to retrieve, but likely not the exact time, unless they print the `patch.timestamp` attribute. 

For `band_count`, I would say all three options listed remain useful, as all combinations may be wanted. However, for the tuple option `(a, b)` I would extend the list to `a, a+1, ..., b-1, b` as one would expect both bounds to be inclusive.

mlubej:
Dear @wouellette,

I wanted to take a look at this PR as I am also interested in the functionality. If you are busy with other things I can take over.

Do you have any local unpushed commits for this PR so that I don't reinvent any code?

Cheers!

io/eolearn/io/local_io.py
```diff
@@ -18,39 +18,89 @@ class ExportToTiff(SaveToDisk):
...
         array = eopatch[feature_type][feature_name]
+        dates = eopatch.timestamp
+        bands = range(array.shape[-1])
 
-        if self.band_count == 1:
-            array = array[..., 0]
+        filename_list = []
+        
+        if type(self.band_count) is list: 
+            if [date for date in self.band_count if type(elem) != int]:
```
mlubej: `date` might not be the appropriate keyword in this case. 
Also sorry for not understanding, but where is `elem` defined?

io/eolearn/io/local_io.py
```diff
@@ -18,39 +18,89 @@ class ExportToTiff(SaveToDisk):
...
+                raise ValueError('Invalid format in {} list, expected integers'.format(self.band_count))
+            array_sub = array[...,np.array(self.band_count)-1]
```
mlubej: Does this mean that if a list of bands is provided, that the list has to be 1-based instead of 0-based? I know hat rasterio count is 1-based, but I find this a bit confusing from the `eo-learn` side. I would drop the `band_count` and the `date_count` and introduce `band_indices` and `date_indices`, which are more in line with `numpy` and the whole `eo-learn` workflow, so that one could specify either:
* lists of various sizes: (i.e. [0] or [3,1,10,-1])
* tuple of (start_index, end_index) (i.e. (0,2)

this reduces the number of different input cases and makes the code simpler.

io/eolearn/io/local_io.py
```diff
@@ -18,39 +18,89 @@ class ExportToTiff(SaveToDisk):
...
+        elif type(self.band_count) is tuple:
+            if tuple(map(type, self.band_count)) != (int, int):
+                raise ValueError('Invalid format in {} tuple, expected integers'.format(self.band_count))
+            array_sub = array[...,np.nonzero(np.where(bands >= self.band_count[0] and bands <= self.band_count[1],bands,0))]
```
mlubej: For me this doesn't work, I need to do the following:

```
bands = np.array(range(array.shape[-1]))
np.nonzero(np.where((bands >= self.band_count[0]) and (bands <= self.band_count[1]),bands,0))
```

io/eolearn/io/local_io.py
```diff
@@ -18,39 +18,89 @@ class ExportToTiff(SaveToDisk):
...
+        elif type(self.band_count) == int:
+            array_sub = array[...,self.band_count-1]
+        else:
+            raise ValueError('Invalid format in {}, expected int, tuple or list'.format(self.band_count))
+            
+        band_dim = len(array_sub.shape[-1])
 
-        dst_shape = array.shape
-        dst_transform = rasterio.transform.from_bounds(*eopatch.bbox, width=dst_shape[1], height=dst_shape[0])
+        if feature_type in [FeatureType.DATA, FeatureType.MASK]:
```
mlubej: I guess this could easily work as is also for `FeatureType.SCALAR` ?

io/eolearn/io/local_io.py
```diff
@@ -18,39 +18,89 @@ class ExportToTiff(SaveToDisk):
...
+                raise ValueError('Invalid format in {}, expected int, tuple or list'.format(self.time_count))
+                
+            time_dim = len(array_sub.shape[0])
+            width = array_sub.shape[2]
+            height = array_sub.shape[1]
+        else:
+            time_dim = 1
+            width = array_sub.shape[1]
+            height = array_sub.shape[0]
+
```
mlubej: Some of the comments from above also apply here I guess

wouellette:
Hey @mlubej,

Thanks for the feedback, much appreciated! Apologies for the quick commit (which I didn't test, my bad).
I think we're getting there though :)

mlubej:
I took the liberty to push to your branch directly and fix some stuff:

* there were quite a few bugs
* also some unused or mislabeled stuff
* I added unit tests
* fixed pylint issues (pylint was updated to 2.2.2 I think)

I would recommend the following for your workflow:
* use PyCharm, it automatically shows you fallacies in your code, unused variables, all kinds of stuff, very useful
* use pylint regularly, less fixes to do then in the end
* try to write unittests as you develop the code, or test the code some other way

I hope I don't sound rude... What you contributed is very welcome and very useful! However, at least in this case it looked like you were occasionally just commiting the code without checking if it works or even does what it should do :)

Just a few comments for a more efficient practice, hope you dont mind.

Keep up the good work! 

Please check if you see any other problems, I would also like to ping @AleksMat to give a second glance before we merge.

Cheers

mlubej:
I see there were some additional pylint errors in `io`, but they are not present when I check locally... Anything I can do about that?

io/eolearn/tests/test_local_io.py
```diff
@@ -0,0 +1,266 @@
...
+            raster = np.moveaxis(raster, -1, 0)
+
+            self.assertTrue(np.all(subset == raster))
+
+    def test_export2tiff_mask_tuple_datetime(self):
+        mask = np.arange(5*3*2*1).reshape(5, 3, 2, 1)
+
+        eop = EOPatch.load(self.PATCH_FILENAME)
+        eop.mask['mask'] = mask
+        dates = np.array(eop.timestamp)
```
mlubej: This is an error, will fix tomorrow, should not really use these timestamps as the newly added features are different... also in other cases

io/eolearn/tests/test_local_io.py
```diff
@@ -0,0 +1,266 @@
...
+
+            self.assertTrue(np.all(subset == raster))
+
+    def test_export2tiff_mask_tuple_string(self):
+        mask = np.arange(5*3*2*1).reshape(5, 3, 2, 1)
+
+        eop = EOPatch.load(self.PATCH_FILENAME)
+        eop.mask['mask'] = mask
+        dates = np.array(eop.timestamp)
+
```
mlubej: also here

io/eolearn/tests/test_local_io.py
```diff
@@ -0,0 +1,266 @@
...
+from eolearn.core import EOPatch, FeatureType
+from eolearn.io import ExportToTiff
+from sentinelhub.io_utils import read_data
+from sentinelhub.time_utils import datetime_to_iso, iso_to_datetime
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+class TestEOPatch(unittest.TestCase):
```
mlubej: add more unittests for cases where the task fails...

wouellette:
Once again, thanks for the advice and sorry for the carelessness, I just wanted to provide the logic of the functionality to save you work, but it might have created you more work with all the unittests and all...

What I suggest is to not waste your time on it for the time being. I'm in the process of revamping my workbench with your suggestions (and @AleksMat's), and I'll get it fixed once that's in place.

Cheers,
William

mlubej:
Waiting for the checks to finish, otherwise, I believe it's good to merge!

wouellette:
Hey,

I just signed the petition "European Parliament: Stop the
censorship-machinery! Save the Internet!" and wanted to see if you could
help by adding your name.

Our goal is to reach 6,000,000 signatures and we need more support. You can
read more and sign the petition here:

http://chng.it/vKQRL7DS5q

Thanks!
William

PR #43, opened 2018-12-17
Wouellette patch 2

AleksMat:
This branch is primarily a fix for masking data in `InterpolationTask` however it also has the following changes (which are connected to the problem):
- Binary masks in `eolearn.io` and `eolearn.mask` are now cast to `numpy.bool` dtype,
- `FeatureParser` has a new parameter `allowed_feature_types`,
- Implemented `eolearn.core.FeatureTypeSet` class with groups of similar feature types as constants.

`InterpolationTask` now supports masking with the following feature types:
- `FeatureType.MASK` (same as before),
- `FeatureType.MASK_TIMELESS - masks entire time series for corresponding pixels in spatial dimension,
- `FeatureType.LABEL` - masks entire images for corresponding times in time dimension.


wouellette:
Excellent thanks for the quick implementation.
I'll let you know soon if it performs as expected.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -148,7 +148,7 @@ def _add_data(self, eopatch, data):
         mask_feature_type, mask_feature_name = next(self.valid_data_mask_feature())
 
         max_value = self.image_format.get_expected_max_value()
-        valid_data = (valid_mask == max_value).astype(np.uint8).reshape(valid_mask.shape + (1,))
+        valid_data = (valid_mask == max_value).astype(np.bool).reshape(valid_mask.shape + (1,))
```
andrejburja: i would put this in two lines

PR #44, opened 2018-12-18
start_date/end_date matched to timestamp

wouellette:
If start_date and/or end_date defined for the resampling range are beyond the range of the eopatch.timestamp, match them to the first and/or last eopatch.timestamp. This will avoid empty arrays at the beginning and end of the resampled eopatch.

### DIFF
--- a/features/eolearn/features/interpolation.py
+++ b/features/eolearn/features/interpolation.py
@@ -307,8 +307,8 @@ class InterpolationTask(EOTask):
             raise ValueError('Invalid resample_range {}, expected tuple'.format(self.resample_range))
 
         if tuple(map(type, self.resample_range)) == (str, str, int):
-            start_date = parser.parse(self.resample_range[0])
-            end_date = parser.parse(self.resample_range[1])
+            start_date = parser.parse(self.resample_range[0]) if parser.parse(self.resample_range[0]) > timestamp[0] else timestamp[0]
+            end_date = parser.parse(self.resample_range[1]) if parser.parse(self.resample_range[1]) < timestamp[-1] else timestamp[-1]
             step = timedelta(days=self.resample_range[2])
             days = [start_date]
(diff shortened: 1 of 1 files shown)


mlubej:
Hmm, while I agree that the empty arrays are problematic, this way ensured that each date after the resampling was evaluated on the same day. It can happen that two eopatches in the same process have different start and end timestamps, so matching the edges of the resampling interval to the timestamp interval might introduce differences between different patches, because the resampled dates would be different and then the interpolation is evaluated at different times.

If you see a workaround, let me know.

wouellette:
The issue was already identified and discussed with @azupanc at [Linear Interpolation Problem in overlapping DataTake areas using WCS requests ](https://forum.sentinel-hub.com/t/linear-interpolation-problem-in-overlapping-datatake-areas-using-wcs-requests/698/8).

For now I am handling it externally by narrowing down the "overall" time interval to the eopatches of my AOI which have the latest start_date and the earliest end_date in the following way:

```
start_date_interp = parser.parse(start_date)
end_date_interp = parser.parse(end_date)

for eopatch in eopatches:
     if start_date_interp < eopatch.timestamp[0]:
         start_date_interp = eopatch.timestamp[0]
     if end_date_interp > eopatch.timestamp[-1]:
         end_date_interp = eopatch.timestamp[-1]

resample_range = (start_date_interp, end_date_interp, interp_interval)
```
I don't have any wonder solution to handle this at the eopatch level rather than at the AOI level, but I think of something I'll shout :)

mlubej:
Hi @wouellette!

Can I close this PR or did you plan to extend it to bypass the problem? Perhaps it's too specific to be included in the official code because you need the info of the entire AOI?

Cheers,
Matic

wouellette:
You can close it for now yes. If I come up with some idea i'll re-open with a suggestion.

Regards

PR #49, opened 2019-01-14
Add rename_feature definition

wouellette:
rename_feature was added alongside remove_feature and add_feature for quick use without needing to set up a workflow.

### DIFF
--- a/core/eolearn/core/eodata.py
+++ b/core/eolearn/core/eodata.py
@@ -255,6 +255,23 @@ class EOPatch:
         """
         self._check_if_dict(feature_type)
         self[feature_type][feature_name] = value
+    
+    def rename_feature(self,feature_type,feature_name,new_feature_name):
+        """Renames the feature ``feature_name`` from dictionary of ``feature_type``.
+
+        :param feature_type: Enum of the attribute we're about to modify
+        :type feature_type: FeatureType
+        :param feature_name: Name of the feature of the attribute
+        :type feature_name: str
+        :param new_feature_name : New Name of the feature of the attribute
+        :type feature_name: str
+        """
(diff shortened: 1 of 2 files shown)


core/eolearn/core/eodata.py
```diff
@@ -255,6 +255,23 @@ def add_feature(self, feature_type, feature_name, value):
         """
         self._check_if_dict(feature_type)
         self[feature_type][feature_name] = value
+    
+    def rename_feature(self,feature_type,feature_name,new_feature_name):
```
mlubej: It might be better to go with the feature parser here, so that we enforce features to be of the same form, sort of like this:

```
def rename_feature(self, feature): 
    feature_type, feature_name, new_feature_name = self._parse_features(feature, new_names=True)
```
The docstring is also shorter then

core/eolearn/core/eodata.py
```diff
@@ -255,6 +255,23 @@ def add_feature(self, feature_type, feature_name, value):
...
+        """Renames the feature ``feature_name`` to ``new_feature_name`` from dictionary of ``feature_type``.
+
+        :param feature_type: Enum of the attribute we're about to modify
+        :type feature_type: FeatureType
+        :param feature_name: Name of the feature of the attribute
+        :type feature_name: str
+        :param new_feature_name : New Name of the feature of the attribute
+        :type feature_name: str
+        """
+        LOGGER.debug("Renaming feature '%s' from attribute '%s'", feature_name, feature_type.value)
```
mlubej: I would put the debug message after the check is completed so that it really does that, and additionally would print and error message if the conditions are not satisfied

core/eolearn/core/eodata.py
mlubej: Additionally, you could add a check if the new feature name is the same as the old one and not waste CPU in this case.

PR #56, opened 2019-01-23
Eoexecutor update

mlubej:
Changing back to a simpler `concurrent.futures` instead of `multiprocessing`. Adding `tqdm` progress bar.


AleksMat:
This branch was also (ab)used to update CI settings, package setups and a lot of code style improvements.

PR #60, opened 2019-01-25
Time interval parsing

batic:
When reading time_interval from meta_info of the existing eo-patch, the end time was not parsed as it was when creating new eo-patch. Now the time_interval from the request is saved to meta_info without any parsing, and rather pass it to sentinelhub-py as is.


io/eolearn/io/sentinelhub_service.py
```diff
@@ -95,7 +94,7 @@ def _get_parameter(self, name, eopatch):
         if name == 'maxcc':
             return 1.0
         if name == 'time_difference':
-            return dt.timedelta(seconds=-1)
+            return dt.timedelta(0)
```
batic: This should be -1 second as before, sorry.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -117,8 +116,7 @@ def _prepare_request_data(self, eopatch, bbox, time_interval):
         return {
             'layer': self.layer,
             'bbox': bbox if bbox is not None else self._get_parameter('bbox', eopatch),
-            'time': datetime_to_iso(time_interval) if isinstance(time_interval, str) else [
-                datetime_to_iso(x) if not isinstance(x, str) else x for x in time_interval],
+            'time': time_interval,
```
mlubej: Can it happen that the time interval in the existing EOPatch is of the format (`str`, `str`)? Previously if the time interval was (`str`, `str`) it would get transformed to (`datetime`, `datetime`). If it already was the latter format, then it was just copied.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -158,12 +156,7 @@ def _add_meta_info(self, eopatch, request_params, service_type):
 
         for param, eoparam in zip(['time', 'time_difference', 'maxcc'], ['time_interval', 'time_difference', 'maxcc']):
             if eoparam not in eopatch.meta_info:
-                if param == 'time':
-                    value = request_params[param]
-                    eopatch.meta_info[eoparam] = value if isinstance(value, str) else [
-                        iso_to_datetime(parse_time(x)) if isinstance(x, str) else x for x in value]
-                else:
-                    eopatch.meta_info[eoparam] = request_params[param]
+                eopatch.meta_info[eoparam] = request_params[param]
```
mlubej: If I understand correctly, `request_params['time']` now always returns instances containing `datetime`?

io/eolearn/tests/test_io.py
```diff
@@ -257,8 +256,7 @@ def test_return_type(self):
     def test_time_interval(self):
         for task in self.task_cases:
             with self.subTest(msg='Test case {}'.format(task.name)):
-                self.assertEqual(task.eop.meta_info['time_interval'],
-                                 [iso_to_datetime(x) for x in ('2017-1-1', '2018-1-1')])
+                self.assertEqual(task.eop.meta_info['time_interval'], ('2017-1-1', '2018-1-1'))
```
mlubej: I don't understand, shouldn't `('2017-1-1', '2018-1-1')` in this test now be instances of `datetime`?

io/eolearn/io/sentinelhub_service.py
batic: time was before always (str, str), now it can be both

io/eolearn/io/sentinelhub_service.py
batic: No, it can return either str or datetime, whatever the user provided as input.

io/eolearn/tests/test_io.py
batic: It can be either.

mlubej:
Good to merge when checks finish.

PR #62, opened 2019-01-30
Added get_tasks method to workflow that returns an odrdered dictionar…

azupanc:
…y of task names (key) and task instances (value). Purpose of this method is to get an easier access to task instances from the workflow.


mlubej:
Ready to merge.

PR #66, opened 2019-02-05
Bug/self abuse and other stuff

mlubej:
Fix cases of self abuse where variable `X` was improperly used in non-init functions of the classes, such as `EOTask`s

```
# bad
if self.X is None:
    self.X = foo

# good
X = foo if self.X is None else self.X
```


mlubej:
Also added a unit test.

PR #69, opened 2019-02-07
Minor: Fix link to data in notebook

ilyakava:
Very nice notebook!

I noticed the data link was broken, and I believe this is what it means to be pointing to on [this page](http://rkg.gov.si/GERK/).

I don't know Slovenian, so could you comment on how this reference data was constructed? Are these ground truth labels drawn exclusively by human experts? Because if these labels were supplemented by some ML algorithm as I've seen is sometimes the case in datasets like these, I think it would be worth mentioning in this notebook the differences between the ML algo that created the data, vs the one being used here to learn from it.

Thanks,
Ilya


mlubej:
Hi @ilyakava!

Yet another PR that I completely missed, sorry again and thanks for your input, you're very good at spotting details!

To answer your question:
I don't know the details, but the Slovenian government runs a very precise system regarding parcel tracking. This dataset was **not** supplemented by any kind of ML algorithms, so it's done by the government and the people/farmers and probably also checked with some additional flights with planes, etc. So the ground truth in Slovenia is quite of high quality :)

mlubej:
I fixed the link and added more text to avoid some confusion in the future. Thanks again!

PR #70, opened 2019-02-07
Minor: Fix number of band features referenced in notebook comment

ilyakava:
Something I noticed while running this notebook and checking `features_train.shape`. 6 bands were selected with `custom_script = 'return [B02, B03, B04, B08, B11, B12];'` earlier, and 3 derived features were added (NDVI, NDWI, NORM). And then 23 is the number of captures over time per location as I understand, explaining 207 = 23 * 9.


ilyakava:
A general question I had @mlubej is that as I understand you are using a single pixel's features for its classification, and not looking in the neighborhood for more features. I'm more familiar with hyperspectral than multispectral imagery classification, but I know this is a popular way to improve classification performance. I'm also curious if you could comment on the choice of LightGBM compared to other standards in the multispectral imagery classification. It did give pretty spatially smooth results though with very little noise though!

This whole notebook is very impressive!

mlubej:
Hi @ilyakava!

Thanks for the nice compliments and your improvements, and sorry for the long reply, it seems that I may need to fix my notification settings...

The only thing I changed is to switch the base branch from `master` to `develop`.

Regarding your questions, since this is a simple example we wanted to use a simple approach and a simple pixel-based decision tree like classifier. If done properly, using the information from the neighbourhood would, of course, increase the accuracy, but sometimes it comes with large overhead in terms of CPU time.

What do you consider as other standards in multi-spectral imagery classification? We used it because it has a nice and intuitive interface in Python and it's easy to understand when you are still at the point of research (so you know which point of the pipeline to blame, if something goes wrong :) ). There does seem to be very little noise present (in terms of single pixel activation or salt&pepper effect), but this is probably due to the fact that the ML model capacity is relatively small so the probability of overfitting is low, I believe.

ilyakava:
@mlubej I recently got to try LGBM for hyper spectral data and compare it to the scikitlearn linear svm. In the majority of cases the SVM outperformed LGBM while also being faster, especially when there was very little training data.

It is interesting that you say the noise is low. The multiple measurements probably help for that!

PR #77, opened 2019-03-17
Introduce dask support for interpolation

wouellette:
This is more a suggestion than a formal pull request, as it naturally introduces an additional dependency to Dask.

The implementation works and shows significantly superior performance than the standard scipy implementation (from 90-100 secs/patch to about 40 secs/patch with 16 cpus, 64 gb ram on single machine).

I would assume you will make a Dask wrapper for certain eo-learn tasks which could benefit from it in a future release, but this one is particularly helpful for interpolating large numbers of patches with a dense temporal stack.

### DIFF
--- a/features/eolearn/features/interpolation.py
+++ b/features/eolearn/features/interpolation.py
@@ -1,6 +1,7 @@
 """ Module for interpolating, smoothing and re-sampling features in EOPatch """
 
 import numpy as np
+import dask.array as da
 
 from dateutil import parser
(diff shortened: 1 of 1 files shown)


wouellette:
Closing this pull request as I made new findings that are promising but introduce new dependencies, thus having an implication on certain classes (in this particular case on `InterpolationTask`), and it would need more structural changes to introduce them fully to eo-learn. I will take over this discussion to the Issues.

PR #78, opened 2019-03-20
Linear numpy interpolation

andrejburja:
For linear interpolation scipy function is replaced by numpy.interp().

### DIFF
--- a/features/eolearn/features/interpolation.py
+++ b/features/eolearn/features/interpolation.py
@@ -5,11 +5,11 @@ Module for interpolating, smoothing and re-sampling features in EOPatch
 import warnings
 import datetime as dt
 import inspect
+from functools import partial
 
 import dateutil
 import scipy.interpolate
 import numpy as np
-from functools import partial
 from sklearn.gaussian_process import GaussianProcessRegressor
 
 from eolearn.core import EOTask, EOPatch, FeatureType, FeatureTypeSet
@@ -295,8 +295,7 @@ class InterpolationTask(EOTask):
         """
         if str(inspect.getmodule(self.interpolation_object))[9:14] == 'numpy':
             return partial(self.interpolation_object, xp=times, fp=series, left=np.nan, right=np.nan)
-        else:
-            return self.interpolation_object(times, series, **self.interpolation_parameters)
+        return self.interpolation_object(times, series, **self.interpolation_parameters)
 
     def get_resampled_timestamp(self, timestamp):
         """ Takes a list of timestamps and generates new list of timestamps according to ``resample_range``
--- a/features/eolearn/features/interpolation.py
+++ b/features/eolearn/features/interpolation.py
@@ -235,7 +235,6 @@ class InterpolationTask(EOTask):
                 valid = ~np.isnan(data[:, obs])
 
                 obs_interpolating_func = self.get_interpolation_function(times[valid], data[valid, obs])
-                print(type(obs_interpolating_func))
 
                 new_data[:, obs] = obs_interpolating_func(resampled_times[:, np.newaxis])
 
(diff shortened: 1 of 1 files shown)


wouellette:
@andrejburja: Cool that you are updating this procedure! 
I have been working on improving the speed of interpolation, and I found an interesting implementation using [xarray_extras](https://xarray-extras.readthedocs.io/en/latest/index.html). It is only a BSpline implementation, but is dask-supported and goes seriously quick compared to the scipy implementations.

Now I wanted to ask how much of an improvement in terms of speed does the numpy implementation merged in this pull request provide? 

Regardless of that, I will share the implementation later today. It would be harder to integrate due to additional dependencies, and maybe the dask support should be done outside of eo-learn anyhow, but we can further discuss.

andrejburja:
Current scipy implementation on test eopatch: 2.18 s
Numpy implementation: 1.42 s
For the next step, I will try adding numba.
I have tried xarray implementation, but I noticed `interpolate_na` is very slow.

wouellette:
Using `xarray_extras` and dask support, `chunk_size=1000`, I divide processing time by 4-6 times on an eopatch of size: `(47, 1077, 899, 9)`

Scipy's `interp1d` took roughly **1 min** to process, while with `dask`+`xarray_extras` it is now around **15 sec** (with higher parallelization potential since the memory is handled much better than in the former method).

Below a potential solution roughly following what you did for numpy (with additional dependencies of course):

Dependencies:
```
import dask.array as da
from dask import compute
import xarray as xr
from xarray_extras.interpolate import splrep, splev
```

BSpline updated class:
```
class BSplineInterpolation(InterpolationTask):
    """
    Implements `eolearn.features.InterpolationTask` by using `scipy.interpolate.BSpline`
    """
    def __init__(self, feature, _library='xarray',  *, spline_degree=3, **kwargs):
        if _library == 'xarray':
            super().__init__(feature, xarray_extras.interpolate.splrep, k=spline_degree, **kwargs)
        else:
            super().__init__(feature, interpolate.make_interp_spline, k=spline_degree, **kwargs)

```

Adaptation in `interpolate_data` function for pixel-wise operation:

```
    def interpolate_data(self, data, times, resampled_times):
        """ Interpolates data feature

        :param data: Array in a shape of t x nobs, where nobs = h x w x n
        :type data: numpy.ndarray
        :param times: Array of reference times relative to the first timestamp
        :type times: numpy.array
        :param resampled_times: Array of reference times relative to the first timestamp in initial timestamp 
         array.
        :type resampled_times: numpy.array
        :return: Array of interpolated values
        :rtype: numpy.ndarray
        """
        # pylint: disable=too-many-locals
        # get size of 2d array t x nobs
        nobs = data.shape[-1]
        if self.interpolate_pixel_wise:
            if 'xarray' in str(inspect.getmodule(self.interpolation_object)):
                locs = np.array(range(data.shape[1]))
                y = xr.DataArray(data, dims=['time','space'], coords=[times,locs])
                y_chunked = y.chunk({'space': 1000})
                tck = self.get_interpolation_function(y_chunked).compute()
                new_data = splev(resampled_times, tck)
            else:
                   # initialise array of interpolated values
                   new_data = data if self.resample_range is None else np.full(
                   (len(resampled_times), nobs),
                   np.nan, dtype=data.dtype)

                   # Interpolate for each pixel, could be easily parallelized
                   for obs in range(nobs):
                         valid = ~np.isnan(data[:, obs])
                         obs_interpolating_func = self.get_interpolation_function(times[valid], data[valid, obs])
                         new_data[:, obs] = obs_interpolating_func(resampled_times[:, np.newaxis])

        # return interpolated values
        return new_data
```

Adaptation in `get_interpolation_function`:

```
    def get_interpolation_function(self, times, series):
        """ Initializes interpolation model
        :param times: Array of reference times in second relative to the first timestamp
        :type times: numpy.array
        :param series: One dimensional array of time series
        :type series: numpy.array
        :return: Initialized interpolation model class
        """
        if str(inspect.getmodule(self.interpolation_object))[9:14] == 'numpy':
            return partial(self.interpolation_object, xp=times, fp=series, left=np.nan, right=np.nan)
        elif 'xarray' in str(inspect.getmodule(self.interpolation_object)):
            return partial(self.interpolation_object, dim='time')
        return self.interpolation_object(times, series, **self.interpolation_parameters)
```

PR #81, opened 2019-03-25
Add ReadFromTiff EOTask

drewbo:
Adds an `EOTask` to read directly from GeoTIFF with `rasterio` (so should accept local or remote paths, TIFFs or VRTs). Works like this:

```python
from eolearn.io import ReadFromTiff
from eolearn.core import EOPatch
patch = EOPatch.load('example_data/TestEOPatch/')
task = ReadFromTiff('https://landsat-pds.s3.amazonaws.com/c1/L8/190/028/LC08_L1TP_190028_20190110_20190131_01_T1/LC08_L1TP_190028_20190110_20190131_01_T1_B2.TIF', 'LANDSAT_AWS')
task.execute(patch)
plt.imshow(patch.data_timeless['LANDSAT_AWS'][:,:,0])
plt.show()
```
Figure_1

Likely needs review, error handling, and tests prior to merging


io/eolearn/io/local_io.py
```diff
@@ -133,3 +134,68 @@ def execute(self, eopatch, *, filename):
...
 
         return eopatch
+
+
+def _get_window(src, bbox):
+    x_res, y_res = src.transform[0], src.transform[4]
+    p_1 = Proj(bbox.get_crs().ogc_string())
+    p_2 = Proj(**src.crs)
+
+    # project mask boundaries from bbox CRS to source CRS
```
drewbo: I think ideally this would use the sentinel hub `BBox.transform` method but I wasn't able to get it to work with the CRS in this example

AleksMat:
Hi Drew,

I have done some restructuring of the code.

Here are most important changes regarding the task for reading from tiff:
- I renamed `ReadFromTiff` to `ImportFromTiff` just to be more alligned with `ExportToTiff` task.
- To avoid code duplications between `ImportFromTiff` and `ExportToTiff` there is now a parent class `BaseLocalIo`.
- There was a bug in line `data = np.empty(shape=(bands, width, height)).astype(src.profile['dtype'])`. Width and height were in wrong order.
- When calculating window for reading from tiff, values are now rounded to nearest integer values instead of floored.
- Task now works with multi-channel tiff images and supports all types of raster features.
- I removed `width` and `height` parameters you were using in `execute` method. These parametes would only be useful if we wanted to spatially resample data read from tiff file. However it is probably better if there is a separate task for spatial resampling.

I have also done some other changes (@mlubej):
- `image_dtype` variable of `ExportToTiff` is now by default set to `None` which will keep dtype of the feature, about to be exported.
- By default `ExportToTiff` now exports all bands and all times into channels of a tiff image and not just the first channel.
- Some other minor fixes in `ExportToTiff`.
- I rewrote unit tests for `local_io` module and got rid of all those code duplications.

io/eolearn/io/local_io.py
AleksMat: I managed to use `BBox.transform` instead of directly using `pyproj` functions. However next steps, where you calculate a window, are not yet supported in sentinelhub-py.

AleksMat:
By the way, my changes are in developmentseed-feature/readfromgeotiff branch and I have made a PR (https://github.com/developmentseed/eo-learn/pull/1) to your branch.

drewbo:
@AleksMat thanks for the refactor, this should be good to go now (the failing test is from another module). I'll take another pass at #84 soon to better match the code style implemented here

AleksMat:
Thank you Drew for preparing this. I'm merging to develop.

(No worries about those failing CI tests. We have already fixed them in another branch but just haven't merged them to develop yet.)

PR #84, opened 2019-03-26
Feature/mapbox

drewbo:
Adds an `EOTask` to read from Mapbox tiles. Works like this:

```python
from eolearn.core import EOPatch
from eolearn.io import MapboxXYZInput
patch = EOPatch.load('.example_data/TestEOPatch/')
task = MapboxXYZInput('mapbox.satellite', 'MAPBOX', ACCESS_TOKEN)
task.execute(patch, 17)
plt.imshow(patch.data_timeless['MAPBOX'])
plt.show()
```

Figure_3

The zoom parameter is optional. When not provided, zoom is calculated to be the lowest value that still gets tiles with resolution under the current `size_x` or `size_y` (generally zoom 14 for 10 meter resolution except at high latitudes)

### DIFF
--- a/io/eolearn/io/__init__.py
+++ b/io/eolearn/io/__init__.py
@@ -7,5 +7,6 @@ from .sentinelhub_service import SentinelHubOGCInput, SentinelHubWMSInput, Senti
     DEMWMSInput, DEMWCSInput, AddSen2CorClassificationFeature
 from .geopedia import AddGeopediaFeature
 from .local_io import ExportToTiff
+from .tile_io import MapboxXYZInput
 
 __version__ = '0.4.2'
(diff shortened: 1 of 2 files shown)


drewbo:
@AleksMat updated this a bit to better match the overall `eo-learn` style. Let me know if you are able to review and need any changes

vincentsarago:
hey @drewbo I wonder how hard would it be to make this a more general tile service handler ? like instead of making it Mapbox only maybe this `EOTask` class could take a template URL (e.g `url="http://api.mapbox.com/v4/mapbox.satellite/{z}/{x}/{y}.png?access_token=afake_access_token_here"`) as input. 

I'll be happy to work on this if you want @drewbo

drewbo:
Closed in favor of #148

PR #87, opened 2019-04-12
raise error for no good_idx

j08lue:
When filtering images in a patch, it can occur that the filter evaluates to `False` for all images (e.g. in all-year cloudy spots where you apply a strict cloud cover threshold).

This situation should be caught nicely, not raise an error deeper down.

This PR adds a simple test for empty `good_idx`.

What do you think?

(with @buwuyou)

### DIFF
--- a/features/eolearn/features/feature_manipulation.py
+++ b/features/eolearn/features/feature_manipulation.py
@@ -52,6 +52,8 @@ class SimpleFilterTask(EOTask):
 
         good_idxs = self._get_filtered_indices(eopatch[feature_type][feature_name] if feature_name is not ... else
                                                eopatch[feature_type])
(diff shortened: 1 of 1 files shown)


AleksMat:
This contribution has now been included in `eo-learn` version `0.5.0`, [here](https://github.com/sentinel-hub/eo-learn/blob/master/features/eolearn/features/feature_manipulation.py#L55).

We didn't merge this PR directly only because it uses f-strings and we would like that eo-learn supports Python 3.5 as well.

Thanks for help. :)

j08lue:
> We didn't merge this PR directly only because it uses f-strings and we would like that eo-learn supports Python 3.5 as well.

We could have fixed that, but no worries.

PR #93, opened 2019-05-16
Make 16bit tiff the default format to be requested

batic:
This commits also allows adding temporal data to previously timeless eopatch. 

Tests check if data after requesting 16bit tiffs are (to some small rounding error) equal to requesting 32bit tiffs.

### DIFF
--- a/io/eolearn/io/sentinelhub_service.py
+++ b/io/eolearn/io/sentinelhub_service.py
@@ -6,9 +6,10 @@ import logging
 import datetime as dt
 
 import numpy as np
-from eolearn.core import EOPatch, EOTask, FeatureType, get_common_timestamps
 
 from sentinelhub import WmsRequest, WcsRequest, MimeType, DataSource, CustomUrlParam, ServiceType
+
+from eolearn.core import EOPatch, EOTask, FeatureType, get_common_timestamps
 
 
 LOGGER = logging.getLogger(__name__)
@@ -36,7 +37,9 @@ class SentinelHubOGCInput(EOTask):
     :type size_y: int or str, depends on the service_type
     :param maxcc: maximum accepted cloud coverage of an image. Float between 0.0 and 1.0. Default is ``1.0``.
     :type maxcc: float
-    :param image_format: format of the returned image by the Sentinel Hub's WMS getMap service. Default is 16-bit TIFF.
+    :param image_format: format of the returned image by the Sentinel Hub's WMS getMap service. Default is `None` which
+        means values will be downloaded in 16-bit tiff but then scaled back to [0,1] float values. The result is the
+        same as by setting `MimeType.TIFF_d32f`, but with less data to download.
     :type image_format: constants.MimeType
     :param instance_id: user's instance id. If ``None`` the instance id is taken from the ``config.json``
                         configuration file from sentinelhub-py package.
@@ -337,14 +340,13 @@ class DEMWCSInput(SentinelHubWCSInput):
     """
     Adds DEM to DATA_TIMELESS EOPatch feature.
     """
-    def __init__(self, layer, feature=None, **kwargs):
+    def __init__(self, layer, feature=None,
+                 valid_data_mask_feature=(FeatureType.MASK_TIMELESS, 'IS_DATA_TIMELESS'), **kwargs):
         if feature is None:
             feature = (FeatureType.DATA_TIMELESS, layer)
         elif isinstance(feature, str):
             feature = (FeatureType.DATA_TIMELESS, feature)
-        # if 'valid_data_mask_feature' not in kwargs:
-        mask_feature = (FeatureType.MASK_TIMELESS, 'IS_DATA_TIMELESS')
-        super().__init__(layer=layer, feature=feature, valid_data_mask_feature=mask_feature,
+        super().__init__(layer=layer, feature=feature, valid_data_mask_feature=valid_data_mask_feature,
                          data_source=DataSource.DEM, **kwargs)
 
 
--- a/io/eolearn/io/sentinelhub_service.py
+++ b/io/eolearn/io/sentinelhub_service.py
@@ -6,9 +6,10 @@ import logging
 import datetime as dt
 
 import numpy as np
+from eolearn.core import EOPatch, EOTask, FeatureType, get_common_timestamps
+
 from sentinelhub import WmsRequest, WcsRequest, MimeType, DataSource, CustomUrlParam, ServiceType
 
-from eolearn.core import EOPatch, EOTask, FeatureType, get_common_timestamps
 
 LOGGER = logging.getLogger(__name__)
 
@@ -22,8 +23,9 @@ class SentinelHubOGCInput(EOTask):
     :param feature: Feature to which the data will be added. By default the name will be the same as the name of the
         layer
     :type feature: str or (FeatureType, str) or None
-    :param valid_data_mask_feature: A feature to which valid data mask will be stored. Default is `'IS_DATA'`.
-    :type valid_data_mask_feature: str or (FeatureType, str)
+    :param valid_data_mask_feature: A feature to which valid data mask will be stored. Default is
+        `(FeatureType.MASK, 'IS_DATA')`.
+    :type valid_data_mask_feature: (FeatureType, str)
     :param service_type: type of OGC service (WMS or WCS)
     :type service_type: ServiceType
     :param data_source: Source of requested satellite data.
@@ -34,7 +36,7 @@ class SentinelHubOGCInput(EOTask):
     :type size_y: int or str, depends on the service_type
     :param maxcc: maximum accepted cloud coverage of an image. Float between 0.0 and 1.0. Default is ``1.0``.
     :type maxcc: float
-    :param image_format: format of the returned image by the Sentinel Hub's WMS getMap service. Default is 32-bit TIFF.
+    :param image_format: format of the returned image by the Sentinel Hub's WMS getMap service. Default is 16-bit TIFF.
     :type image_format: constants.MimeType
     :param instance_id: user's instance id. If ``None`` the instance id is taken from the ``config.json``
                         configuration file from sentinelhub-py package.
@@ -56,15 +58,17 @@ class SentinelHubOGCInput(EOTask):
     :type raise_download_errors: bool
     """
 
-    def __init__(self, layer, feature=None, valid_data_mask_feature='IS_DATA', service_type=None, data_source=None,
-                 size_x=None, size_y=None, maxcc=None, image_format=MimeType.TIFF_d32f, instance_id=None,
-                 custom_url_params=None, time_difference=None, raise_download_errors=True):
+    def __init__(self, layer, feature=None, valid_data_mask_feature=(FeatureType.MASK, 'IS_DATA'),
+                 service_type=None, data_source=None, size_x=None, size_y=None, maxcc=None, image_format=None,
+                 instance_id=None, custom_url_params=None, time_difference=None, raise_download_errors=True):
         # pylint: disable=too-many-arguments
         self.layer = layer
         self.feature_type, self.feature_name = next(self._parse_features(layer if feature is None else feature,
                                                                          default_feature_type=FeatureType.DATA)())
-        self.valid_data_mask_feature = self._parse_features(valid_data_mask_feature,
-                                                            default_feature_type=FeatureType.MASK)
+        self.valid_data_mask_feature_type, self.valid_data_mask_feature_name = \
+            next(self._parse_features(layer+'_MASK' if valid_data_mask_feature is None else valid_data_mask_feature,
+                                      default_feature_type=FeatureType.MASK)())
+
         self.service_type = service_type
         self.data_source = data_source
         self.size_x = size_x
@@ -97,6 +101,8 @@ class SentinelHubOGCInput(EOTask):
             return dt.timedelta(seconds=-1)
         if name in ('size_x', 'size_y'):
             return None
+        if name == 'time_interval' and not self.feature_type.is_time_dependent():
+            return None
 
         raise ValueError('Parameter {} was neither defined in initialization of {} nor is contained in '
                          'EOPatch'.format(name, self.__class__.__name__))
@@ -119,7 +125,7 @@ class SentinelHubOGCInput(EOTask):
             'time': time_interval,
             'time_difference': self._get_parameter('time_difference', eopatch),
             'maxcc': self._get_parameter('maxcc', eopatch),
-            'image_format': self.image_format,
+            'image_format': self.image_format if self.image_format else MimeType.TIFF_d16,
             'custom_url_params': self.custom_url_params,
             'data_source': self.data_source,
             'instance_id': self.instance_id,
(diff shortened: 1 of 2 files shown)


io/eolearn/io/sentinelhub_service.py
```diff
@@ -34,7 +36,7 @@ class SentinelHubOGCInput(EOTask):
     :type size_y: int or str, depends on the service_type
     :param maxcc: maximum accepted cloud coverage of an image. Float between 0.0 and 1.0. Default is ``1.0``.
     :type maxcc: float
-    :param image_format: format of the returned image by the Sentinel Hub's WMS getMap service. Default is 32-bit TIFF.
+    :param image_format: format of the returned image by the Sentinel Hub's WMS getMap service. Default is 16-bit TIFF.
```
AleksMat: If you say that default is 16-bit that is not correct and confusing to the user because it is not the same as specifying `image_format=MimeType.TIFF_d16`.

Better to write more precise explanation:
*"Default is `None` which means values will be downloaded in 16-bit tiff but then cast into reflectance float values. The result is the same as by setting `MimeType.TIFF_d32f`, but with less data to download."*

io/eolearn/io/sentinelhub_service.py
```diff
@@ -132,6 +138,10 @@ def _add_data(self, eopatch, data):
         valid_mask = data[..., -1]
         data = data[..., :-1]
 
+        if self.image_format is None:
+            #image format was set to MimeType.TIFF_d16
+            data = data/2**16
```
AleksMat: Shouldn't this be `data / 10 ** 4` (and also casting to `numpy.float32`)?
I have seen you updated a lot of unit tests but do you anywhere check if `image_format=None` returns the same result as `image_format=MimeType.TIFF_d32f`? I think such test would be mandatory.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -325,7 +342,10 @@ def __init__(self, layer, feature=None, **kwargs):
             feature = (FeatureType.DATA_TIMELESS, layer)
         elif isinstance(feature, str):
             feature = (FeatureType.DATA_TIMELESS, feature)
-        super().__init__(layer=layer, feature=feature, data_source=DataSource.DEM, **kwargs)
+        # if 'valid_data_mask_feature' not in kwargs:
+        mask_feature = (FeatureType.MASK_TIMELESS, 'IS_DATA_TIMELESS')
```
AleksMat: You should put this as an init parameter: 
```Python
def __init__(self, layer, feature=None, valid_data_mask_feature=(FeatureType.MASK_TIMELESS, 'IS_DATA_TIMELESS'), **kwargs):
```
and then pass it on as such.
The problem in current version is that if user specifies different value for `valid_data_mask_feature` parameter it gets clashed with what you specify here by default.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -6,9 +6,10 @@
 import datetime as dt
 
 import numpy as np
+from eolearn.core import EOPatch, EOTask, FeatureType, get_common_timestamps
```
AleksMat: Why changing order of imports? In general imports from the same package (or the same namespace package in this case) are written last.

io/eolearn/tests/test_io.py
```diff
@@ -321,7 +345,192 @@ def test_feature_dimension(self):
...
+    def test_time_data_to_dem_eopatch(self):
+        with self.subTest(msg='Test to add timed data to DEM (timeless) eopatch'):
+            dem_eopatch = self.dem_request.execute(bbox=self.bbox)
+            timed_eopatch = self.timed_request.execute(eopatch=dem_eopatch, bbox=self.bbox,
+                                                       time_interval=self.time_interval)
+            self.assertTrue(len(timed_eopatch.timestamp) == 1, 'timed_eopatch should have 1 timestamp!')
+            self.assertTrue(timed_eopatch.meta_info['time_interval'], '\'time_interval\' in meta_info is expected!')
+
+
+class TestImageFormats(unittest.TestCase):
```
AleksMat: I haven't checked all the details but it looks to me there is some code repetition between `TestEOPatch` and `TestImageFormats` classes.
Couldn't you just add `test_image_format` function to `TestEOPatch` class? (That way you also reduce the amount of data needed to download.)

Btw, feel free to rename `TestEOPatch` class, not very good name for here.

io/eolearn/tests/test_io.py
```diff
@@ -262,11 +264,33 @@ def setUpClass(cls):
                     instance_id=instance_id
                 ),
                 bbox=bbox,
-                time_interval=cls.time_interval,
+                time_interval=long_time_interval,
                 feature_type=FeatureType.MASK
             ),
         ]
 
+        cls.task_cases_image_format = [
```
AleksMat: This test case is not used anywhere.

io/eolearn/io/sentinelhub_service.py
batic: Right you are, will fix.

io/eolearn/io/sentinelhub_service.py
batic: No. Sentinel-Hub OGC will always (as the last step in processing) scale values according to requested image type (uint8 for png/jpg/tiff, uint16 for tiff/depth=16 and float32 for tiff/depth=32f). Scaling is done as follows:
* int8: [0,1] -> [0, 256]
* int16: [0,1] -> [0, 2**16]
* float32: identity (no scaling)

The additional tests in class `TestImageFormats` all check that results are the "same"; not for identity, but rather for equality to some precision. Each test case does two requests, one in 32float and another with default `image_format` (`None`), and the results are compared.

io/eolearn/io/sentinelhub_service.py
batic: Am I missing something? The next line sets this precisely. Unfortunately, the comment is a bit misleading, so I'll remove it.

batic: Ah, now I see what you mean.

As this is `DEMWCSInput`, and as such a "helper" for the user, I would even go as far as not allow user to specify this parameter (e.g. removing from `**kwargs` if it exists). Will add as a default value in `__init__` for the time being.

io/eolearn/io/sentinelhub_service.py
batic: Because of linter. Honestly, I don't care.

io/eolearn/tests/test_io.py
batic: It is the only one with defined `image_format=MimeType.TIFF_d32f`. I guess I forgot to add it to list of `test_cases`.

io/eolearn/tests/test_io.py
batic: I had two purposes when working on tests:
* reduce downloaded data from SH (and making the tests faster). In order to do that I've picked a smaller BBox and defined much narrower `time_interval` (even split tests into two distinct sets, one with single timestamp and one with a few)
* make sure that refactoring `image_format` will result with equal results.
For the second task I've added `TestImageFormats`, where for each test two EOpatches are created and then compared. This is different from `TestEOPatch` where the comparison is done more-or-less on metadata of the resulting EOPatch.

batic: Additionally, `TestTimelessFeatures` test was added to test the fix in the code when temporal data is being added to completely timeless EOPatch (which was not working before, but is allowed now).

io/eolearn/io/sentinelhub_service.py
AleksMat: 2 more things regarding this:
- `DEMWCSInput' now has this `valid_data_mask_feature` parameter, but `DEMWMSInput` doesn't. Should also `DEMWMSInput` have it?
- Why is this mask parameter necessary? I thought DEM is available globally and valid data mask would always have all True values?

io/eolearn/tests/test_io.py
AleksMat: Ok, I agree with this.

io/eolearn/io/sentinelhub_service.py
AleksMat: You are right about that. However I still have 2 concerns:

- Maximal value in 16-bit tiff is actually `65535 = 2 ** 16 - 1`. So I assume the scaling is done with [0,1] -> [0, 2**16 - 1]. That would mean that you have to divide by `2**16 - 1` instead of `2**16`. The differences between values you get from these 2 divisions have order of magnitude 1e-6 (I checked), so you won't see them in unit tests. But it still might be better to verify this with the core team. By the way, you can use `MimeType.TIFF_d16.get_expected_max_value()` to get the value `2 ** 16 - 1`.

- Sometimes reflectance values are greater than `1.0`. In that case if you request  data in `MimeType.TIFF_d32f`, will the service return you these original values `>1.0`? Because when requesting data in `MimeType.TIFF_d16` the values will be transformed down to `2**16 - 1` and that could cause inconsistencies between using `image_format=None` and `image_format=MimeType.TIFF_d32f`.

Sorry for complicating, but this is the most sensitive part of the PR and I would like us to be as consistent as possible.

batic: You are not complicating, that is the whole point of having PR and reviewing them. Thanks for the notes, will check and get back to you.

io/eolearn/io/sentinelhub_service.py
batic: Probably, yes, will check and add if needed.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -355,12 +377,46 @@ def __init__(self, sen2cor_classification, layer, **kwargs):
             'SNW': FeatureType.DATA
         }
 
+        if 'image_format' in kwargs:
+            image_format = kwargs.get('image_format')
+            kwargs.pop('image_format', None)
+            sample_type = image_format.get_sample_type()
+        else:
+            image_format = MimeType.TIFF_d8
+            sample_type = 'INT8'
```
AleksMat: Instead of `sample_type = 'INT8'` it is better that you do `sample_type = image_format.get_sample_type()` (and move it out of `if` sentence.)
That is because you don't want to hardcode `'INT8'` here since it already exists in sentinelhub-py.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -355,12 +377,46 @@ def __init__(self, sen2cor_classification, layer, **kwargs):
...
+
         if sen2cor_classification not in classification_types:
             raise ValueError('Unsupported Sen2Cor classification type: {}.'
                              ' Possible types are: {}'.format(sen2cor_classification, classification_types))
 
-        evalscript = 'return [{}];'.format(sen2cor_classification)
+        evalscript = '''\
```
AleksMat: I suggest moving this evalscript template outside the class and have it as a constant (maybe even in a separate module?). Defining it here in the middle of the method looks weird.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -355,12 +377,46 @@ def __init__(self, sen2cor_classification, layer, **kwargs):
...
+    return {{
+        default: [
+            samples[0].{classification_type}
+        ]
+    }};
+}}\
+'''
+
+        evalscript_args = {'classification_type': sen2cor_classification, 'sample_type': sample_type}
+        evalscript = evalscript.format(**evalscript_args)
```
AleksMat: You could join the 2 lines into 1:
```Python
evalscript = evalscript.format(classification_type=sen2cor_classification, sample_type=sample_type)
```

io/eolearn/tests/test_io.py
```diff
@@ -321,7 +345,192 @@ def test_feature_dimension(self):
...
+
+        ]
+
+        cls.dem_test = [
+        ]
+
+    def test_image_format(self):
+        for test in self.tests:
+            with self.subTest(msg='Test case {}'.format(test.name)):
+                if test.feature_type == FeatureType.DATA:
```
AleksMat: 2 comments here:

- You have 2 blocks of code in (if statements) that are almost duplicated. You could do:
```Python
atol_value = 1e-4 if test.feature_type is FeatureType.DATA else 0
```
or even nicer:
```Python
atol_value = 0 if test.feature_type.is_discrete() else 1e-4
```
and then run `np.testing.assert_allclose`

- For the future: it is more correct to compare 2 objects (e.g. Enums) with `is` instead of `==`. The `==` is meant for comparing values.

io/eolearn/io/sentinelhub_service.py
batic: Missed this one.

io/eolearn/io/sentinelhub_service.py
batic: General idea (with upcoming Processing API) is that evalscripts will be extracted in the way you describe here. As we are not there yet, I believe the evalscript pertaining to extracting SCL masks is better defined here. I suggest creating a new module for evalscripts as a new PR.

io/eolearn/io/sentinelhub_service.py
batic: 👍

io/eolearn/tests/test_io.py
batic: 👍

io/eolearn/io/sentinelhub_service.py
AleksMat: Ok, then we can leave it as it is for now.

io/eolearn/io/sentinelhub_service.py
batic: I didn't add `valid_data_mask_feature` to `DEMWMSInput` as it didn't have it so far. In any case WMS is returning (typically) a 3-band jpg image, so I believe it is ok as is.

batic: The DEM we have at SentinelHub should be available everywhere, so you are right that it is probably not needed. 
Regarding the mask itself; I don't recall why I added it there. I'll remove it.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -34,7 +37,9 @@ class SentinelHubOGCInput(EOTask):
     :type size_y: int or str, depends on the service_type
     :param maxcc: maximum accepted cloud coverage of an image. Float between 0.0 and 1.0. Default is ``1.0``.
     :type maxcc: float
-    :param image_format: format of the returned image by the Sentinel Hub's WMS getMap service. Default is 32-bit TIFF.
+    :param image_format: format of the returned image by the Sentinel Hub's WMS getMap service. Default is `None` which
```
azupanc: Why is `image_format` parameter set to `None` by default here, if actually it is set to `MimeType.TIFF_d16` later on? Why not explicitly set the default value here so that later on the checks (if ... None ... else ...)?

batic: Will have to check. It is quite some time since this PR has been made (and closed).

io/eolearn/io/sentinelhub_service.py
batic: The idea was to make a request in uint16, but then convert it to 0-1 (or better: to numbers that would be received using float32 request). 
Later on the data in eopatches would be (as is now) stores in 32bit floats.

azupanc: Yes, I understand. I recommend to review also an option where data would be stored in its native precision (int16) and only convert (and optionally scale) it on-the-fly when needed.

PR #94, opened 2019-05-26
Feat/super pixel

AleksMat:
Added tasks for super-pixel segmentation + minor fix in visualizations.


geometry/eolearn/geometry/superpixel.py
```diff
@@ -0,0 +1,88 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+class SuperpixelSegmentation(EOTask):
+    """ Super-pixel segmentation task
+
+    Given a raster feature it will segment data into super-pixels. Representation of super-pixels will be returned as
+    a mask timeless feature where all pixels with the same value belong to one super-pixel
+    """
+    def __init__(self, feature, superpixel_feature, *, segmentation_object=skimage.segmentation.felzenszwalb,
```
batic: Why is there a * (additional positional arguments, if I understand correctly) in the `__init__`? They are never used, and never passed onwards. If this is to be future proof, then at least make them `*args` or so.

AleksMat: `*` doesn't specify any new parameters, it is the way how to force keyword-only arguments. In this case it means that some can do
```Python
SuperpixelSegmentation(my_feature, my_superpixel_feature,
    segmentation_object=my_segmentation_object)
```
but can't do
```Python
SuperpixelSegmentation(my_feature, my_superpixel_feature,
    my_segmentation_object)
```
This way you leave an option that more keyword-only arguments might be added in the future and their order might change. Because you force users to write parameter names, their code will be more consistent and won't break because of such changes.
Some reference: https://stackoverflow.com/questions/2965271/forced-naming-of-parameters-in-python

batic:
When running the test, I've noticed the warning:
DeprecationWarning: Bounding box of your EOPatch is saved in old format which in the future won't be supported anymore. Please save bounding box again, you can overwrite the existing one 

Could you please fix the bbox in the test eopatch with this PR?

geometry/eolearn/geometry/superpixel.py
```diff
@@ -0,0 +1,88 @@
...
+                 **segmentation_params):
+        """
+        :param feature: Raster feature which will be used in segmentation
+        :param superpixel_feature: A new mask timeless feature to hold super-pixel mask
+        :param segmentation_object: A function (object) which performs superpixel segmentation, by default that is
+            `skimage.segmentation.felzenszwalb`
+        :param segmentation_params: Additional parameters which will be passed to segmentation_object function
+        """
+        self.feature_checker = self._parse_features(feature, allowed_feature_types=FeatureTypeSet.SPATIAL_TYPES)
```
batic: Any reason why `feature_checker` is build in `__init__`, and then `next` is called to retrieve the feature only in `execute`, while `superpixel_feature` is constructed here?

AleksMat: The idea is that you can use `FeatureParser` with or without having an EOPatch. If you parse with an EOPatch it will also make sure that parsed features exist in the EOPatch before returning them.

In this case the `feature` parameter is an input feature for which we have to check that it exists in any given EOPatches. So we collect parsed feature from `FeatureParser` only in `execute` method, when we have an EOPatch.
On the other hand `superpixel_feature` is a new feature which does not have to exist in EOPatches yet. So we can finish parsing the feature right away in the init.

I agree that the way this is done looks a bit weird. An for the future there is a plan to improve the way features are parsed, but that will have to be done in the core package.

AleksMat:
@batic I re-saved the bbox in test EOPatch so that `DeprecationWarning` isn't being raised anymore.

I also suppressed a redundant warning from `skimage.segmentation`.

PR #95, opened 2019-05-28
Feat/vector raster transform improve

AleksMat:
The following PR contains some code breaking changes in `VectorToRaster` task so it should be reviewed carefully.

Change log:
- Split `eolearn.geometry.utilities` module into `eolearn.geometry.transformations` and `eolearn.geometry.morphology`.
- Large changes in `VectorToRaster` - new/renamed parameters and improve functionalities
- Some improvements in `RasterToVector`
- Added unit tests for `VectorToRaster` and `RasterToVector`

Note: This PR already includes changes from PR #94, so that one should be merged first.


AleksMat:
Note: This PR also contains changes from https://github.com/sentinel-hub/eo-learn/pull/99 so it should be merged after that one.

core/eolearn/core/eodata.py
```diff
@@ -103,37 +107,59 @@ def _parse_feature_type_value(feature_type, value):
...
+
+        return self.__setattr__(FeatureType(feature_type).value, value, feature_name=feature_name)
+
+    @staticmethod
+    def _check_tuple_key(key):
+        """ A helper function that checks a tuple, which should hold (feature_type, feature_name)
+        """
+        if len(key) != 2:
+            raise ValueError('Given element should be a feature_type or a tuple of (feature_type, feature_name),'
+                             'but {} found'.format(key))
```
mlubej: Just a general question, do you prefer this over f-strings? Or is it just that old habits die hard?

geometry/eolearn/geometry/transformations.py
```diff
@@ -0,0 +1,340 @@
...
+        :type eopatch: EOPatch
+        :return: New EOPatch with vector data transformed into a raster feature
+        :rtype: EOPatch
+        """
+        if eopatch.bbox is None:
+            raise ValueError('EOPatch has to have a bounding box')
+
+        rasterization_shapes = self._get_rasterization_shapes(eopatch)
+        if not rasterization_shapes:
+            return eopatch
```
mlubej: Should there be some warning here that nothing happens or that no shapes are present?

mlubej: Ah, nevermind, I guess that sometimes vector data just isn't present in an EOPatch

geometry/eolearn/geometry/transformations.py
```diff
@@ -0,0 +1,340 @@
...
+                             "".format(eopatch.bbox.crs, vector_data_crs))
+
+        bbox_poly = eopatch.bbox.geometry
+        vector_data = vector_data[vector_data.geometry.intersects(bbox_poly)].copy(deep=True)
+
+        if vector_data.empty:
+            return None
+
+        if not vector_data.geometry.is_valid.all():
+            warnings.warn('Given vector polygons contain some invalid geometries, they will be fixed', RuntimeWarning)
```
mlubej: Just a question, what is considered as an invalid geometry in this case?

geometry/eolearn/geometry/transformations.py
```diff
@@ -0,0 +1,340 @@
...
+        """
+        if eopatch.bbox is None:
+            raise ValueError('EOPatch has to have a bounding box')
+
+        rasterization_shapes = self._get_rasterization_shapes(eopatch)
+        if not rasterization_shapes:
+            return eopatch
+
+        height, width = self._get_raster_shape(eopatch)
+        data_transform = rasterio.transform.from_bounds(*eopatch.bbox, width=width, height=height)
```
mlubej: If I understand correctly, the `data_transform` here is the transformation function?

core/eolearn/core/eodata.py
AleksMat: This is because we are trying to support Python 3.5 as well which doesn't have f-strings.

geometry/eolearn/geometry/transformations.py
AleksMat: Yes, this only happens if vector feature has no polygons which have to be rasterized.

geometry/eolearn/geometry/transformations.py
AleksMat: This property is from `shapely`: [is_valid docs](https://shapely.readthedocs.io/en/stable/manual.html#object.is_valid). Basically the geometry must not be self-intersecting.

geometry/eolearn/geometry/transformations.py
AleksMat: Not a function but an `Affine` transformation vector object that rasterio returns for given bounding box coordinates and dimensions. Maybe it would be better if it would be named `affine_transform`?

mlubej:
Looks good to me!

PR #99, opened 2019-06-18
Feat/eoexecutor methods

AleksMat:
Changes:
- Added `EOExecutor` methods `get_successful_executions` and `get_failed_executions`
- Added an option of `EOExecutor.run` to switch between multiprocessing and multithreading
- Logs in a report are now by default hidden in a collapsible
- Minor logging improvements
- Added a simplified way to access features in EOPatch: e.g. `eopatch[(FeatureType.DATA, 'BANDS')]

### DIFF
--- a/core/eolearn/core/eoexecution.py
+++ b/core/eolearn/core/eoexecution.py
@@ -31,7 +31,7 @@ from jinja2 import Environment, FileSystemLoader
 
 from .eoworkflow import EOWorkflow
 
-LOGGER = logging.getLogger(__file__)
+LOGGER = logging.getLogger(__name__)
 
 
 class EOExecutor:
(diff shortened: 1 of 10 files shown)


core/eolearn/core/eoexecution.py
```diff
@@ -75,12 +75,15 @@ def _parse_execution_args(execution_args):
...
         """ Runs the executor with n workers.
 
-        :param workers: Number of parallel processes used in the execution. Default is a single process. If set to
-            `None` the number of workers will be the number of processors of the system.
+        :param workers: Maximum number of workflows which will be executed in parallel. Default is a single workflow.
+            If set to `None` the number of workers will be the number of processors of the system.
         :type workers: int or None
+        :param multiprocess: If `True` it will use multiple processors for execution. If `False` it will run on
+            multiple threads of a single processor.
+        :type multiprocess: bool
```
mlubej: What about the fact that we don't really see this happening when we use the threadpoolexecutor?

AleksMat: You are right, this docs string is not clear enough. Added a better explanation.

mlubej:
Looks good to me, you can merge.

PR #107, opened 2019-07-12
ExportToTiff: Create dir if it doesn't exist

lojzezust:
Added flag `create_dir` to function `_get_file_path` in `BaseLocalIo`. If set to `True` it creates the missing directories in the path. Useful for `EOTasks` that save files such as `ExportToTiff`.


AleksMat:
I agree with the new functionality. It falls in line with saving of EOPatches, which also automatically creates missing folders.

The code seems to be ok, no comments about code style.

Good job :+1:

PR #110, opened 2019-07-18
Improvement: new resizing method

lojzezust:
Implemented a `resize_images` utility function. The new implementation uses OpenCV for image smoothing and resizing which results in big performance gains (around 4x faster on average). Updated `AddCloudMaskTask` to use the new function.

New function correctly computes the sigma used for anti aliasing and correctly aligns the images on resize. The function can be reused in other modules, where resizing is done.

The values in tests for the `AddCloudMaskTask` were also updated to allow for the new sigma computation and alignment.

Also added `map_image_slices` utilitiy function. It maps a given function over all `(height, width)` image slices in an array of shape `(time, height, width, channels)`. Might be of use elswhere. If not we can refactor it inside `resize_images`.


core/eolearn/core/utilities.py
```diff
@@ -475,3 +476,80 @@ def constant_pad(X, multiple_of, up_down_rule='even', left_right_rule='even', pa
...
+        new_size = [int(d * f) for d,f in zip(old_size, scale_factors)]
+    else:
+        raise ValueError('Exactly one of the arguments new_size, scale_factors must be given.')
+
+    if interpolation not in INTER_METHODS:
+        raise ValueError('Invalid interpolation method: %s' % interpolation)
+
+    interpolation_method = INTER_METHODS[interpolation]
+    downscaling = scale_factors[0] < 1 or scale_factors[0] < 1
+    new_size = tuple(new_size)
```
andrejburja: i would put this in `resized = cv2.resize(image, tuple(new_size), interpolation=interpolation_method)`

core/eolearn/core/utilities.py
```diff
@@ -475,3 +476,80 @@ def constant_pad(X, multiple_of, up_down_rule='even', left_right_rule='even', pa
...
+
+    def _resize(image):
+        # Perform anti-alias smoothing if downscaling
+        if downscaling and anti_alias:
+            sx, sy = [((1/s) - 1)/2 for s in scale_factors]
```
andrejburja: if `sx` is sigmax just write it like that

core/eolearn/core/utilities.py
```diff
@@ -475,3 +476,80 @@ def constant_pad(X, multiple_of, up_down_rule='even', left_right_rule='even', pa
...
+
+def resize_images(data, new_size=None, scale_factors=None, anti_alias=True, interpolation='linear'):
+    """Resizes the images acording to given size or scale factors.
+
+    To specify the new scale use one of `new_size` or `scale_factors` parameters.
+    
+    :param data: input array
+    :type data: array of shape (timestamps, rows, columns, channels)
+    :param new_size: New size of the data (width, height)
+    :type new_size: (int, int)
```
andrejburja: what happens if new_size is not int?

devisperessutti:
I've realised the `test_cloud_mask` uses a local eopatch `mask/eolearn/test/eopatch_sample`. 

This was probably used before we introduced a unified `TestEOPatch` in `example_data`. Can you update the unti-tests to use the unified `TestEOPatch` and delete the `mask/eolearn/test/eopatch_sample`?

core/eolearn/core/utilities.py
```diff
@@ -475,3 +476,80 @@ def constant_pad(X, multiple_of, up_down_rule='even', left_right_rule='even', pa
...
+    elif scale_factors is not None and new_size is None:
+        new_size = [int(d * f) for d,f in zip(old_size, scale_factors)]
+    else:
+        raise ValueError('Exactly one of the arguments new_size, scale_factors must be given.')
+
+    if interpolation not in INTER_METHODS:
+        raise ValueError('Invalid interpolation method: %s' % interpolation)
+
+    interpolation_method = INTER_METHODS[interpolation]
+    downscaling = scale_factors[0] < 1 or scale_factors[0] < 1
```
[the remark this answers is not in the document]
lojzezust: Thanks! Changed to `scale_factors[1]`. We  check if any of the scaling factors is less than one.

core/eolearn/core/utilities.py
```diff
@@ -475,3 +476,80 @@ def constant_pad(X, multiple_of, up_down_rule='even', left_right_rule='even', pa
...
+        'cubic': cv2.INTER_CUBIC
+    }
+
+    # Old width and height
+    old_size = (data.shape[2], data.shape[1])
+
+    if new_size is not None and scale_factors is None:
+        scale_factors = [new/old for old, new in zip(old_size, new_size)]
+    elif scale_factors is not None and new_size is None:
+        new_size = [int(d * f) for d,f in zip(old_size, scale_factors)]
```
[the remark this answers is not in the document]
lojzezust: `d` as dimension and `f` as factor. Changed to `size` and `factor`.

core/eolearn/core/utilities.py
```diff
@@ -475,3 +476,80 @@ def constant_pad(X, multiple_of, up_down_rule='even', left_right_rule='even', pa
...
+    
+    return output
+
+def resize_images(data, new_size=None, scale_factors=None, anti_alias=True, interpolation='linear'):
+    """Resizes the images acording to given size or scale factors.
+
+    To specify the new scale use one of `new_size` or `scale_factors` parameters.
+    
+    :param data: input array
+    :type data: array of shape (timestamps, rows, columns, channels)
```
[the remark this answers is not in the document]
lojzezust: numpy. Updated the docstrings.

core/eolearn/core/utilities.py
lojzezust: Currently the error is raised by OpenCV. (`TypeError: integer argument expected, got float`)

There are other options. We could raise our own error, or just round to the nearest integer?

lojzezust:
> I've realised the `test_cloud_mask` uses a local eopatch `mask/eolearn/test/eopatch_sample`.
> 
> This was probably used before we introduced a unified `TestEOPatch` in `example_data`. Can you update the unti-tests to use the unified `TestEOPatch` and delete the `mask/eolearn/test/eopatch_sample`?

@devisperessutti Sure. I will look into it.

lojzezust:
I wrote the code to only work on 4d arrays. Maybe it would be better if it was generalized to work also on 3d features (timeless) or maybe even 2d if needed.

andrejburja:
> I wrote the code to only work on 4d arrays. Maybe it would be better if it was generalized to work also on 3d features (timeless) or maybe even 2d if needed.

can you please adapt the code to work also on 4d, 3d, 2d data.

lojzezust:
> I've realised the `test_cloud_mask` uses a local eopatch `mask/eolearn/test/eopatch_sample`.
> 
> This was probably used before we introduced a unified `TestEOPatch` in `example_data`. Can you update the unti-tests to use the unified `TestEOPatch` and delete the `mask/eolearn/test/eopatch_sample`?

I've added the tests for the TestEOPatch example.

lojzezust:
> > I wrote the code to only work on 4d arrays. Maybe it would be better if it was generalized to work also on 3d features (timeless) or maybe even 2d if needed.
> 
> can you please adapt the code to work also on 4d, 3d, 2d data.

I've adapted the code to work on 4d, 3d and 2d data. Should I add some tests for the resizing. If so, where should I put them?

core/eolearn/core/utilities.py
```diff
@@ -475,3 +476,113 @@ def constant_pad(X, multiple_of, up_down_rule='even', left_right_rule='even', pa
...
+    return res
+
+def resize_images(data, new_size=None, scale_factors=None, anti_alias=True, interpolation='linear'):
+    """Resizes the image(s) acording to given size or scale factors.
+
+    To specify the new scale use one of `new_size` or `scale_factors` parameters.
+    
+    :param data: input image array
+    :type data: numpy array with shape (timestamps, height, width, channels), (height, width, channels), or (height, width)
+    :param new_size: New size of the data (width, height)
```
azupanc: Probably a typo: `new_size` is tuple of `(height, width)` and not (width, height) as stated in the docs.

core/eolearn/core/utilities.py
```diff
@@ -475,3 +476,113 @@ def constant_pad(X, multiple_of, up_down_rule='even', left_right_rule='even', pa
...
+    :type new_size: (int, int)
+    :param scale_factors: Factors (fx,fy) by which to resize the image
+    :type scale_factors: (float, float)
+    :param anti_alias: Use anti aliasing smoothing operation when downsampling.
+    :type anti_alias: bool
+    :param interpolation: Interpolation method used for resampling. One of 'nearest', 'linear', 'cubic'.
+    :type interpolation: bool
```
azupanc: typo: type of `interpolation` is `str` and not `bool`. It's also a good practice IMO to mention explicitly what the default value is for parameters that have one (like this one).

core/eolearn/core/utilities.py
```diff
@@ -475,3 +476,113 @@ def constant_pad(X, multiple_of, up_down_rule='even', left_right_rule='even', pa
...
+    if interpolation not in INTER_METHODS:
+        raise ValueError('Invalid interpolation method: %s' % interpolation)
+
+    interpolation_method = INTER_METHODS[interpolation]
+    downscaling = scale_factors[0] < 1 or scale_factors[1] < 1
+
+    def _resize2d(image):
+        # Perform anti-alias smoothing if downscaling
+        if downscaling and anti_alias:
+            sigmaX, sigmaY = [((1/s) - 1)/2 for s in scale_factors]
```
azupanc: Is it possible to give a reference to a source where the motivation and/or explanation is given why `sigma = ((1/s)-1)/2` is good option?

mask/eolearn/mask/cloud_mask.py
```diff
@@ -156,15 +156,13 @@ def _downscaling(self, hr_array, meta_info, interp='linear', smooth=True):
         if (self.cm_size_y is None) and (self.cm_size_x is None):
             return hr_array, None
 
-        # Rescaling factor in spatial (height, width) dimensions
+        # Rescaling factor in spatial (width, height) dimensions
```
azupanc: Why is order of dimensions `(width, height)` and not `(height, width)`? Aren't images in EOPatches saved as `(height, width, bands)`?

core/eolearn/core/utilities.py
andrejburja: i would suggest raising an error

core/eolearn/core/utilities.py
lojzezust: This is the default value in the [skimage implementation of antialiasing](https://scikit-image.org/docs/dev/api/skimage.transform.html#skimage.transform.resize).

Here is an example using the old sigma values (sigma = 1/s). Band B02 is resized to the same resolution as B01 and compared with it. As you can see the blurring is excessive.

orig_b01

Here is an example using the skimage default sigma.

skimage_b01

core/eolearn/core/utilities.py
lojzezust: It's actually width and height. Even though images are stored as (height, width) matrices, most image processing libraries refer to the size as width, height as it's more natural.

It can be changed of course. Both options have pros and cons and in the end you always have to look at the documentation to see which abstraction is used.

mask/eolearn/mask/cloud_mask.py
lojzezust: See answer above.

core/eolearn/core/utilities.py
lojzezust: Done

core/eolearn/core/utilities.py
azupanc: I agree that this depends on the convention. Since in EOPatch the imagery is stored as `(height, width)` I would recommend that we follow this convention everywhere in our code.

core/eolearn/core/utilities.py
azupanc: Thanks for this info. Could you please add a comment to this code block describing from where this formula is taken?

core/eolearn/core/utilities.py
lojzezust: Sure. Will change.

lojzezust: That applies to scale factors as well, right? `(fx, fy)` -> `(fy, fx)`

core/eolearn/core/utilities.py
```diff
@@ -7,6 +7,7 @@
 from collections import OrderedDict
 
 import numpy as np
+import cv2
```
AleksMat: Unfortunately OpenCV is not a dependency of `eo-learn-core` and I'm not sure if it should be.

Suggestion:
- This utility is for now moved to `eo-learn-mask`. In it will be needed by other subpackages in the future we can move it to core.
- Dependency `opencv-contrib-python-headless` is added to `eo-learn-mask`

core/eolearn/core/utilities.py
```diff
@@ -7,6 +7,7 @@
 from collections import OrderedDict
```
AleksMat: There are quite a few pylint issues in module `eolearn.core.utilities`.

Issues: https://travis-ci.org/sentinel-hub/eo-learn/jobs/568268544#L1274

PR #112, opened 2019-07-22
Bugfix/vector2raster empty

mlubej:
Fixed bug in Vector2Raster, added unittest for testing the case of empty vector data (should return an empty mask)

### DIFF
--- a/geometry/eolearn/geometry/transformations.py
+++ b/geometry/eolearn/geometry/transformations.py
@@ -200,11 +200,15 @@ class VectorToRaster(EOTask):
         if eopatch.bbox is None:
             raise ValueError('EOPatch has to have a bounding box')
 
+        feature_type, feature_name = self.raster_feature
+        height, width = self._get_raster_shape(eopatch)
+
         rasterization_shapes = self._get_rasterization_shapes(eopatch)
         if not rasterization_shapes:
+            eopatch[feature_type][feature_name] = \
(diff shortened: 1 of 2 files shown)


geometry/eolearn/geometry/transformations.py
```diff
@@ -200,19 +200,22 @@ def execute(self, eopatch):
         if eopatch.bbox is None:
             raise ValueError('EOPatch has to have a bounding box')
 
+        feature_type, feature_name = self.raster_feature
+        height, width = self._get_raster_shape(eopatch)
+
         rasterization_shapes = self._get_rasterization_shapes(eopatch)
         if not rasterization_shapes:
+            eopatch[feature_type][feature_name] = \
+                np.full((height, width), self.no_data_value)[..., np.newaxis].astype(self.raster_dtype)
```
AleksMat: Nicer and quicker way: `np.full((height, width, 1), self.no_data_value, dtype=self.raster_dtype)`

geometry/eolearn/geometry/transformations.py
```diff
@@ -200,19 +200,22 @@ def execute(self, eopatch):
         if eopatch.bbox is None:
             raise ValueError('EOPatch has to have a bounding box')
 
+        feature_type, feature_name = self.raster_feature
```
AleksMat: In the new version of eo-learn you don't have to do this anymore. You can simply use `eopatch[self.raster_feature]` instead of `eopatch[feature_type][feature_name]`

PR #114, opened 2019-07-26
Added ValueFilloutTask

iovsn:


features/eolearn/tests/test_feature_manipulation.py
```diff
@@ -43,6 +43,52 @@ def test_content_after_timefilter(self):
...
+
+        # nothing to be filled, return the same eopatch object immediately
+        eopatch_new = ValueFilloutTask(feature, operations='fb', axis=0)(eopatch)
+        self.assertEqual(eopatch, eopatch_new)
+
+        eopatch[feature][0, 0, 0, :] = np.nan
+
+        # filling forward temporally should not fill nans
+        eopatch = ValueFilloutTask(feature, operations='f', axis=0)(eopatch)
+        self.assertTrue(np.isnan(eopatch[feature][0, 0, 0, :]).all())
```
AleksMat: A few of these tests could be improved. Here `ValueFilloutTask` shouldn't change the feature at all. So you could even check that an array before this task and an array after it are completely the same. Similarly you could do in the next test. On the other hand you also might have to check that you don't compare the same object. If I'm not mistaken the task creates a new array object in the process.

features/eolearn/tests/test_feature_manipulation.py
```diff
@@ -43,6 +43,52 @@ def test_content_after_timefilter(self):
...
+
+        # filling nans backwards temporally should fill nans
+        eopatch = ValueFilloutTask(feature, operations='b', axis=0)(eopatch)
+        self.assertFalse(np.isnan(eopatch[feature]).any())
+
+        # try filling something else than nan (e.g.: -1)
+        eopatch[feature][0, :, 0, 0] = -1
+        eopatch = ValueFilloutTask(feature, operations='b', value=-1, axis=-1)(eopatch)
+        # !!
+        self.assertFalse(np.isnan(eopatch[feature]).any())
```
AleksMat: Did you want to check here that there are no -1 values? This check for nan values doesn't hurt, but the main thing here should be to check if all -1 values have been replaced.

iovsn:
Tests are now updated.

PR #116, opened 2019-07-30
Feature/osm via overpass

drewbo:
Initial implementation of pulling vector data from OpenStreetMap (OSM) via the overpass API. Please review for accuracy of implementation and code style. Here is is shown alongside `VectorToRaster` for visualization purposes. Currently each new OSM filter will require a separate feature (could in the future be combined with `VectorToRaster` `values_column` parameter for a single task to create multiple classes)

Example:

```
from eolearn.io import OSMInput
from eolearn.core import FeatureType, EOPatch
from eolearn.geometry import VectorToRaster

osm_task = OSMInput('meadow', 'way["landuse"="meadow"]')
patch = EOPatch.load('example_data/TestEOPatch/')

osm_task.execute(patch)

vtr = VectorToRaster((FeatureType.VECTOR_TIMELESS, 'meadow'), (FeatureType.MASK_TIMELESS, 'meadow_mask'), values=255, raster_shape=(101, 100))
vtr.execute(patch)

plt.imshow(patch.mask_timeless['meadow_mask'][:,:,0])
plt.show()
```

Screen Shot 2019-07-30 at 3 21 39 PM


drewbo:
Noting that there is an optional parameter `polygonize` to handle whether OSM ways are converted to polygons or note. Here is how that looks in the case of roads:

```
osm_task = OSMInput('road', 'way["highway"="unclassified"]', polygonize=True)
```

Screen Shot 2019-07-30 at 4 30 51 PM


```
osm_task = OSMInput('road', 'way["highway"="unclassified"]', polygonize=False)
```

Screen Shot 2019-07-30 at 4 31 19 PM

Because the OSM data is stored as a geopandas dataframe (vector), you can make manual changes, like buffering prior to rasterizing

```
buffered_road = patch.vector_timeless['road'].buffer(10)
patch.add_feature(FeatureType.VECTOR_TIMELESS, 'road_buffered', buffered_road)
...
```

Screen Shot 2019-07-30 at 4 43 05 PM

iovsn:
Hi @drewbo, your PR would be merged into develop, not master branch (I've already edited your PR's target branch), so could you merge our latest develop branch into your branch, so we can proceed with the review?

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,71 @@
...
+        if not eopatch.bbox:
+            raise ValueError('Each EOPatch requires a bbox to fetch data')
+
+        # handling for various bounds variables
+        ll_bounds = eopatch.bbox.transform(CRS.WGS84)
+        clip_shape = box(*ll_bounds)
+        osm_bbox = tuple([*ll_bounds.reverse()])
+
+        # make the overpass request
+        response = self.api.get(f'{self.query}{osm_bbox}', verbosity='geom')
```
iovsn: We have to be python 3.5 compatible so f strings are not allowed in the eo-learn package, the build for 3.5 will fail.

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,71 @@
...
+    In case of timeouts or too many requests against the main Overpass endpoint, find additional
+    endpoints at see other options https://wiki.openstreetmap.org/wiki/Overpass_API#Public_Overpass_API_instances
+
+    :param feature_name: EOPatch feature into which data will be imported
+    :type feature_name: (FeatureType, str)
+    :param query: Overpass API Querystring: https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL
+    :type query: str
+    :param polygonize: Whether or not to treat ways as polygons, defaults to True
+    :type polygonize: bool
+    :param overpass_opts: Options to pass to the Overpass API constructor, see: https://github.com/mvexel/overpass-api-python-wrapper#api-constructor
```
iovsn: Code should pass pylint check, otherwise travis build will fail.
pylint: io/eolearn/io/osm.py:24:0: C0301: Line too long (149/120) (line-too-long)

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,71 @@
...
+    :type overpass_opts: dict
+    """
+
+    def __init__(self, feature_name, query, polygonize=True, overpass_opts={}):
```
iovsn: pylint: io/eolearn/io/osm.py:28:4: W0102: Dangerous default value {} as argument (dangerous-default-value)

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # clip geometries to bounding box
+        for feat in response['features']:
+            geom = shape(feat['geometry'])
+            if self.polygonize:
+                geom = geom.convex_hull
+            clipped_geom = geom.intersection(clip_shape)
+            feat['geometry'] = mapping(clipped_geom)
+
+
```
iovsn: Remove an excessive newline.

drewbo:
@iovsn thanks for the feedback. I linted and added a unit test to compare roads found in the example patch area. The unit test currently hits the actual Overpass API which is a bit of a no-no for testing purposes but I didn't see a API testing library in place yet so I left it as is for now. This also risks the underlying OSM data changing and failing the test (since the fixture won't change) but I'm going to just flag for now

iovsn:
Hi @drewbo. Travis now fails because of import order in osm.py, could you check that please?

We have also check in more detail the implementation of polygonize=True scenario, where you use geom.convex_hull. Is there another way of polygonizing geomentry? Convex hull changes the geometry and we don't think this is the appropriate way of solving invalid geometries that you'd otherwise get.

drewbo:
Thanks for the feedback @iovsn, I've resolved the linting/import error.

Do you have any thoughts on other ways to resolve the geometries? As an example, building outlines are stored as OSM ways which return as LineStrings from the Overpass API. Then rasterization of them appears like this:

Screen Shot 2019-12-02 at 11 42 28 PM

devisperessutti:
What users probably expect is a binary mask where 1s denote the class pixel-wise, not only the edge.

The issue with convex hull is that it alters the original goemetry.

drewbo:
Thanks for help @iovsn @devisperessutti ; the API doesn't support returning relations directly but they [provide some examples](https://github.com/mvexel/overpass-api-python-wrapper/issues/106#issuecomment-421446118) of wrapping the response with `shapely` to produce a valid polygon (concave corners and all). I've updated accordingly and also set the default of `polygonize` to `False`

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,75 @@
...
+    :param query: Overpass API Querystring: https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL
+    :type query: str
+    :param polygonize: Whether or not to treat ways as polygons, defaults to True
+    :type polygonize: bool
+    :param overpass_opts: Options to pass to the Overpass API constructor,
+    see: https://github.com/mvexel/overpass-api-python-wrapper#api-constructor
+    :type overpass_opts: dict
+    """
+
+    def __init__(self, feature_name, query, polygonize=False, overpass_opts=None):
```
devisperessutti: change `feature_name` to `feature` for consistency with other tasks (please update docstring as well)

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def execute(self, eopatch):
+        """ Execute function which adds new VECTOR_TIMELESS layer to the EOPatch
+        :param eopatch: input EOPatch
+        :type eopatch: EOPatch
+        :return: New EOPatch with added VECTOR_TIMELESS layer
+        :rtype: EOPatch
+        """
+
+        if self.feature_name is None:
```
devisperessutti: this should not be necessary anymore

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,75 @@
...
+    :type query: str
+    :param polygonize: Whether or not to treat ways as polygons, defaults to True
+    :type polygonize: bool
+    :param overpass_opts: Options to pass to the Overpass API constructor,
+    see: https://github.com/mvexel/overpass-api-python-wrapper#api-constructor
+    :type overpass_opts: dict
+    """
+
+    def __init__(self, feature_name, query, polygonize=False, overpass_opts=None):
+        self.feature_name = feature_name
```
devisperessutti: `self.feature = next(self._parse_features(feature)())`

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,75 @@
...
+                # https://github.com/mvexel/overpass-api-python-wrapper/issues/48#issuecomment-418770376
+                geom = Polygon(geom)
+            clipped_geom = geom.intersection(clip_shape)
+            feat['geometry'] = mapping(clipped_geom)
+
+        # import to geopandas, transform and return
+        gdf = geopandas.GeoDataFrame.from_features(response['features'])
+        gdf.crs = {'init' :'epsg:4326'} # all Osmium data is returned with this CRS
+        gdf = gdf.to_crs({'init': eopatch.bbox.crs.ogc_string().lower()})
+        eopatch[FeatureType.VECTOR_TIMELESS][self.feature_name] = gdf
```
devisperessutti: `eopatch[self.feature] = gdf`

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # clip geometries to bounding box
+        for feat in response['features']:
+            geom = shape(feat['geometry'])
+            if self.polygonize:
+                # currently the API doesn't support relations so we use shapely to wrap geometries
+                # https://github.com/mvexel/overpass-api-python-wrapper/issues/106#issuecomment-421446118
+                # https://github.com/mvexel/overpass-api-python-wrapper/issues/48#issuecomment-418770376
+                geom = Polygon(geom)
+            clipped_geom = geom.intersection(clip_shape)
```
devisperessutti: add check on validity, e.g. `if geom.is_valid`

io/eolearn/io/osm.py
```diff
@@ -0,0 +1,75 @@
...
+            feat['geometry'] = mapping(clipped_geom)
+
+        # import to geopandas, transform and return
+        gdf = geopandas.GeoDataFrame.from_features(response['features'])
```
devisperessutti: add check on `response['features']`, return eopatch instead with user warning

devisperessutti:
Hi @drewbo,

Thank you for the changes. The following request

`osm_task = OSMInput('road', 'way["highway"="unclassified"]', polygonize=True)`

does not work due to invalid geometries.

I suggest to change code in line 67 to something like:

```python
if geom.is_valid:
    clipped_geom = geom.intersection(clip_shape)
    feat['geometry'] = mapping(clipped_geom)
```

Furthermore, if no geometry is present, the code fails abruptly. I suggest to raise a warning in such cases and return the original eopatch, something like this

```python
import logging

# current line 70 import to geopandas, transform and return
if response['features']:
    gdf = geopandas.GeoDataFrame.from_features(response['features'])
    gdf.crs = {'init' :'epsg:4326'} # all Osmium data is returned with this CRS 
    gdf = gdf.to_crs({'init': eopatch.bbox.crs.ogc_string().lower()})
    eopatch[FeatureType.VECTOR_TIMELESS][self.feature_name] = gdf
else:    
    LOGGER.warning('No valid polygon for query {} was found'.format(self.query))
         
return eopatch
```

Can you please also update the parsing of the input feature (using `_parse_features`) as suggested in the comments? Thanks!

mlubej:
Hi everyone. Since there was no activity in this PR, and since we see value in this task, I suggest to switch this merge into a secondary eo-learn branch and so that we can proceed with the fixes there ourselves.

I'll wait a bit if there are any objections, otherwise will proceed as proposed.

devisperessutti:
Closing due to inactivity

We can reopen and continue if relevant.

PR #117, opened 2019-07-30
Added TrainSetMaskTask

iovsn:


ml_tools/eolearn/ml_tools/train_set_mask.py
```diff
@@ -0,0 +1,60 @@
...
+        np.random.seed(seed)
+
+        ftype, fname, new_name = self.feature
+        data = np.copy(eopatch[(ftype, fname)])
+
+        polys = set(np.unique(data)) - {self.no_data_value}
+        rands = np.random.rand(len(polys))
+        split = np.digitize(rands, self.bins)
+
+        for poly, split_class in zip(polys, split):
```
AleksMat: This could be optimized. There are a few options [here](https://stackoverflow.com/questions/16992713/translate-every-element-in-numpy-array-according-to-key). Another one is also a recursive approach I used for superpixel postprocessing tasks.

ml_tools/eolearn/ml_tools/train_set_mask.py
```diff
@@ -0,0 +1,60 @@
...
+            data[data == poly] = split_class
+
```
AleksMat: If `no_data_value` would be the same as one of the values in `split` that would cause unwanted problems. Maybe a warning could be raised in such case?

ml_tools/eolearn/tests/test_train_set_mask.py
```diff
@@ -0,0 +1,58 @@
...
+        data[4:8, 4:8, :] = 77
+
+        patch = EOPatch()
+        patch[(FeatureType.MASK_TIMELESS, 'TEST')] = data
+        patch[(FeatureType.LABEL_TIMELESS, 'TEST')] = data.copy().reshape((size,))
+
+        bins = [0.2, 0.5, 0.8]
+
+        patch = TrainSetMaskTask((*input_mask_feature, new_name), bins)(patch, seed=1)
+        self.assertTrue(set(np.unique(patch[new_mask_feature])) <= set(range(len(bins) + 1)))
```
AleksMat: Above you could define
```Python
interval_list = [(0, 2, 0, 2), (0, 2, 2, 4) ,...]
```
use this indices from such list to set `data[i1: i2, j1: j2, :] = 11 * index`.
Then  you could use these indices again here to make sure that each of these slices has a unique value

ml_tools/eolearn/tests/test_train_set_mask.py
```diff
@@ -0,0 +1,58 @@
...
+
+        patch = TrainSetMaskTask((*input_label_feature, new_name), bins)(patch, seed=1)
```
AleksMat: If you run the same task again, it makes sense to check that the resulting feature is completely the same as the one in the previous run.

ml_tools/eolearn/ml_tools/train_set_mask.py
```diff
@@ -0,0 +1,60 @@
...
+from eolearn.core import EOTask
+
+
+class TrainSetMaskTask(EOTask):
+    """ Randomly assigns each polygon to a subset of polygons by generating a mask where each value represents a polygon
+    subset. Subsets are defined by a list of cumulative probabilities, passed as the *bins* argument, the same way as
+    the *bins* argument in `numpy.digitize <https://docs.scipy.org/doc/numpy/reference/generated/numpy.digitize.html>`_.
+    """
+
+    def __init__(self, feature, bins, no_data_value=None):
```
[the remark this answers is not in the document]
azupanc: Doesn't this defeat its purpose? Probably I'm missing something. Anyway, if it's a new feature idea that extends the functionality let's put it on our backlog.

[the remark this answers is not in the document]
AleksMat: Ok, we can leave it as it is for now. If it ever turns out this might come useful we can add it then.

ml_tools/eolearn/ml_tools/train_set_mask.py
```diff
@@ -0,0 +1,60 @@
+"""
+Tasks used for train set preparation.
+"""
+
+import numpy as np
+
+from eolearn.core import EOTask
+
+
+class TrainSetMaskTask(EOTask):
```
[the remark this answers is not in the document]
iovsn: I named it TrainSplitTask.

ml_tools/eolearn/ml_tools/train_set_mask.py
```diff
@@ -0,0 +1,60 @@
...
+    """ Randomly assigns each polygon to a subset of polygons by generating a mask where each value represents a polygon
```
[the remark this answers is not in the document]
iovsn: I've used the term 'value classes' instead of polygons.

ml_tools/eolearn/ml_tools/train_set_mask.py
```diff
@@ -0,0 +1,60 @@
...
+        :param no_data_value: A value to ignore and not assign it to any subsets.
+        :type no_data_value: any numpy.dtype
+        """
+        self.feature = next(self._parse_features(feature, new_names=True)())
+
+        if np.isscalar(bins):
+            bins = [bins]
+
+        if not isinstance(bins, list) or not all(isinstance(bi, float) for bi in bins) \
+                or np.any(np.diff(bins) <= 0) or bins[-1] >= 1:
```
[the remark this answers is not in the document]
iovsn: Done.

ml_tools/eolearn/ml_tools/train_set_mask.py
iovsn: Added a warning.

ml_tools/eolearn/ml_tools/train_split.py
```diff
@@ -0,0 +1,67 @@
+"""
+Tasks used for train set preparation.
+"""
+import numpy as np
+
+from eolearn.core import EOTask
+
+
+class TrainSplitTask(EOTask):
```
azupanc: In ML community the methods that split samples in two or more independent sets usually have`train_test` in their name. I would suggest to follow this nomenclature and suggest to rename the task to `TrainTestSplitTask`.

ml_tools/eolearn/ml_tools/train_split.py
```diff
@@ -0,0 +1,67 @@
...
+    """ Randomly assigns each value to a class of values by generating a class mask.
```
azupanc: Can the documentation be more explicit? Suggestion:

Randomly assign each or group of pixels to a train, test, and/or validation sets. The group of pixels is defined by an input feature (i.e. MASK_TIMELESS with polygon ids, or connected component id, or similar that groups pixels with similar properties). If input feature is None, than each pixel is randomly assigned to one of the target sets.

Target sets are defined by a list of ...


At the end I would add an example of definition of bins for train/test split in 80/20 and train/val/test split in 60/20/20.

azupanc: While writing above documentation it became clear to me that we don't always have polygon ids or similar feature that groups pixels. In that case it would be nice to be able to split pixels randomly. I understand that this is a new feature and now may not be the right time to implement it. I leave it up to you to decide.

ml_tools/eolearn/ml_tools/train_split.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Classes are defined by a list of cumulative probabilities, passed as the *bins* argument, the same way as the *bins*
+    argument in `numpy.digitize <https://docs.scipy.org/doc/numpy/reference/generated/numpy.digitize.html>`_. Valid
+    classes are enumerated from 1 onwards and if no_data_value is provided, all values equal to it get assigned to class
+    0.
```
azupanc: What I miss in the docs is description of _output_. In what respect will the EOPatch be different after the execution of this task.

Suggestion: After the execution of this task an EOPatch will have a new (FeatureType, new_name) feature where each pixel will have a value assigned representing the train, test and/or validation set.

ml_tools/eolearn/ml_tools/train_split.py
iovsn: Done.

ml_tools/eolearn/ml_tools/train_split.py
iovsn: Updated docs and implemented random pixel assignment to train subsets.

ml_tools/eolearn/ml_tools/train_split.py
iovsn: Done.

ml_tools/eolearn/ml_tools/train_split.py
AleksMat: > While writing above documentation it became clear to me that we don't always have polygon ids or similar feature that groups pixels. In that case it would be nice to be able to split pixels randomly. I understand that this is a new feature and now may not be the right time to implement it. I leave it up to you to decide.

@azupanc By the way, this is exactly what I suggested in [this conversation](https://github.com/sentinel-hub/eo-learn/pull/117#discussion_r312134825). But probably my explanation wasn't clear enough.

PR #118, opened 2019-08-01
Copyright statements

azupanc:
Added Copyright statements to LICENSE file and all source files. From now on whenever a new contributor contributes to the package his/her name and affiliation should be added to the LICENSE file and corresponding source file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,13 @@
 MIT License
 
-Copyright (c) 2018 Sentinel Hub
+Copyright (c) 2017-2019 Matej Aleksandrov, Matej Batič, Andrej Burja, Eva Erzin (Sinergise)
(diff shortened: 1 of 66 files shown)


AleksMat:
I had a look how to avoid parsing copyright statements into the documentation. There is a simple solution:

- First we put some marker string a line before the start of contribution statements, e.g. 'Credits:'.
- Next we put the following lines anywhere in `docs/source/conf.py` file
```Python
import sphinx.ext.autodoc

def setup(app):
    app.connect('autodoc-process-docstring', sphinx.ext.autodoc.between('Credits:', what=['module'], exclude=True))
```

This way Sphinx will automatically exclude any text between appearance of a string 'Credits:' and the next appearance of the same string. Since there won't be a next appearance of this string it will exclude all text until the end of the docs string. With parameter `what` it is specified that this operation is only applied on any module docs string and not on a class, method or function docs string.

azupanc:
Thanks. This has been done and the docs look fine.

AleksMat:
Looks good, merging to develop

PR #123, opened 2019-08-16
Fixes feature parsing in MoveTask. 

azupanc:
For example, using ellipsis ... to indicate all features of certain type to be moved did not work. This is fixed now.

### DIFF
--- a/core/eolearn/tests/test_core_tasks.py
+++ b/core/eolearn/tests/test_core_tasks.py
@@ -224,6 +224,17 @@ class TestCoreTasks(unittest.TestCase):
             self.assertTrue(id(data[i]) != id(patch_dst[feature]))
             self.assertTrue(np.array_equal(data[i], patch_dst[feature]))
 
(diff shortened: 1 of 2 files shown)


core/eolearn/tests/test_core_tasks.py
```diff
@@ -224,6 +224,17 @@ def test_move_feature(self):
             self.assertTrue(id(data[i]) != id(patch_dst[feature]))
             self.assertTrue(np.array_equal(data[i], patch_dst[feature]))
 
+        features = [(FeatureType.MASK_TIMELESS, ...)]
+        patch_dst = EOPatch()
+        patch_dst = MoveFeature(features)(patch_src, patch_dst)
+
+        self.assertTrue(FeatureType.MASK_TIMELESS in patch_dst.get_features().keys())
```
AleksMat: You don't have to write `.keys()`. In Python 3 dictionaries are by default iterated by their keys.

PR #59, opened 2019-01-24
Snow Mask Tasks

eerzin:

PR #128, opened 2019-08-21
Add refactored snow masking tasks

devisperessutti:
PR that re-factors and harmonizes the code contributed in PR #59 .

Two snow-masking algorithms are added:

 * a simple thresholding of `NDSI`
 * a re-implementation of the Theia snow masking product.

Default thresholds and parameters can be adjusted to work with L1C and L2A products.


devisperessutti:
one thing that was not done yet is to harmonize the resizing method to be consistent with the one used in cloud masking (`opencv` vs `Pillow`)

mask/eolearn/mask/snow_mask.py
```diff
@@ -0,0 +1,261 @@
...
+
+        snow_mask = np.where(np.logical_and(np.logical_or(ndsi >= self.ndsi_threshold,
+                                                          np.abs(ndvi - self.NDVI_THRESHOLD) < self.NDVI_THRESHOLD / 2),
+                                            bands[..., 0] >= self.brightness_threshold), 1, 0)
+
+        snow_mask = self._apply_dilation(snow_mask)
+
+        snow_mask[np.logical_or(ndsi_invalid, ndvi_invalid)] = self.undefined_value
+
+        eopatch.add_feature(FeatureType.MASK, self.mask_name, snow_mask[..., np.newaxis].astype(np.uint8))
```
AleksMat: 3 comments here:
- This could be written in the new notation, either  `eopatch.mask[feature_type][feature_name] = ...` or  `eopatch.mask[snow_mask_feature] = ...`
- As I understand this task is doing single temporal classification. Therefore it could either work on `FeatureType.DATA` or `FeatureType.DATA_TIMELESS`. Up to this point either feature type would work. Here (or even better - already in `__init__`) the task could decide to store results in either `FeatureType.MASK` of `FeatureType.MASK_TIMELESS`.
- Does the mask have to be cast to `np.uint8`? Since it is a binary mask it seems more natural to store it in a boolean type. `AddCloudMaskTask` is also returning mask values in boolean.

I think everything here also applies to the same case in `TheiaSnowMask` task. 
(Even the second bullet should be valid as you can at the beginning add temporal dimension to timeless data.)

mask/eolearn/mask/snow_mask.py
```diff
@@ -0,0 +1,261 @@
...
+        :return: resampled values
+        """
+        size = np.array(input_array.shape[1:]) // self.red_params[0]
+        resampled = [np.array(Image.fromarray(frame).resize(size, Image.BICUBIC).resize(frame.shape, Image.NEAREST))
+                     for frame in input_array]
+        return np.array(resampled)
+
+    def _adjust_cloud_mask(self, bands, cloud_mask, dem, b10):
+        """ Adjust existing cloud mask using cirrus band if L1C data and resampled red band """
+        # add to the existing cloud mask pixels found thresholding down-sampled red band and cirrus band/DEM
```
AleksMat: This comment could probably go in a docstring 1 line above.

mask/eolearn/mask/snow_mask.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def execute(self, eopatch):
+        raise NotImplementedError
+
+
+class SnowMask(BaseSnowMask):
+    """
+    The task calculates the snow mask using the given thresholds.
+
+    THe default values were optimised based on the Sentinel-2 L1C processing level. Values might not be optimal for L2A
```
AleksMat: Typo in the first word "THe"

mask/eolearn/mask/snow_mask.py
```diff
@@ -0,0 +1,261 @@
...
+        return snow_mask_pass1, snow_frac, dem_edges
+
+    def _apply_second_pass(self, bands, ndsi, dem, clm_temp, snow_mask_pass1, snow_frac, dem_edges):
+        """ Second pass of snow detection """
+        _, height, width, _ = bands.shape
+        total_snow_frac = np.sum(snow_mask_pass1, axis=(1, 2)) / (height*width)
+        snow_mask_pass2 = np.zeros(snow_mask_pass1.shape)
+        for date in range(bands.shape[0]):
+            if total_snow_frac[date] > self.ndsi_params[2]:
+                if snow_frac is not None and np.any(snow_frac[date] > self.dem_params[1]):
```
AleksMat: Probably these 2 `if` sentences could be joined into a single one?

mask/eolearn/mask/snow_mask.py
```diff
@@ -0,0 +1,261 @@
...
+        """ Method to resample the values of the red band
+
+        The input array is first down-scaled using bicubic interpolation and up-scaled back using nearest neighbour
+        interpolation
+
+        :param input_array: input values
+        :return: resampled values
+        """
+        size = np.array(input_array.shape[1:]) // self.red_params[0]
+        resampled = [np.array(Image.fromarray(frame).resize(size, Image.BICUBIC).resize(frame.shape, Image.NEAREST))
```
AleksMat: Now that new resizing utility (`eolearn.mask.utilities.resize_images`) has been merge to develop, could it be used here?

mask/eolearn/mask/snow_mask.py
devisperessutti: fixed typo

mask/eolearn/mask/snow_mask.py
devisperessutti: added to docstring

mask/eolearn/mask/snow_mask.py
devisperessutti: changed pillow with `resize_images` method

mask/eolearn/mask/snow_mask.py
devisperessutti: changed returned type to `bool`. The `IS_DATA` array is `uint8`, so we should unify all binary masks type to be `bool`.

Changed initialisation of a `snow_feature` in `__init__`.

I think the task should work on `DATA` and return `MASK` only, to avoid confusions. If a single time frame is provided, the first dimension of the output will have size 1. This is how the cloud masking works, and I don't think we should change it.

mask/eolearn/mask/snow_mask.py
devisperessutti: Changed as suggested

mask/eolearn/mask/snow_mask.py
AleksMat: At some point we'll have to update `IS_DATA` dtype as well. Probably when we'll be updating `eo-learn-io` with new SH API.

I guess we can leave snow masking limited to `DATA` features for now. Some time in the future we'll have to unify cloud and snow masking tasks. At that point we can also think about such extensions.

PR #132, opened 2019-08-28
Added constant to the NDI calculation

iovsn:


features/eolearn/features/bands_extraction.py
```diff
@@ -47,17 +47,21 @@ def map_method(self, feature):
 class NormalizedDifferenceIndexTask(MapFeatureTask):
     """ The task calculates a Normalized Difference Index (NDI) between two bands A and B as:
 
-        :math:`NDI = \\dfrac{A-B}{A+B}`
+        :math:`NDI = \\dfrac{A-B+c}{A+B+c}`,
+
+        where c is provided as the *constant* argument.
```
AleksMat: Here it is very important to explain that this constant has a scientific explanation: http://www.cesbio.ups-tlse.fr/multitemp/?p=12746 . Otherwise one might thing it was added just for our convenience.
Therefore I suggest adding a bit more explanation here and the upper link.

features/eolearn/features/bands_extraction.py
```diff
@@ -47,17 +47,21 @@ def map_method(self, feature):
...
     """
 
-    def __init__(self, input_feature, output_feature, bands, undefined_value=np.nan):
+    def __init__(self, input_feature, output_feature, bands, constant=0, undefined_value=np.nan):
```
AleksMat: Would it make sense rename `constant` to `acorvi_constant`? This way users might get some context what this constant represents.

features/eolearn/features/bands_extraction.py
iovsn: Done.

features/eolearn/features/bands_extraction.py
iovsn: Done.

PR #134, opened 2019-09-02
Bugfix: eotask execution exceptions

iovsn:
EOTask execution wasn't working properly for exceptions that didn't have the expected __init__ signature (accepting only errmsg parameter). When such an exception occurred it wasn't clear where exactly it happened, because another exception was raised when trying to instantiate such an exception. I fixed it by avoiding instantiation of a new object through type(exception)(...) and instead I just change the errmsg of the existing exception object.

### DIFF
--- a/core/eolearn/core/eotask.py
+++ b/core/eolearn/core/eotask.py
@@ -68,21 +68,19 @@ class EOTask(ABC):
         """
         self.private_task_config.start_time = datetime.datetime.now()
 
-        caught_exception = None
         try:
             return_value = self.execute(*eopatches, **kwargs)
+            self.private_task_config.end_time = datetime.datetime.now()
+            return return_value
         except BaseException as exception:
-            caught_exception = exception, sys.exc_info()[2]
+            exception, traceback = exception, sys.exc_info()[2]
 
-        if caught_exception is not None:  # Exception is not raised in except statement to prevent duplicated traceback
-            exception, traceback = caught_exception
-
-            exception = exception.with_traceback(traceback)
-            exception.errmsg = 'During execution of task {}: {}'.format(self.__class__.__name__, exception)
-            raise exception
-
-        self.private_task_config.end_time = datetime.datetime.now()
-        return return_value
+            # Some special exceptions don't accept error message as a parameter and a TypeError is raised in such case.
+            try:
+                errmsg = 'During execution of task {}: {}'.format(self.__class__.__name__, exception)
+                raise type(exception)(errmsg).with_traceback(traceback)
+            except TypeError:
+                raise exception
 
     @abstractmethod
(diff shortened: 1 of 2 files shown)


core/eolearn/core/eotask.py
```diff
@@ -68,19 +68,19 @@ def _execute_handling(self, *eopatches, **kwargs):
...
-                                                                           exception)).with_traceback(traceback)
-
-        self.private_task_config.end_time = datetime.datetime.now()
-        return return_value
+            exception, traceback = exception, sys.exc_info()[2]
+
+            # Some special exceptions don't accept error message as a parameter and a TypeError is raised in such case.
+            try:
+                errmsg = 'During execution of task {}: {}'.format(self.__class__.__name__, exception)
+                raise type(exception)(errmsg).with_traceback(traceback)
```
AleksMat: There is an edge case if `exception` is already a `TypeError`. In such case this wouldn't produce an expected result. I suggest a slight restructuring:
```Python
try:
    errmsg = 'During execution of task {}: {}'.format(self.__class__.__name__, exception)
    improved_exception = type(exception)(errmsg)
except TypeError:
    improved_exception = exception

raise improved_exception.with_traceback(traceback)
```
This way the internal try-catch will only catch `TypeError` obtained from exception initialization and not from raising the final exception.

core/eolearn/tests/test_eotask.py
```diff
@@ -44,6 +72,20 @@ def test_chained(self):
         for i in range(5):
             self.assertEqual(composite(i), 6 * i + 9)
 
+    def test_execution_handling(self):
+        task = ExceptionTestingTask('test_exception')
+        self.assertRaises(TestException, task, 'test')
```
AleksMat: At least 1 of these tests should also verify that an error message has been edited
```Python
try:
    task('test')
except TestException as exception:
    message = str(exception)
    self.assertTrue(message.startswith('During execution of task TestException: '))
```

AleksMat: Ah, the upper code is not ok. But this would work:

```Python
for parameter, exception_type in [('test_exception_fail', TypeError), ('value_error', ValueError)]:
    task = ExceptionTestingTask(parameter)
    self.assertRaises(exception_type, task, 'test')
    try:
        task('test')
    except exception_type as exception:
        message = str(exception)
        self.assertTrue(message.startswith('During execution of task ExceptionTestingTask: '))
```

core/eolearn/core/eotask.py
iovsn: Good one, I agree.

core/eolearn/tests/test_eotask.py
iovsn: Done.

PR #136, opened 2019-09-03
Possibility to reproject to new crs for ExportToTiff

wouellette:
Hi,

It's a simple addition but a very useful one, especially when processing large areas using the newly introduced `UTMZoneSplitter` and `UTMGridSplitter` which are combining various UTM projections within one `AreaSplitter`. Therefore, reprojecting `EOPatch` features to a global `CRS` like `EPSG:4326` or `EPSG:3857` is valuable.


io/eolearn/io/local_io.py
```diff
@@ -179,24 +180,50 @@ def execute(self, eopatch, *, filename=None):
...
+                               dtype=image_dtype, nodata=self.no_data_value,
+                               transform=dst_transform, crs=dst_crs) as dst:
+                output_array = array_sub.astype(image_dtype)
+                output_array = np.moveaxis(output_array, -1, 1).reshape(index, width, height)
+                dst.write(output_array)
+        else:
+            src_crs = {'init': CRS.ogc_string(eopatch.bbox.crs)}
+            dst_crs = {'init': CRS.ogc_string(self.crs)}
+            src_transform = rasterio.transform.from_bounds(*eopatch.bbox, width=width, height=height)
+            dst_transform, width, height = rasterio.warp.calculate_default_transform(
```
devisperessutti: one minor comment, I would rename `width` and `height` to `dst_width` and `dst_height` to avoid confusion with original dataset. 

Similarly, on line 212 `width=dst_width` and `height=dst_height`, and on line 217 `.reshape(index, height, width)`

devisperessutti:
Thanks for the changes William, I've realised that now pylint complains about too many variables declared, which is probably why you over-wrote the `height` and `width` parameters in the first place. Sorry about that.

We can either revert the changes, remove some variables, or deactivate the pylint check for this task. Perhaps @AleksMat can suggest which option is preferred.

AleksMat:
I had a look at the code, rearranged some parts to avoid code duplications, merged the latest develop into this branch and rearranged the code some more. Now the pylint is passing successfully.

@wouellette I will now make a PR of my changes into your branch. Let me know if you have any comments and suggestions. Otherwise, I think it is ready to merge.

PR #141, opened 2019-09-16
Added stream-temporal feature extraction task from D4.6.

jO-Osko:
Added stream-temporal feature extraction task from D4.6.

### DIFF
--- a/features/eolearn/features/temporal_features.py
+++ b/features/eolearn/features/temporal_features.py
@@ -14,7 +14,9 @@ import itertools as it
 
 import numpy as np
 
-from eolearn.core import EOTask
+from eolearn.core import EOTask, FeatureType
+
+from eolearn.ml_tools.utilities import rolling_window
 
 
 class AddSpatioTemporalFeaturesTask(EOTask):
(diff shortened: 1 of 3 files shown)


features/eolearn/features/temporal_features.py
```diff
@@ -243,3 +245,369 @@ def execute(self, eopatch):
...
+        eopatch.data_timeless[self.pos_transition_feature] = data_pos_tr[..., np.newaxis]
+
+        eopatch.data_timeless[self.neg_len_feature] = data_neg_len[..., np.newaxis]
+        eopatch.data_timeless[self.neg_surf_feature] = data_neg_surf[..., np.newaxis]
+        eopatch.data_timeless[self.neg_rate_feature] = data_neg_rate[..., np.newaxis]
+        eopatch.data_timeless[self.neg_transition_feature] = data_neg_tr[..., np.newaxis]
+
+        return eopatch
+
+    def get_data(self, patch):
```
iovsn: I don't think this function belongs here, because it just serves your particular case of data arrangement. All tasks get an eopatch, process it and return an eopatch. I think that here you convert the data into a form that you need in your particular application. Perhaps this data re-arrangement should also happen there in the app instead of in eo-learn, which should be more general.

features/eolearn/features/temporal_features.py
```diff
@@ -243,3 +245,369 @@ def execute(self, eopatch):
...
+        """
+        # pylint: disable=too-many-locals
+        self.data_feature = next(iter(self._parse_features(data_feature, default_feature_type=FeatureType.DATA)))
+        self.data_index = data_index or 0
+        self.mask_data = mask_data
+        self.ndvi_feature_name = next(iter(self._parse_features(ndvi_feature_name,
+                                                                default_feature_type=FeatureType.DATA)))
+
+        if feature_name_prefix:
+            self.feature_name_prefix = feature_name_prefix
```
iovsn: This case doesn't get covered by the unit test.

features/eolearn/features/temporal_features.py
```diff
@@ -243,3 +245,369 @@ def execute(self, eopatch):
...
+
+        all_dates = np.asarray([x.toordinal() for x in eopatch.timestamp])
+
+        data = eopatch[self.data_feature[0]][self.data_feature[1]][..., self.data_index]
+        valid_data_mask = np.ones_like(data)
+
+        if data.ndim == 3:
+            _, h, w = data.shape
+        else:
+            raise ValueError('{} feature has incorrect number of dimensions'.format(self.data_feature))
```
iovsn: Raising the ValueError doesn't get covered by the unit test.

mlubej:
Hello everyone.

I think @iovsn showed nicely and objectively that this PR does too much for a single task and also does tasks which could/should be handled by other existing tasks.

@jO-Osko we thank you for your contribution and we are happy to see such activities. We still see value in the ideas which @iovsn extracted from your PR, so if you are willing to implement those, we would be grateful and ask you to do this in a new PR.

Otherwise, I suggest closing this PR. I'll wait for any objections.

mlubej:
Closing.

jO-Osko:
I agree with the proposal.

PR #142, opened 2019-09-17
Multitemporal cloud detection task

lojzezust:
Ports `AddMultiCloudMaskTask` from CMIX task for use in eo-learn.

Notebook with usage examples and notes is available [here](https://git.sinergise.com/eo/unsorted/cmix/blob/master/cmix/public/interssim_turkey_demo.ipynb).

### DIFF
--- a/mask/eolearn/mask/cloud_mask.py
+++ b/mask/eolearn/mask/cloud_mask.py
@@ -321,7 +321,7 @@ class AddTwinCloudMaskTask(EOTask):
         - before thresholding, probabilities are averaged with a disk of size 1
         - after thresholding, the mask is dilated with a disk of size 1
     """
-    
+
     def __init__(self,
                  mono_classifier=None,
                  multi_classifier=None,
@@ -341,71 +341,71 @@ class AddTwinCloudMaskTask(EOTask):
                  average_over=1,
                  dilation_size=1
                 ):
-        
+
         # Load classifiers
         if mono_classifier is None or multi_classifier is None:
             classifier_dir = os.path.dirname(__file__)
-            
+
             if mono_classifier is None:
                 mono_classifier = joblib.load(os.path.join(classifier_dir, MONO_CLASSIFIER_NAME))
-            
+
             if multi_classifier is None:
                 multi_classifier = joblib.load(os.path.join(classifier_dir, MULTI_CLASSIFIER_NAME))
 
         self.mono_classifier = mono_classifier
         self.multi_classifier = multi_classifier
-                
+
         # Set data info
         self.data_feature = data_feature
         self.is_data_feature = is_data_feature
         self.band_indices = (0,1,3,4,7,8,9,10,11,12) if all_bands else tuple(range(10))
-        
+
         # If only resolution of the source is specified, the sigma alone can be adjusted
         if src_res is not None and proc_res is None:
-            
+
             src_res_ = src_res if type(src_res) == int else int(re.match('\d+', src_res).group())
-            
+
             self.sigma = 100. / src_res_
             self.scale_factors = None
-            
+
         # If both resolution of the source and mid-product is specified, resizing is taken into account
         elif src_res is not None and proc_res is not None:
-            
+
             src_res_ = src_res if type(src_res) == int else int(re.match('\d+', src_res).group())
             proc_res_ = proc_res if type(proc_res) == int else int(re.match('\d+', proc_res).group())
-            
+
             self.sigma = 100. / proc_res_
             self.scale_factors = (src_res_ / proc_res_,)*2
-            
+
         # In any other case, no resizing is performed and sigma is set to a non-volatile level
         else:
-            
+
             self.sigma = 1.0
             self.scale_factors = None
-        
+
         # Set max frames for single iteration
         self.max_proc_frames = max_proc_frames
-        
+
         # Set feature info
         self.mono_proba_feature = mono_proba_feature
         self.multi_proba_feature = multi_proba_feature
         self.mono_mask_feature = mono_mask_feature
         self.multi_mask_feature = multi_mask_feature
         self.mask_feature = mask_feature
-        
+
         # Set thresholding and morph. ops. parameters and kernels
         self.mono_threshold = mono_threshold
         self.multi_threshold = multi_threshold
-        
+
         self.average_over = average_over
         self.dilation_size = dilation_size
-        
+
         if average_over is not None and average_over > 0:
             self.avg_kernel = disk(average_over) / np.sum(disk(average_over))
-            
+
         if dilation_size is not None and dilation_size > 0:
             self.dil_kernel = disk(dilation_size).astype(np.uint8)
-            
+
     @staticmethod
     def _get_max(x):
         return np.ma.max(x, axis=0).data
@@ -421,7 +421,7 @@ class AddTwinCloudMaskTask(EOTask):
     @staticmethod
     def _get_std(x):
         return np.ma.std(x, axis=0).data
-    
+
     def _frame_indices(self, num_of_frames, target_idx):
         """
         Returns frame indices within a given time window, with the target index relative to it.
@@ -442,25 +442,25 @@ class AddTwinCloudMaskTask(EOTask):
         nt_rel = target_idx - nt_min
 
         return nt_min, nt_max, nt_rel
-    
+
     def _red_ssim(self, x, y, valid_mask, mu1, mu2, sigma1_2, sigma2_2, c1=1e-6, c2=1e-5):
         """
         Slightly reduced (pre-computed) SSIM computation.
         """
-        
+
         # Increase precision and mask invalid regions
         valid_mask = valid_mask.astype(np.float64)
         x = x.astype(np.float64) * valid_mask
         y = y.astype(np.float64) * valid_mask
-        
-        # Init        
+
+        # Init
         mu1_2 = mu1 * mu1
         mu2_2 = mu2 * mu2
         mu1_mu2 = mu1 * mu2
-        
+
         sigma12 = cv2.GaussianBlur((x*y).astype(np.float64), (0,0), self.sigma, borderType=cv2.BORDER_REFLECT)
         sigma12 -= mu1_mu2
-        
+
         # Formula
         tmp1 = 2. * mu1_mu2 + c1
         tmp2 = 2. * sigma12 + c2
@@ -471,36 +471,36 @@ class AddTwinCloudMaskTask(EOTask):
         den = tmp1 * tmp2
 
         return np.divide(num, den)
-    
+
     def _win_avg(self, x):
         """
         Spatial window average.
         """
-        
+
         return cv2.GaussianBlur(x.astype(np.float64), (0,0), self.sigma, borderType=cv2.BORDER_REFLECT)
-    
+
     def _win_prevar(self, x):
         """
         Incomplete spatial window variance.
         """
-        
+
         return cv2.GaussianBlur((x*x).astype(np.float64), (0,0), self.sigma, borderType=cv2.BORDER_REFLECT)
-    
+
     def _resize(self, x):
         downscaling = self.scale_factors[0] < 1 or self.scale_factors[0] < 1
         old_size = (x.shape[1], x.shape[0])
         new_size = tuple([int(d * f) for d,f in zip(old_size, self.scale_factors)])
-        
+
         # Perform anti-alias smoothing if downscaling
         if downscaling:
             sx, sy = [((1/s) - 1)/2 for s in self.scale_factors]
             x = cv2.GaussianBlur(x, (0,0), sigmaX=sx, sigmaY=sy, borderType=cv2.BORDER_REFLECT)
 
         return cv2.resize(x, new_size, interpolation=cv2.INTER_LINEAR)
-    
+
     def _average(self, x):
         return cv2.filter2D(x.astype(np.float64), -1, self.avg_kernel, borderType=cv2.BORDER_REFLECT)
-    
+
     def _dilate(self, x):
         return (cv2.dilate(x.astype(np.uint8), self.dil_kernel) > 0).astype(np.uint8)
 
@@ -518,20 +518,20 @@ class AddTwinCloudMaskTask(EOTask):
 
         t = data.shape[0]
         b = data.shape[3]
-        
+
         indices = np.ndindex(t,b)
         output = None
-        
+
         for i,j in indices:
             res = func2d(data[i,...,j])
-            
+
             if output is None:
                 output = np.empty((t, *res.shape, b), dtype=res.dtype)
 
             output[i,...,j] = res
 
         return output
-    
+
     def _resize_all(self, data, new_size=None, scale_factors=None, anti_alias=True, interpolation=cv2.INTER_LINEAR):
         """
         Resizes the images acording to given size or scale factors.
@@ -548,16 +548,16 @@ class AddTwinCloudMaskTask(EOTask):
         :param interpolation: Interpolation method used for resampling. One of 'nearest', 'linear', 'cubic'.
         :type interpolation: bool
         """
-        
+
         # Old width and height
         old_size = (data.shape[2], data.shape[1])
 
         if new_size is not None and scale_factors is None:
             scale_factors = [new/old for old, new in zip(old_size, new_size)]
-            
+
         elif scale_factors is not None and new_size is None:
             new_size = [int(d * f) for d,f in zip(old_size, scale_factors)]
-            
+
         else:
             raise ValueError('Exactly one of the arguments new_size, scale_factors must be given.')
 
@@ -565,7 +565,7 @@ class AddTwinCloudMaskTask(EOTask):
         new_size = tuple(new_size)
 
         def _resize(image):
-            
+
             # Perform anti-alias smoothing if downscaling
             if downscaling and anti_alias:
                 sx, sy = [((1/s) - 1)/2 for s in scale_factors]
@@ -579,28 +579,28 @@ class AddTwinCloudMaskTask(EOTask):
         resized = self._map_sequence(data, _resize)
 
         return resized
-    
+
     def _average_all(self, data):
         return self._map_sequence(data, self._average)
-    
+
     def _dilate_all(self, data):
         return self._map_sequence(data, self._dilate)
-    
+
     def _ssim_stats(self, bands, is_data, mu, var, nt_rel):
-        
+
         ssim_max = np.empty((1, *bands.shape[1:]), dtype=np.float32)
         ssim_mean = np.empty_like(ssim_max)
         ssim_std = np.empty_like(ssim_max)
-        
+
         bands_r = np.delete(bands, nt_rel, axis=0)
         mu_r = np.delete(mu, nt_rel, axis=0)
         var_r = np.delete(var, nt_rel, axis=0)
-        
+
         n_frames = bands_r.shape[0]
         n_bands = bands_r.shape[-1]
-        
+
         valid_mask = np.delete(is_data, nt_rel, axis=0) & is_data[nt_rel,...,0].reshape(1,*is_data.shape[1:-1],1)
-        
+
         for b_i in range(n_bands):
             local_ssim = []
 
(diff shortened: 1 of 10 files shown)


lojzezust:
@devisperessutti The task should now be ready for basic use. The code might need to be further optimised and refactored.

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,563 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+        # Intersect
+        if self.mask_feature is not None:
+            inter_mask = mono_mask & multi_mask
+
+        # Add features
+        is_data = eopatch.mask[self.is_data_feature].astype(bool)
+
+        if mono_mask_feature is not None:
+            mono_mask = self._dilate_all(mono_mask)
+            eopatch.mask[mono_mask_feature] = (mono_mask * is_data).astype(bool)
```
devisperessutti: not sure if `is_data` should be included in the resulting masks and probabilities, or should be kept separate

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,563 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+        :param multi_classifier: Classifier used for multi-temporal cloud detection.
+                                 Default value: None (SSIM-based model is used)
+        :type multi_classifier: sklearn Estimator that works on the 90 multi-temporal features:
+                                 - raw reflectance value in the target frame,
+                                 - average value within a spatial window in the target frame,
+                                 - maximum, mean and standard deviation of the structural similarity (SSIM)
+                                 - indices between a spatial window in the target frame and every other,
+                                 - minimum and mean reflectance of all available time frames,
+                                 - maximum and mean difference in reflectances between the target frame and every other.
+        :param data_feature: Name of the key in the `eopatch.data` dictionary, which stores raw reflectance data.
```
devisperessutti: new tasks use tuples of `FeatureType` and `str` to define a feature and parse them with `self._parse_features`. Might be good to apply this where possible, although can be changed later on. The allowed `FeatureType` can be added to the parser (e.g. `Featuretype.DATA` for bands and `FeatureType.MASK` for `is_data`)

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,563 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+    downscaled by specifying the source and processing resolutions. This should be done
+    for the following reasons:
+        - faster execution
+        - lower memory consumption
+        - noise mitigation
+
+    Resizing is performed with linear interpolation. After classification, the cloud
+    probabilities are themselves upscaled to the original dimensions, before proceeding
+    with masking operations.
+    """
```
devisperessutti: given the high functionality of the task, it might be good to add to docstring a couple of call examples for running either or both single and MT

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,563 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+            if self.processing_resolution is not None:
+                rescale = (self.processing_resolution / height, self.processing_resolution / width)
+            else:
+                rescale = (1., 1.)
+
+            # Use default sigma value, since resolution is unknown
+            sigma = 1.0
+
+        elif service_type == ServiceType.WCS:
+            hr_res_x, hr_res_y = float(meta_info['size_x'].strip('m')), int(meta_info['size_y'].strip('m'))
```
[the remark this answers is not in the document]
lojzezust: Thanks for spotting!

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,563 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+        :param is_data_feature: Name of the key in the `eopatch.mask` dictionary, which indicates whether data is valid.
+                                Default value: `'IS_DATA'`.
+        :type is_data_feature: str
+        :param all_bands: Flag, which indicates whether images will consist of all 13 Sentinel-2 bands or only
+                          the required 10. Default value:  `True`.
+        :type all_bands: bool
+        :param processing_resolution: Resolution to be used during the computation of cloud probabilities and masks.
+                                      Default is `None` (source resolution).
+        :type processing_resolution: str or int
+        :param max_proc_frames: Maximum number of frames (including the target, for multi-temporal classification)
```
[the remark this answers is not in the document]
lojzezust: The number of features computed independent of the number of frames used. The data from all frames is aggregated into statistics. While it makes sense to use the same number of processing frames as the training, a different number of frames can be used without major differences in results.

mask/eolearn/mask/cloud_mask.py
JernejPuc: This step is only to ensure that masks stay within valid parts of the images. One can expect that areas, where `is_data == False`, are to be predicted as "clear" by the classifier itself, but dilation alone causes the masks to extend into invalid regions.

mask/eolearn/mask/cloud_mask.py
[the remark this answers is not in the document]
JernejPuc: It can be expected that the quality of the statistics will be improved, when more frames are considered at once, otherwise they can be randomly biased in one direction or another (eg. if we're unlucky and have predominantly cloudy images in our batch). Of course, this comes at a cost in terms of processing time and memory consumption, hence the option.

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,563 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+        :param data_feature: Name of the key in the `eopatch.data` dictionary, which stores raw reflectance data.
+                             Default value:  `'BANDS-S2-L1C'`.
+        :type data_feature: str
+        :param is_data_feature: Name of the key in the `eopatch.mask` dictionary, which indicates whether data is valid.
+                                Default value: `'IS_DATA'`.
+        :type is_data_feature: str
+        :param all_bands: Flag, which indicates whether images will consist of all 13 Sentinel-2 bands or only
+                          the required 10. Default value:  `True`.
+        :type all_bands: bool
+        :param processing_resolution: Resolution to be used during the computation of cloud probabilities and masks.
```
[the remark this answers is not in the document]
lojzezust: Yeah, it makes more sense to use two values. I will update the code accordingly.

mask/eolearn/mask/cloud_mask.py
lojzezust: If I understand correctly `self._parse_features` is intended for input features. Is this also needed for output features, or can they remain just strings (since the type of the feature is determined in the task)?

mask/eolearn/mask/cloud_mask.py
lojzezust: Added an example for combined masks (mono + multi) and only mono masks.

mask/eolearn/mask/cloud_mask.py
devisperessutti: it's usually used for input features, unless the output feature type needs to be specified by the user

mask/eolearn/mask/cloud_mask.py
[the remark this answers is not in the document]
devisperessutti: got it

mask/eolearn/mask/cloud_mask.py
devisperessutti: I see. OK for me to include `is_data` here already

mask/eolearn/mask/cloud_mask.py
lojzezust: Okay great. The input features are now parsed using `self._parse_features`. I've left the output features as they are.

lojzezust:
@devisperessutti Thank you for the feedback. All of the comments should be resolved now.

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
                                 average_over=average_over,
                                 dilation_size=dilation_size,
                                 all_bands=all_bands)
+
+
+# Twin classifier
+MONO_CLASSIFIER_NAME = 'pixel_s2_cloud_detector_lightGBM_v0.1.joblib.dat'
+MULTI_CLASSIFIER_NAME = 'ssim_s2_cloud_detector_lightGBM_v0.2.joblib.dat'
+
```
mlubej: Missing extra new line (pycharm error)

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+                 processing_resolution=None,
+                 max_proc_frames=11,
+                 mono_features=None,
+                 multi_features=None,
+                 mask_feature='CLM_INTERSSIM',
+                 mono_threshold=0.4,
+                 multi_threshold=0.5,
+                 average_over=1,
+                 dilation_size=1
+                ):
```
mlubej: misaligned bracket (pycharm error)

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+                pres = (pres_x + pres_y) / 2.0
+
+                rescale = (hr_res_y / pres_y, hr_res_x / pres_x)
+            else:
+                pres = (hr_res_x + hr_res_y) / 2.
+                rescale = (1., 1.)
+
+            sigma = 100. / pres
+
+        return rescale, sigma
```
mlubej: 1) If WMS and WCS are the only possible choices, then end with `else`, otherwise add an extra else for an unknown/incompatible service type.

2) In case that it's neither a WCS nor WMS, the `rescale` and `sigma` parameters are potentially referenced before assignment. Perhaps it's better to set some default values at the appropriate scope level first and set them later if the appropriate conditions are satisfactory.

3)
instead of
```python
if condition:
    var = something
else:
    var = something_generic
```

you can do 

```python
var = something_generic
if condition:
    var = something
```

it's the same, but more intuitive

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+        Returns a new array with the combined results.
+
+        :param data: input array
+        :type data: array of shape (timestamps, rows, columns, channels)
+        :param func2d: Mapping function that is applied on each 2d image slice. All outputs must have the same shape.
+        :type func2d: function (rows, columns) -> (new_rows, new_columns)
+        """
+
+        func3d = lambda x: map_over_axis(x, func2d, axis=2) # Map over channel dimension on 3d tensor
+        func4d = lambda x: map_over_axis(x, func3d, axis=0) # Map over time dimension on 4d tensor
```
mlubej: two spaces for inline comments (pycharm)

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+            bands_t = bands[nt_min:nt_max]
+
+            mono_features = bands_t.reshape(np.prod(bands_t.shape[:-1]), bands_t.shape[-1])
+
+            # Run mono classifier
+            mono_proba[nt_min*img_size:nt_max*img_size] = self.mono_classifier.predict_proba(mono_features)[..., 1:]
+
+        return mono_proba
+
+
```
mlubej: one blank line too much (pycharm)

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+            # Upscale if necessary
+            if scale_factors is not None:
+                mono_proba = resize_images(mono_proba, new_size=original_shape)
+
+            # Average over and threshold
+            mono_mask = self._average_all(mono_proba) >= self.mono_threshold
+
+        # Run SSIM-based multi-temporal classifier if needed
+        if any(feature is not None for feature in [self.mask_feature, multi_mask_feature, multi_proba_feature]):
+            multi_proba = self._multi_iterations(bands, is_data_sm)
```
mlubej: `is_data_sm` might be unreferenced because it's defined under an `if` condition in line 835.

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+            mono_mask = self._dilate_all(mono_mask)
+            eopatch.mask[mono_mask_feature] = (mono_mask * is_data).astype(bool)
+
+        if multi_mask_feature is not None:
+            multi_mask = self._dilate_all(multi_mask)
+            eopatch.mask[multi_mask_feature] = (multi_mask * is_data).astype(bool)
+
+        if self.mask_feature is not None:
+            inter_mask = self._dilate_all(inter_mask)
+            eopatch.mask[self.mask_feature] = (inter_mask * is_data).astype(bool)
```
mlubej: Same as before for `mono_mask`, `multi_mask` and `inter_mask`

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+            multi_proba = self._multi_iterations(bands, is_data_sm)
+            multi_proba = multi_proba.reshape(*bands.shape[:-1], 1)
+
+            # Upscale if necessary
+            if scale_factors is not None:
+                multi_proba = resize_images(multi_proba, new_size=original_shape)
+
+            # Average over and threshold
+            multi_mask = self._average_all(multi_proba) >= self.multi_threshold
+
```
mlubej: extra blank line (pycharm)

mask/eolearn/tests/test_cloud_mask.py
```diff
@@ -108,5 +108,86 @@ def test_wms_request(self):
...
+        self.assertAlmostEqual(np.mean(eop_clm.data['CLP_S2C']), mean_clp_provided, places=2)
+
+        # Classifier is run on downscaled version of data array
+        add_tcm = AddMultiCloudMaskTask(data_feature='ALL_DATA',
+                                       processing_resolution='120m',
+                                       mono_features=('CLP_S2C', None),
+                                       mask_feature='CLM_INTERSSIM',
+                                       average_over=16,
+                                       dilation_size=8
+                                      )
```
mlubej: misaligned indentations (pycharm)

mask/eolearn/tests/test_cloud_mask.py
```diff
@@ -108,5 +108,86 @@ def test_wms_request(self):
...
+
+        # Compare mean cloud coverage with provided reference
+        mean_clm_provided = np.mean(eop_clm.mask['CLM'])
+        mean_clp_provided = np.mean(eop_clm.data['CLP'])
+        self.assertAlmostEqual(np.mean(eop_clm.mask['CLM_INTERSSIM']), mean_clm_provided, places=2)
+        self.assertAlmostEqual(np.mean(eop_clm.data['CLP_S2C']), mean_clp_provided, places=1)
+
+        # Check if most of the same times are flagged as cloudless
+        cloudless = np.mean(eop_clm.mask['CLM_INTERSSIM'], axis=(1,2,3)) == 0
+        self.assertTrue(np.mean(cloudless == eop_clm.label['IS_CLOUDLESS'][:,0]) > 0.94)
```
mlubej: Missing whitespace after `,` (pycharm)

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+        :param average_over: Size of the pixel neighbourhood used in the averaging post-processing step.
+                             A value of `0` or `None` skips this post-processing step. Default value: `1`.
+        :type average_over: int or None
+        :param dilation_size: Size of the dilation post-processing step. A value of `0` or `None` skips
+                              this post-processing step. Default value: `1`.
+        :type dilation_size: int or None
+        """
+
+        # Load classifiers
+        if mono_classifier is None or multi_classifier is None:
```
mlubej: This should always be true, or not? If not, there should rather be a check for this anyway.

mask/eolearn/mask/cloud_mask.py
```diff
@@ -292,3 +297,592 @@ def get_s2_pixel_cloud_detector(threshold=0.4, average_over=4, dilation_size=2,
...
+            classifier_dir = os.path.dirname(__file__)
+
+            if mono_classifier is None:
+                mono_classifier = joblib.load(os.path.join(classifier_dir, 'models', MONO_CLASSIFIER_NAME))
```
mlubej: This line fails for me if I do the following in a jupyter notebook:

```python
from eolearn.core import EOPatch, LoadFromDisk
from eolearn.mask import AddCloudMaskTask, get_s2_pixel_cloud_detector, AddMultiCloudMaskTask

eop = EOPatch.load('/mnt/s3/turlcms/tr-project/2018/L1C/data/yearly/eopatch-id-468-col-14-row-58', lazy_loading=True)
add_multi = AddMultiCloudMaskTask(processing_resolution='120m',
                                  mask_feature='CLM_INTERSSIM',
                                  average_over=16,
                                  dilation_size=8)
```

Probably the model pickle version is too old? Because for the lower one it works

mask/eolearn/mask/cloud_mask.py
lojzezust: If you look at the code before, when at least one of `self.mask_feature, mono_mask_feature, mono_proba_feature` is not None `mono_mask` is computed. Similar for `multi_mask`.

Therefore the when the mentioned if statements will be true (and executed), the required masks will always be computed.

mask/eolearn/mask/cloud_mask.py
lojzezust: It's not always true. Both classifiers may be provided by the user.

What do you mean by there should be a check?

mask/eolearn/mask/cloud_mask.py
lojzezust: What do you mean by the lower one?

mask/eolearn/mask/cloud_mask.py
lojzezust: Added else for unknown service types. I used default variable values as you mentioned where I thought appropriate.

mask/eolearn/mask/cloud_mask.py
lojzezust: The outside if can be removed without problems if that's what you meant. I agree that there's no need for `classifier_dir` to only be initialized when needed.

The code might be more readable this way.

mask/eolearn/mask/cloud_mask.py
lojzezust: Ok now the tests fail on Travis as well. It seems the libraries were updated and the model can't be read anymore.

mask/eolearn/mask/cloud_mask.py
mlubej: OK, no problem. I just wanted to point this out as a possible problem, without really going deep into it. Thanks!

mask/eolearn/mask/cloud_mask.py
mlubej: I checked that the models were pickled with two different versions of pickle and the one that fails seems to be too old. Did you manage to pinpoint this, otherwise I can give you more information.

mlubej: OK, here is some more info. I Try this:

```python
import os
import joblib

DIR = '/home/users/mlubej/projects/eo-learn/mask/eolearn/mask/models/'
MONO_CLASSIFIER_NAME = 'pixel_s2_cloud_detector_lightGBM_v0.1.joblib.dat'
MULTI_CLASSIFIER_NAME = 'ssim_s2_cloud_detector_lightGBM_v0.2.joblib.dat'

models = [MONO_CLASSIFIER_NAME, MULTI_CLASSIFIER_NAME]

for model in models:
    try:
        joblib.load(os.path.join(DIR, model))
        print(f'Loaded {model} successfully!')
    except:
        print(f'Failed to load {model}')
```

and the output is:

```
Failed to load pixel_s2_cloud_detector_lightGBM_v0.1.joblib.dat
Loaded ssim_s2_cloud_detector_lightGBM_v0.2.joblib.dat successfully!
```

While the error for loading the `pixel_s2_cloud_detector_lightGBM_v0.1.joblib.dat` is:

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-17-090eb69ef74f> in <module>
----> 1 joblib.load(os.path.join(DIR, MONO_CLASSIFIER_NAME))

~/base/lib/python3.6/site-packages/joblib/numpy_pickle.py in load(filename, mmap_mode)
    596                     return load_compatibility(fobj)
    597 
--> 598                 obj = _unpickle(fobj, filename, mmap_mode)
    599 
    600     return obj

~/base/lib/python3.6/site-packages/joblib/numpy_pickle.py in _unpickle(fobj, filename, mmap_mode)
    524     obj = None
    525     try:
--> 526         obj = unpickler.load()
    527         if unpickler.compat_mode:
    528             warnings.warn("The file '%s' has been generated with a "

/usr/lib/python3.6/pickle.py in load(self)
   1048                     raise EOFError
   1049                 assert isinstance(key, bytes_types)
-> 1050                 dispatch[key[0]](self)
   1051         except _Stop as stopinst:
   1052             return stopinst.value

KeyError: 0
```

lojzezust: Thanks! We updated the model pickle version.

lojzezust: Can you test if it works for you now?

mlubej: The code now runs for me and the model loads, so that's fixed.

Can you look into why the Travis build is failing?

Also, do you know why running `pytest` didn't give this error? That's weird.

lojzezust: Travis build fails on this `The command "if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then pytest --ignore io/eolearn/tests; fi" exited with 1.`

Looks like a problem in `io` module? 
Not sure why, but Travis builds started failing this week. Code changes were minor and shouldn't break the tests.

mlubej: @AleksMat any ideas?

mlubej:
Will merge to develop when/if checks finish successfully

PR #151, opened 2019-10-09
Added doubly logistic approximation task.

bsircelj:


features/eolearn/features/doubly_logistic_approximation.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        data = eopatch.data[self.feature].squeeze(axis=3)
+
+        times = np.asarray([time.toordinal() for time in eopatch.timestamp])
+        times = (times - times[0]) / (times[-1] - times[0])
+
+        time, height, width = data.shape
+
+        if self.mask_data:
+            valid_data_mask = eopatch.mask['VALID_DATA']
```
azupanc: Please avoid hardcoding feature names. I suggest to change `mask_data` (bool) argument to `mask_feature` (FeatureType, feature_name) to make it more generalizable.

mlubej:
Hi @bsircelj!

Thanks again for preparing this and sorry for the wait. Could you please update the branch so the tests work? If it's OK with you, we can also push the changes ourselves to your fork before merging.

Cheers.

mlubej:
>  Merge pull request #10 from sentinel-hub/master 

same here :)

mlubej:
New conflicts arose when merging your other PR, please fix so we can merge this as well :)

features/eolearn/tests/test_doubly_logistic_approximation.py
```diff
@@ -0,0 +1,77 @@
...
+from eolearn.core import EOPatch, FeatureType
+from eolearn.features.doubly_logistic_approximation import DoublyLogisticApproximationTask
+from datetime import datetime
+
+
+class TestDoublyLogisticApproximation(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        dates = [datetime(2017, 1, 1), datetime(2017, 1, 11), datetime(2017, 2, 7), datetime(2017, 2, 17),
```
mlubej, line 13: Instead of using hand-crafted values, could you perhaps change this so that it uses the test EOPatch from `example_data/TestEOPatch`?

bsircelj: Inverted mask
I found out that IS_VALID mask in the test patch appears to be inverted. The pixels marked with True have a value of nan.

bsircelj:
I used only the first pixel of test patch as the computation is really slow. On my computer it takes 20s to fit one function. It would take about 2h for the whole patch. Also I inverted the IS_VALID mask. The test data is seemingly random and doesn't accurately represents the crop growth cycles so this is why I avoided it initially. This are the results for fitting on the first pixel.
doubly_test

features/eolearn/tests/test_doubly_logistic_approximation.py
mlubej: Not sure what you mean by inverted. `IS_VALID` is 0 when the image is invalid (clouds or no data) and vice versa otherwise, i.e. in the image below, the cloudy parts are masked where `IS_VALID` is `FALSE`

image

For which temporal index in the TestEOPatch do you see this? I'm also not sure what you are plotting above.

bsircelj: image
I should've specified what colors mean. Here is another picture. On the left is displayed which pixels of the `(FeatureType.DATA` `'ndvi')` are equal to nan. White - "is nan", black -"not nan, data".
On the right is displayed the `IS_VALID` mask where White-True, Black-False.

The pixels which are marked with mask `IS_VALID` as True don't actually have any data and vice versa.

The offender is the patch in eo-learn/features/eolearn/tests/TestInputs/TestPatch.

mlubej: You are right, thanks for finding this. This must be some weird patch. I'll see what the damage is, meanwhile could you please use a different test EOPatch? It's in `eo-learn/example_data/TestEOPatch`. This one has a proper valid mask, but you have to calculate te ndvi

mlubej:
> I used only the first pixel of test patch as the computation is really slow. On my computer it takes 20s to fit one function. It would take about 2h for the whole patch. Also I inverted the IS_VALID mask. The test data is seemingly random and doesn't accurately represents the crop growth cycles so this is why I avoided it initially. This are the results for fitting on the first pixel.
> doubly_test

This looks super weird.. :D There is also a lot of points for just one pixel. Perhaps you can try the other test patch which I mentioned above?

features/eolearn/features/doubly_logistic_approximation.py
```diff
@@ -0,0 +1,84 @@
...
+        :type x_axis: List of floats
+        :param y_axis: Vertical coordinates of points
+        :type y_axis: List of floats
+        :return: List of optimized parameters [c1, c2, a1, a2, a3, a4, a5]
+        """
+        bounds_lower = [np.min(y_axis), -np.inf, x_axis[0], 0.15, 1, 0.15, 1, ]
+        bounds_upper = [np.max(y_axis), np.inf, x_axis[-1], np.inf, np.inf, np.inf, np.inf, ]
+        if self.initial_parameters is None:
+            self.initial_parameters = [np.mean(y_axis), 0.2, (x_axis[-1] - x_axis[0]) / 2, 0.15, 10, 0.15, 10]
+        optimal_values = curve_fit(doubly_logistic, x_axis, y_axis, self.initial_parameters,
```
mlubej, line 52: I think that the `sigma` parameter of the fit might be very useful as well, in cases where you have a mean and st.dev. NDVI and you want to fit the DL curve to these values, instead of the values of one pixel.

features/eolearn/features/doubly_logistic_approximation.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    EOTask class for calculation of doubly logistic approximation on each pixel for a feature. The task creates new
+    feature with the function parameters for each pixel as vectors.
+    :param feature: A feature on which the function will be approximated
+    :type feature: str
+    :param new_feature: Name of the new feature where parameters of the function are saved
+    :type new_feature: str
+    :param initial_parameters: Initial parameter guess
+    :type initial_parameters: List of floats length 7 corresponding to each parameter
+    :param mask_feature: A feature used as a mask for valid regions. If left as None the whole patch is used
```
mlubej, line 30: Perhaps it's clearer if this parameter is `valid_mask` or something, we've had issues in the past due to the confusion.

features/eolearn/features/doubly_logistic_approximation.py
mlubej: Nevermind, I forgot this is done per pixel.

bsircelj:
doubly_pixel_00
Much better.

bsircelj:
Locally it only works like this:
 `from features.eolearn.features.doubly_logistic_approximation import DoublyLogisticApproximationTask`
But travis builds without the first "features" 
`from eolearn.features.doubly_logistic_approximation import DoublyLogisticApproximationTask`
Picture of error I get if I try to run the version that Travis accepts:
module not found

mlubej:
>  `from features.eolearn.features.doubly_logistic_approximation import DoublyLogisticApproximationTask`

This looks like a local file import, not an import from the installed package. Try reinstalling the package to incorporate the changes, or install eolearn in edit mode with `python install_all.py -e`

Then it should work in the way Travis does it.

mlubej:
Tests pass, I think there is no other change needed. @azupanc, unless there are some other specifications, this can be merged.

azupanc:
> Tests pass, I think there is no other change needed. @azupanc, unless there are some other specifications, this can be merged.

No, no additional requirements. Thank you both for cooperation and hard work!

PR #155, opened 2019-10-16
Bugfix/logging multithread

veseln:
When running in multithreading mode, the log files for single eopatches
would also log the information for the eopatches running in different
threads. Adding the filter that filters messages based on the
thread name fixes this problem.
In addition, a message 'Pipeline finished.' or 'Pipeline failed.' is written to the end of every log file depending on the successful execution.

### DIFF
--- a/core/eolearn/core/eoexecution.py
+++ b/core/eolearn/core/eoexecution.py
@@ -16,6 +16,7 @@ file in the root directory of this source tree.
 
 import os
 import logging
+import threading
 import traceback
 import concurrent.futures
 import datetime as dt
@@ -23,6 +24,8 @@ import datetime as dt
 from tqdm.auto import tqdm
 
 from .eoworkflow import EOWorkflow
+
+from .utilities import LogFileFilter
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


core/eolearn/core/utilities.py
```diff
@@ -21,6 +22,20 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class LogFileFilter(Filter):
+    """ Filters log messages passed to log file
+    """
+
+    def __init__(self, thread_name, name=''):
+        self.thread_name = thread_name
+        super().__init__(name=name)
```
AleksMat: It would be safer to write
```
    def __init__(self, thread_name, *args, **kwargs):
        self.thread_name = thread_name
        super().__init__(*args, **kwargs)
```
Even though current implementation of `Filter` class only has `name` parameter we don't know what will happen in the future Python versions. Hence it's better to just pass any possible parameter forward.

core/eolearn/core/eoexecution.py
```diff
@@ -169,6 +171,7 @@ def _execute_workflow(cls, process_args):
         stats[cls.STATS_END_TIME] = dt.datetime.now()
 
         if log_path:
+            logger.info(msg='Pipeline failed.' if cls.STATS_ERROR in stats else 'Pipeline finished.')
             handler.close()
             logger.removeHandler(handler)
```
AleksMat: It might be good to put this code under try-catch statement as well. This is for the case if adding a log handler would fail. Current code would catch it but then removing a non-existing log handler would fail as well. Maybe we could even split this logic into new methods and make this main method more readable:

```Python
workflow, input_args, log_path, return_results = process_args

logger, handler = self._try_add_logging(log_path)

stats = {cls.STATS_START_TIME: dt.datetime.now()}
try:
    results = workflow.execute(input_args, monitor=True)

    if return_results:
        stats[cls.RESULTS] = results

except BaseException:
    stats[cls.STATS_ERROR] = traceback.format_exc()

stats[cls.STATS_END_TIME] = dt.datetime.now()

self._try_remove_logging(log_path, logger, handler)

return stats
```
where content of `_try_add_logging` and `_try_remove_logging` would be approximately in a form
```Python
if log_path:
    try:
        ...
    except BaseException:
        pass
```

PR #156, opened 2019-10-24
empty dataframe bug

batic:
Handles case when vector dataframe is empty after buffering of geometries.


mlubej:
The PR was done on an old version of `develop`. I merged `develop` into this branch to check if Travis tests still fail.

PR #169, opened 2019-12-19
Implementation of a filesystem-agnostic loading using PyFilesystem2

iovsn:


core/eolearn/core/eodata.py
```diff
@@ -796,6 +803,49 @@ def _load_data(file_handler, is_pickle):
...
+    @staticmethod
+    def walk_filtered(filesystem, patch_location, features):
+        """ Based on EOPatch.walk generator, but applies feature filtering based on the `features` argument
+        """
+        itr = EOPatch.walk(filesystem, patch_location)
+        itr = ((FeatureType(ftype), fname, path) for ftype, fname, path in itr)
+
+        features = FeatureParser(features).feature_collection
+        for ftype, fname, path in itr:
+            if ftype in [FeatureType.BBOX, FeatureType.TIMESTAMP, FeatureType.META_INFO]:
```
AleksMat: You can check `ftype.is_meta()` instead.

core/eolearn/core/eodata.py
```diff
@@ -796,6 +803,49 @@ def _load_data(file_handler, is_pickle):
...
+                yield ftype, fname, path
+
+            if ftype not in features.keys():
```
AleksMat: You don't need `.keys()`

core/eolearn/core/eodata.py
```diff
@@ -796,6 +803,49 @@ def _load_data(file_handler, is_pickle):
...
+                    yield path, file.split('.')[0], fs.path.combine(subdir, file)
+            else:
+                yield path.split('.')[0], None, fs.path.combine(patch_location, path)
+
+    @staticmethod
+    def walk_filtered(filesystem, patch_location, features):
+        """ Based on EOPatch.walk generator, but applies feature filtering based on the `features` argument
+        """
+        itr = EOPatch.walk(filesystem, patch_location)
+        itr = ((FeatureType(ftype), fname, path) for ftype, fname, path in itr)
```
AleksMat: Would it be better/nicer to do `FeatureType(ftype)` (i.e. casting from string to enum) already in `walk` method?

core/eolearn/core/eodata.py
```diff
@@ -796,6 +803,49 @@ def _load_data(file_handler, is_pickle):
...
+        features = FeatureParser(features).feature_collection
+        for ftype, fname, path in itr:
+            if ftype in [FeatureType.BBOX, FeatureType.TIMESTAMP, FeatureType.META_INFO]:
+                yield ftype, fname, path
+
+            if ftype not in features.keys():
+                continue
+
+            ftrs = features[ftype]
+            if ftrs == Ellipsis or fname in ftrs:
```
AleksMat: It is better to check `ftrs is Ellipsis` instead of `ftrs == Ellipsis`

core/eolearn/core/eodata.py
```diff
@@ -796,6 +803,49 @@ def _load_data(file_handler, is_pickle):
...
+        for path in filesystem.listdir(patch_location):
+            if '.' not in path:
+                subdir = fs.path.combine(patch_location, path)
+                for file in filesystem.listdir(subdir):
+                    yield path, file.split('.')[0], fs.path.combine(subdir, file)
+            else:
+                yield path.split('.')[0], None, fs.path.combine(patch_location, path)
+
+    @staticmethod
+    def walk_filtered(filesystem, patch_location, features):
```
AleksMat: In case we set features filter, there is still some room to reduce the number of requests that are made to the filesystem. We can talk more in person.

core/eolearn/core/eodata.py
```diff
@@ -1030,6 +1080,38 @@ def _parse_feature_value(self, value):
...
+
+        if self.path.endswith('.gz'):
+            with gzip.open(file_handle, 'rb') as gzip_fp:
+                return self._decode(gzip_fp, self.path)
+
+        return self._decode(file_handle, self.path)
+
+    @staticmethod
+    def _decode(file, path):
+        if '.pkl' in path:
```
AleksMat: To avoid using strings you could do `FileFormat.PICKLE.extension() in path` and similarly everywhere else.

Also, it might be nice if we turn `FileFormat.extension` method into a property.

core/eolearn/core/eodata.py
```diff
@@ -1030,6 +1080,38 @@ def _parse_feature_value(self, value):
...
+    def load(self):
+        file_handle = BytesIO()
+        self.filesystem.download(self.path, file_handle)
+        file_handle.seek(0)
+
+        if self.path.endswith('.gz'):
+            with gzip.open(file_handle, 'rb') as gzip_fp:
+                return self._decode(gzip_fp, self.path)
+
+        return self._decode(file_handle, self.path)
```
AleksMat: As discussed a few moments ago, we could get rid of intermediate `BytesIO` and do the following:
```Python
        with self.filesystem.open(self.path, 'rb') as file_handle:
            if self.path.endswith('.gz'):
                with gzip.open(file_handle, 'rb') as gzip_fp:
                    return self._decode(gzip_fp, self.path)

            return self._decode(file_handle, self.path)
```
This way we might avoid some memory duplications.

AleksMat: Actually, instead of `self.filesystem.open(self.path, 'rb')` we could even call `self.filesystem.openbin(self.path)` as `open` is only a wrapper around `openbin`.

core/eolearn/core/eodata.py
```diff
@@ -796,6 +803,49 @@ def _load_data(file_handler, is_pickle):
             eopatch[feature] = data
         return eopatch
 
+    @staticmethod
+    def walk(filesystem, patch_location):
+        """ Recursively reads a patch_location and returns yelds tuples of (feature_type, feature_name, file_path)
```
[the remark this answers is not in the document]
iovsn: Done.

core/eolearn/core/eodata.py
iovsn: Done.

core/eolearn/core/eodata.py
iovsn: Done.

core/eolearn/core/eodata.py
iovsn: I agree. Done.

core/eolearn/core/eodata.py
iovsn: Done.

core/eolearn/core/eodata.py
iovsn: Will improve this at later stage, when implementing saving.

core/eolearn/core/eodata.py
iovsn: Done.

core/eolearn/core/eodata.py
iovsn: Done.

PR #170, opened 2020-01-05
Feat/fs improvements

AleksMat:
This branch should cover everything that was still missing regarding s3 IO support and a bit more.

A summary of changes:

- Moved all logic about `EOPatch` saving and loading to a new module `eodata_io`
- Added a module `fs_utils`
- Updated all unit tests regarding saving and loading. Tests now check working both with a local fs and s3 fs.
- Removed all redundant code.
- Added tasks `SaveTask` and `LoadTask` and deprecated `SaveToDisk` and `LoadFromDisk`.
- Optimized loading and saving with multithreading.
- Fixed failing unit tests in other subpackages.
- Fixed CI build, added support for Python 3.8.


core/eolearn/core/eodata_io.py
```diff
@@ -0,0 +1,260 @@
...
+    for (ftype, fname, _), value in zip(features, loading_data):
+        eopatch[(ftype, fname)] = value
+
+    return eopatch
+
+
+def walk_filesystem(filesystem, patch_location, features=...):
+    """ Recursively reads a patch_location and returns yields tuples of (feature_type, feature_name, file_path)
+    """
+    existing_features = {ftype: {} for ftype in FeatureType}
```
veseln: existing_features = defaultdict(dict) does what you want here... Not sure if its better tho, might be worth considering.

core/eolearn/core/eodata_io.py
```diff
@@ -0,0 +1,260 @@
...
+                    yield ftype, feature_name, path
+        else:
+            yield ftype, fname, existing_features[ftype][fname]
+
+
+def walk_main_folder(filesystem, folder_path):
+    """ Walks the main EOPatch folders and yields tuples (feature type, feature name, path in filesystem)
+    """
+    for path in filesystem.listdir(folder_path):
+        raw_path = path.split('.')[0].strip('/')
```
veseln: is this filesystem agnostic?

core/eolearn/core/eodata_io.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if '/' in raw_path:
+            ftype_str, fname = raw_path.split('/', 1)
+        else:
```
veseln: Would it be possible to rewrite this using [fs info api](https://docs.pyfilesystem.org/en/latest/reference/info_objects.html#fs.info.Info) using `filesystem.getbasic(path)` or `filesystem.getinfo(path)` calls?

veseln: With this I mean rewriting all the path juggling / splitting in the code.

core/eolearn/core/eodata_io.py
```diff
@@ -0,0 +1,260 @@
...
+    """ A utility function used by EOPatch.save method
+    """
+    patch_exists = filesystem.exists(patch_location)
+
+    if overwrite_permission is OverwritePermission.OVERWRITE_PATCH and patch_exists:
+        filesystem.removetree(patch_location)
+        patch_exists = False
+
+    if not patch_exists:
+        filesystem.makedir(patch_location, recreate=True)
```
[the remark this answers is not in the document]
AleksMat: This was because of a special case when `patch_location='/'` and `overwrite_permission=OverwritePermission.OVERWRITE_PATCH`. But I see it wasn't a very nice solution and I changed how we handle that case. Now we just use `filesystem.makedir(patch_location)` which by default sets `recreate=False`.

core/eolearn/core/eodata_io.py
```diff
@@ -0,0 +1,260 @@
...
+
+    eopatch_features = list(walk_eopatch(eopatch, patch_location, features))
+
+    if overwrite_permission is OverwritePermission.ADD_ONLY or \
+            (sys_is_windows() and overwrite_permission is OverwritePermission.OVERWRITE_FEATURES):
```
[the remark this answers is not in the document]
AleksMat: Yes, on windows you can't have 2 files in the same folder which differ only in a casing (e.g. `feature.npy` and `FEATURE.npy`).

core/eolearn/core/eodata_io.py
AleksMat: Good point, changed.

core/eolearn/core/eodata_io.py
AleksMat: Yes, every subclass of `fs.base.FS` object has to implement `listdir` method ([source](https://github.com/PyFilesystem/pyfilesystem2/blob/master/fs/base.py#L172)).

core/eolearn/core/eodata_io.py
AleksMat: I wanted to do that but then I have seen `fs.path` is internally also using hardcoded `/` sign everywhere ([source](https://github.com/PyFilesystem/pyfilesystem2/blob/master/fs/path.py)). So I'm not sure if it makes sense to complicate at this point. Also not sure how that would work on Windows though...
But I did change `raw_path.rsplit('/', 1)` into `fs.path.split(raw_path)`.

core/eolearn/core/eodata_io.py
```diff
@@ -0,0 +1,260 @@
...
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            loading_data = executor.map(lambda loader: loader.load(), loading_data)
+
+    for (ftype, fname, _), value in zip(features, loading_data):
+        eopatch[(ftype, fname)] = value
+
+    return eopatch
+
+
+def walk_filesystem(filesystem, patch_location, features=...):
```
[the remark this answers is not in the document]
AleksMat: I share your concern but, as discussed, I couldn't find a nicer way to cover all the cases.

core/eolearn/core/eodata.py
```diff
@@ -83,11 +71,11 @@ def __setattr__(self, key, value, feature_name=None):
 
         In case they are a dictionary they are cast to _FeatureDict class
         """
-        if feature_name is not None and FeatureType.has_value(key):
+        if feature_name not in (None, Ellipsis) and FeatureType.has_value(key):
```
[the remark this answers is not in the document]
AleksMat: This covers both options - what a user might provide and what a `FeatureParser` might provide. For now, I would leave it as it is but I agree that it is a bit confusing and that sometime in the future it could be improved. But that would require quite some time.

core/eolearn/core/eodata.py
```diff
@@ -104,30 +92,30 @@ def _parse_feature_type_value(feature_type, value):
...
         value = super().__getattribute__(key)
 
-        if isinstance(value, (_FileLoader, _Loader)) and load:
+        if isinstance(value, FeatureIO) and load:
             value = value.load()
             setattr(self, key, value)
             value = getattr(self, key)
 
-        if feature_name is not None and isinstance(value, _FeatureDict):
+        if feature_name not in (None, Ellipsis) and isinstance(value, _FeatureDict):
```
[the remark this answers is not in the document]
AleksMat: Same answer.

core/eolearn/core/eodata_io.py
AleksMat: I had another look at this - it turns out working with `/` is ok also on Windows machines. That is because the Windows `OSFS` filesystem will always be accepting and returning paths with `/` separators and only internally switch to `\` separators.

veseln: Great, thanks for looking into it. Makes sense how it is now then.

PR #171, opened 2020-01-08
Feat/fs with multiprocessing

AleksMat:
Changes:
- Modified `SaveTask` and `LoadTask` in order to allow being used with multiprocessing
- A temporary fix in .travis.yml

### DIFF
--- a/core/eolearn/core/core_tasks.py
+++ b/core/eolearn/core/core_tasks.py
@@ -62,16 +62,23 @@ class SaveTask(EOTask):
             to 9 (highest compression).
         :type compress_level: int
         :param filesystem: An existing filesystem object. If not given it will be initialized according to the EOPatch
-            path
-        :type filesystem: fs.FS or None
+            path. If you intend to run this task in multiprocessing mode you shouldn't specify this parameter.
+        :type filesystem: fs.base.FS or None
         """
         self.path = path
-        self.filesystem = filesystem
+        self._filesystem = filesystem
         self.kwargs = kwargs
 
-        if self.filesystem is None:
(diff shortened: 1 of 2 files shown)


core/eolearn/core/core_tasks.py
```diff
@@ -62,16 +62,23 @@ def __init__(self, path, filesystem=None, **kwargs):
             to 9 (highest compression).
         :type compress_level: int
         :param filesystem: An existing filesystem object. If not given it will be initialized according to the EOPatch
-            path
-        :type filesystem: fs.FS or None
+            path. If you intend to run this task in multiprocessing mode you shouldn't specify this parameter.
+        :type filesystem: fs.base.FS or None
```
veseln: Just being pedantic, but I think the commonly used notation for this would be Optional[fs.base.Fs]. For exmaple if you used  MyPy typechecker and typing, you would write it like this. 
But no need to change this.

AleksMat: Good point, but this would require a more systematic change everywhere. Besides, Sphinx recognizes the current way correctly, which can be seen from links in documentation.

AleksMat:
> Also the signatures of save/load functions are the same. Since we're deprecating the old `SaveToDisk` and `LoadFromDisk`, do we really need to keep the two tasks or could we join them together into one? There could be just one `IOTask`, covering both saving and loading.

Good point, let's add a common `IOTask`. But I would still keep `SaveTask` and `LoadTask`, which would inherit from `IOTask`. That is because the initialization parameters between these 2 tasks are still quite different, but they are passed forward to EOPatch methods.

PR #172, opened 2020-01-08
Implementation of DEM support and 'single_scene' mosaicing requests

iovsn:
- SentinelHubInputBase class has been implemented as a basis for other Processing API input tasks
- specifics have been moved to SentinelHubInputTask and SentinelHubDEMInputTask


io/eolearn/io/processing_api.py
```diff
@@ -6,84 +6,51 @@
...
 
 
-class SentinelHubInputTask(EOTask):
-    """ A processing API input task that loads 16bit integer data and converts it to a 32bit float feature.
-    """
-    def __init__(self, data_source, size=None, resolution=None, bands_feature=None, bands=None, additional_data=None,
-                 maxcc=1.0, time_difference=None, cache_folder=None, max_threads=None, config=None,
-                 bands_dtype=np.float32):
+class SentinelHubInputBase(EOTask):
+    ''' Base class for Processing API input tasks
```
AleksMat: Could you replace `'''` with `"""` everywhere. Using double-quotes for docs string is the standard way.

io/eolearn/io/processing_api.py
```diff
@@ -6,84 +6,51 @@
...
-        elif bands_feature is not None:
-            self.bands = data_source.bands()
-        else:
-            self.bands = []
-
-        if additional_data is None:
-            self.additional_data = []
-        else:
-            self.additional_data = list(self._parse_features(additional_data, new_names=True)())
+        self.timestamp = None
```
AleksMat: If I'm not mistaken `self.timestamp` is not used anywhere and could be removed.

io/eolearn/io/processing_api.py
```diff
@@ -6,84 +6,51 @@
...
         :param cache_folder: Path to cache_folder. If set to None (default) requests will not be cached.
         :type cache_folder: str
+        :param config: An instance of SHConfig defining the service
+        :type config: SHConfig or None
         :param max_threads: Maximum threads to be used when downloading data.
         :type max_threads: int
         """
+
         self.size = size
         self.resolution = resolution
```
AleksMat: Maybe the following check could be added:
```Python
if (self.size is None) is (self.resolution is None):
    raise ValueError("Exactly one of the parameters 'size' and 'resolution' should be given, not both")
```
Otherwise cases where none or both are set would be poorly handled.

io/eolearn/io/processing_api.py
```diff
@@ -98,64 +65,118 @@ def execute(self, eopatch=None, bbox=None, time_interval=None):
...
+    def _add_meta_info(self, eopatch):
+        """ Add any additional meta data to the eopatch
+        """
+
+
+class SentinelHubInputTask(SentinelHubInputBase):
+    """ A processing API input task that loads 16bit integer data and converts it to a 32bit float feature.
+    """
+    def __init__(self, data_source, size=None, resolution=None, bands_feature=None, bands=None, additional_data=None,
+                 maxcc=1.0, time_difference=None, cache_folder=None, max_threads=None, config=None,
```
AleksMat: Parameter `max_threads` is not passed forward to parent class. (PyCharm marks such things ;) )

io/eolearn/io/processing_api.py
```diff
@@ -230,3 +293,75 @@ def _extract_bands(self, eopatch, images, shape):
...
+    def _add_meta_info(self, eopatch):
+        eopatch.meta_info['maxcc'] = self.maxcc
+        eopatch.meta_info['time_difference'] = self.time_difference
+
+
+class SentinelHubDEMInputTask(SentinelHubInputBase):
+    ''' A processing API input task that downloads the digital elevation model
+    '''
+    def __init__(self, dem_feature, size=None, resolution=None, cache_folder=None, config=None,
+                 max_threads=None):
```
AleksMat: Parameter `max_threads` is not passed forward to parent class.

io/eolearn/io/processing_api.py
```diff
@@ -230,3 +293,75 @@ def _extract_bands(self, eopatch, images, shape):
...
+        :param bands_feature: Target feature into which to save the downloaded images.
+        :type bands_feature: tuple(sentinelhub.FeatureType, str)
+        :param bands: An array of band names.
+        :type bands: list[str]
+        :param additional_data: A list of additional data to be downloaded, such as SCL, SNW, dataMask, etc.
+        :type additional_data: list[tuple(sentinelhub.FeatureType, str)]
+        :param maxcc: Maximum cloud coverage.
+        :type maxcc: float
+        :param time_difference: Minimum allowed time difference, used when filtering dates, None by default.
+        :type time_difference: datetime.timedelta
```
AleksMat: These docs strings need to be updated / removed.

io/eolearn/io/processing_api.py
```diff
@@ -98,64 +65,118 @@ def execute(self, eopatch=None, bbox=None, time_interval=None):
...
+        :param config: An instance of SHConfig defining the service
+        :type config: SHConfig or None
+        :param max_threads: Maximum threads to be used when downloading data.
+        :type max_threads: int
+        :param bands_dtype: dtype of the bands array
+        :type bands_dtype: np.dtype
+        :param single_scene: If true, the service will compute a single image for the given time interval using
+                             mosaicing.
+        :type single_scene: bool
+        :param mosaicking_order: Mosaicing order, which has to be either 'mostRecent', 'leastRecent' or 'leastCC'.
```
AleksMat: Typo: "Mosaicking" instead "Mosaicing"

io/eolearn/io/processing_api.py
```diff
@@ -98,64 +65,118 @@ def execute(self, eopatch=None, bbox=None, time_interval=None):
...
 
         LOGGER.debug('Downloading %d requests of type %s', len(requests), str(self.data_source))
-        LOGGER.debug('Downloading bands: [%s]', ', '.join(self.all_bands))
         client = SentinelHubDownloadClient(config=self.config)
         images = client.download(requests, max_threads=self.max_threads)
         LOGGER.debug('Downloads complete')
 
-        images = ((img['default.tif'], img['userdata.json']) for img in images)
-        images = [(img, meta.get('norm_factor', 0)) for img, meta in images]
+        temporal_dim = len(timestamp) if timestamp else 1
```
AleksMat: This one is important: 
As far as I can see `timestamp` variable is never passed to `eopatch.timestamp`.

Besides fixing this we should also improve unit tests so that they would test such a bug.

io/eolearn/tests/test_processing_dem.py
```diff
@@ -0,0 +1,44 @@
...
+        )
+
+        eopatch = task.execute(bbox=self.bbox, time_interval=self.time_interval)
+        dem = eopatch[(FeatureType.MASK_TIMELESS, 'DEM')]
+
+        import ipdb; ipdb.set_trace()
+
+
+if __name__ == "__main__":
+    unittest.main()
```
AleksMat: `SentinelHubProcessingDEM` does not exist. This entire module should probably be removed as you already test the dem task in `test_processing_io.py`.

io/eolearn/io/__init__.py
```diff
@@ -7,6 +7,6 @@
     DEMWMSInput, DEMWCSInput, AddSen2CorClassificationFeature
 from .geopedia import AddGeopediaFeature
 from .local_io import ExportToTiff, ImportFromTiff
-from .processing_api import SentinelHubInputTask
+from .processing_api import SentinelHubInputTask, SentinelHubDEMInputTask
```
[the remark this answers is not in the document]
iovsn: I will go with `SentinelHubDemTask`.

io/eolearn/io/processing_api.py
iovsn: I keep repeating this mistake...

io/eolearn/io/processing_api.py
iovsn: Done.

io/eolearn/io/processing_api.py
iovsn: Done.

io/eolearn/tests/test_processing_dem.py
iovsn: Yes, this should have been removed. Done.

io/eolearn/io/processing_api.py
iovsn: Done.

io/eolearn/io/processing_api.py
iovsn: I agree.

io/eolearn/io/processing_api.py
iovsn: Done :)

io/eolearn/io/processing_api.py
iovsn: Done.

PR #179, opened 2020-01-24
changing dtype of DEM

mlubej:
Based on the documentation [here](https://docs.sentinel-hub.com/api/latest/#/data/DEM?id=dataset-specific-constraints), the suggested `dtype` of DEM is `float32`, due to possible negative values that can occur. 

Since this is `data_timeless`, I guess it's not such a problem to have it as `float32`?

### DIFF
--- a/io/eolearn/io/processing_api.py
+++ b/io/eolearn/io/processing_api.py
@@ -369,7 +369,7 @@ class SentinelHubDemTask(SentinelHubInputBase):
                     output:{
                         id: "default",
                         bands: 1,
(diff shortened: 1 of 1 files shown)


iovsn:
You should also set the DEM feature array to `np.float32`. Check the last line of `processing_api.py`.

mlubej:
> You should also set the DEM feature array to `np.float32`. Check the last line of `processing_api.py`.

Done, thanks!

mlubej:
Turns out it's not a problem, closing PR.

PR #181, opened 2020-01-28
Minor processing api fix.

bsircelj:
Currently the only way to download data is to provide bbox and time interval. If only the eopatch is provided as argument, the bbox and time interval are never initialized. I adjusted the code so that in case bbox and interval are set to None and eopatch is not None the eopatch's bbox and timestamp are used.


AleksMat:
Thanks for the fix. :+1:

PR #184, opened 2020-02-05
Bugfix: added evalscript version

iovsn:
Sentinel-hub service has recently been updated and now `eo-learn`'s Processing API requests fail, because they didn't include evalscript version.

In this PR, evalscript version is added.


veseln:
Seems ok.

PR #187, opened 2020-02-06
Correct function call in Docstrings

raajtilaksarma:
Corrected function calls in Docstrings of functions `MapFeatureTask` and `ZipFeatureTask`. Previously `multiply` function was called where it should have been `maximum` and `calc` wherever required.


AleksMat:
Thanks for the fixes. Merged to develop. :+1:

PR #205, opened 2020-05-14
Docs/m2r fix

AleksMat:
I tried a few different options and this was the only one that worked:

Package `m2r` is outdated but also essential for our build. Therefore I replaced it with my fork of a fork that fixes issues in `m2r`. The solution is not optimal but it should hold until newer versions of Sphinx don't break `m2r` again.

I also updated `.readthedocs.yml` file to version 2 of their style.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -4,11 +4,11 @@ channels:
 dependencies:
 - cartopy=0.17.0
 - pip:
-  - Sphinx==2.4.4
+  - Sphinx
   - sphinx-rtd-theme
   - nbsphinx
   - ipython
-  - m2r>=0.2.0
+  - git+https://github.com/CrossNox/m2r@dev
 
   - ./../core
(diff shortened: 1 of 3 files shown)


docs/environment.yml
```diff
@@ -4,11 +4,11 @@ channels:
 dependencies:
 - cartopy=0.17.0
 - pip:
-  - Sphinx==2.4.4
+  - Sphinx
   - sphinx-rtd-theme
   - nbsphinx
   - ipython
-  - m2r>=0.2.0
+  - git+https://github.com/AleksMat/m2r@dev
```
mlubej, line 11: Did you make any specific changes here? or why is the official package not ok

mlubej: Ah, sorry, it's in the description.

mlubej:
Builds successfully.

PR #210, opened 2020-06-07
Enable ExportToTiff and ImportFromTiff to handle export/import to/from file list

wouellette:
The rational behind this pull request is to enable the `local_io` methods to accept file lists as inputs/outputs, both for `time_dependent` and `timeless` features.
This need arose from having to load the generated output from the [Batch API](https://docs.sentinel-hub.com/api/latest/#/BATCH_API/batch_processor) (GeoTIFFs in S3) into an EOPatch with its associated timestamps using `ImportToTiff`. However, since the `ImportToTiff` method was improved to enable loading a list of file to an `EOPatch`, I also made the `ExportToTiff` method allow exporting to multiple GeoTiff files.

For `time_dependent` features, the files provided must each correspond to a `timestamp`. Each file from the file list provided will therefore end up in its respective `timestamp` for the `ImportToTiff` method, and reciprocally for `ExportToTiff`, each `timestamp` will end up in its respective GeoTIFF file. Additionally, if a single `filename` or `folder` is passed and the `split_time` parameter is set to `True`, the timestamp in ISO string format will be appended to the filenames in output.
 This pull request does not cover the importing/exporting of individual `channels`, only timestamps.

For `timeless` features, the behavior of `ExportToTiff` does not change, and those features still get exported to a single GeoTIFF (providing a list of files for that method is not allowed if a `timeless` features is targetted). For `ImportToTiff` however, if a list of files is provided and a `timeless` feature is specified as `EOPatch` feature, the input files are combined through a `np.nanmean` operation.

From the description above, the following considerations should be taken into account:
- List of input GeoTIFF files could be combined in many other ways for `timeless` features than simply using a `np.nanmean` operation. Therefore, should we even allow the creation of `timeless` features from a list of file? I personally find it useful for my use case, but then maybe we should expand the number of operations that can be performed to aggregate the GeoTIFFs into a `timeless` features (e.g. `np.nansum`) if other use cases can justify it.
- Only allowing the `timestamps` to be concatenated or aggregated is a conscious choice. I do not see a use case for allowing to export/import `channels`.


AleksMat:
I checked the PR and see an issue on the conceptual level:

In order to export multiple tiffs you would now have to pass `ExportToTiff` a list of filenames. That means that you have to know in advance exactly how many timestamps you have to prepare such a list. But that would mean that you could also simply initialize N instances of `ExportToTiff` task and pass a single filename to each of them in order to get all N exported tiffs. Because of that, I don't think we should overload `ExportToTiff` with functionalities which can already be achieved in a different way. And the preferred way in eo-learn is to use multiple simpler tasks rather than 1 complex task to achieve the same result. Similar logic applies to the implemented inverse functionality in `ImportToTiff` task.

There is however a slightly different use case that would benefit from having an option to export multiple tiffs with a single task. It is when you have a large number of EOPatches, each one can have some different number of timestamps and you don't know in advance what these numbers are. Then you might want to export all time slices no matter how many there are. For such a use case you cannot pass a list of filenames but rather a definition of how to generate all filenames. Example
```
filename='image-%Y-%m-%dT%H-%M-%S.tiff'
```
could result in exporting images
```
image-2020-06-01T11-12-13.tiff,
image-2020-06-06T11-12-13.tiff,
image-2020-06-11T11-12-13.tiff,
...
```
This is something I think would be useful to have.

wouellette:
Hi @aleksMat, 

Apologies for the delay in response. I guess summer happened... !

I made an implementation which I think implements the use case of exporting a time dependent eopatch to separate geotiffs corresponding to a unique timestamp.

The current implementation relies on providing a `*` wildcard or a `datetime.strftime` substring as part of the `folder` or `filename` parameters. I figured it would be more thorough than providing a filename without extension, because that may not mean you want to export the data to separate timestamps. In fact, if the extension is not provided, I made it that `.tif` gets automatically appended to the filename.

AleksMat (approved):
Hi @wouellette,

You have perfect timing - at the moment I'm working on something closely related to this. I'm adding a task to import batch results from s3 bucket into EOPatches. Therefore it's great that we merge this and I can maybe extend this approach to `ImportFromTiff` task as well.

I checked your newest changes and I agree with the implementation. I think it's ready to merge. Thanks for your help! :smiley:

PR #211, opened 2020-06-14
MERGE_FEATURES overwrite permission type for the save method

### LINKED ISSUE #80: Append new data to an existing EOPatch saved to disk
I have been acquiring Sentinel-2 data sand saving the EOPatch to disk using the following: 

```  
input_task = S2L1CWCSInput('TRUE-COLOR-S2-L1C', resx='20m', resy='20m', maxcc=maxcc, instance_id=INSTANCE_ID, time_difference=datetime.timedelta(days=1))
add_bands = S2L1CWCSInput(layer='BANDS-S2-L1C', resx='20m', resy='20m', maxcc=maxcc, instance_id=INSTANCE_ID, time_difference=datetime.timedelta(days=1))

save = SaveToDisk(str('data/eopatch'), overwrite_permission=OverwritePermission.OVERWRITE_PATCH)

time_interval = [start_dat, end_dat]

workflow = LinearWorkflow(input_task, add_bands, save)
result = workflow.execute({
    input_task: {
        'bbox': bbox,
        'time_interval': time_interval
    },
    add_bands: {
        'bbox': bbox,
        'time_interval': time_interval
    },
    save:{'eopatch_folder':'eopatch_{}'.format(sitename)}
})
```

I now wish to change `start_dat` and `end_dat` in my time_interval - saving any additional data to the already created EOPatch on disk. Changing the OverwritePermission to eg. OverwritePermission.ADD_ONLY has been unsuccessful.

Is there an efficient way to achieve this?

wouellette:
This is a pull request aimed at allowing temporal merging of a filesystem eopatch together with a newly generated in-memory eopatch which may contain one the following feature types in common:
`DATA`, `MASK`, `SCALAR`, `DATA_TIMELESS`, `MASK_TIMELESS`, `SCALAR_TIMELESS`.

The other feature types are out of scope as they are non-array features, and are therefore heterogeneous and too complex to merge.

The requirements for temporally merging an in-memory EOPatch with a filesystem EOPatch (or at least sub-set of their features) are that all dimensions (besides the `t` dimension) match.

It is in response to [issue #80](https://github.com/sentinel-hub/eo-learn/issues/80), but it does not limit itself to appending data, it also combines timeless features via the `np.nanmean` method, which is why the overwrite permission type was chosen to be `MERGE_FEATURES` rather than `APPEND_FEATURES`.

Whether to cover the merging of `timeless` features is debatable and could be scrapped from this pull request, but I thought the functionality would be interesting, much like in the case of [PR 210](https://github.com/sentinel-hub/eo-learn/pull/210).

closes #80


core/eolearn/core/eodata_io.py
```diff
@@ -154,6 +158,45 @@ def walk_eopatch(eopatch, patch_location, features=...):
...
+                        array = np.concatenate((eopatch[feat], fs_eopatch[feat][mask]), axis=0)
+                        eopatch[feat] = array[sorted_indices]
+            elif ftype in [FeatureType.DATA_TIMELESS, FeatureType.MASK_TIMELESS, FeatureType.SCALAR_TIMELESS]:
+                for fname in eopatch[ftype]:
+                    feat = (ftype, fname)
+                    if eopatch[feat].shape == fs_eopatch[feat].shape:
+                        eopatch[feat] = np.nanmean((eopatch[feat], fs_eopatch[feat]), axis=0)\
+                            .astype(eopatch[feat].dtype)
+
+        eopatch.timestamp = sorted(timestamp)
```
mlubej, line 194: this line is repeated at line 178

mlubej:
Hi @wouellette , thanks for adding this, definitely useful.

One comment. For example, the `save` functionality in `eodata_io` is also used in the form of the EOPatch class method https://github.com/sentinel-hub/eo-learn/blob/master/core/eolearn/core/eodata.py#L487 

Could you prepare such a method, so that one could do something like

```python
eop.merge(eop_other, features_other)
```

or something like that?

core/eolearn/core/eodata_io.py
```diff
@@ -154,6 +158,45 @@ def walk_eopatch(eopatch, patch_location, features=...):
...
+
+    intersection = filesystem_features.intersection(eopatch_features)
+    if intersection:
+        time_periods = [(fs_eopatch.timestamp[0].strftime("%Y-%m-%d"), fs_eopatch.timestamp[-1].strftime("%Y-%m-%d")),
+                        (eopatch.timestamp[0].strftime("%Y-%m-%d"), eopatch.timestamp[-1].strftime("%Y-%m-%d"))]
+
+        mask = np.isin(fs_eopatch.timestamp, list(set(fs_eopatch.timestamp).difference(set(eopatch.timestamp))))
+
+        timestamp = eopatch.timestamp + [tstamp for tstamp, to_keep in zip(fs_eopatch.timestamp, mask) if to_keep]
+        sorted_indices = sorted(range(len(timestamp)), key=lambda k: timestamp[k])
```
mlubej, line 176: just a comment, I'm usually more used to using `np.argsort`, which already gives you the indices of the sorted array, without the need for lambdas. Perhaps it's a bit clearer, but I'm not 100 % sure how it behaves with datetime objects.

core/eolearn/core/eodata_io.py
```diff
@@ -154,6 +158,45 @@ def walk_eopatch(eopatch, patch_location, features=...):
...
+                    feat = (ftype, fname)
+                    if eopatch[feat].shape[1:] == fs_eopatch[feat].shape[1:]:
+                        array = np.concatenate((eopatch[feat], fs_eopatch[feat][mask]), axis=0)
+                        eopatch[feat] = array[sorted_indices]
+            elif ftype in [FeatureType.DATA_TIMELESS, FeatureType.MASK_TIMELESS, FeatureType.SCALAR_TIMELESS]:
+                for fname in eopatch[ftype]:
+                    feat = (ftype, fname)
+                    if eopatch[feat].shape == fs_eopatch[feat].shape:
+                        eopatch[feat] = np.nanmean((eopatch[feat], fs_eopatch[feat]), axis=0)\
+                            .astype(eopatch[feat].dtype)
```
mlubej, line 192: Should there perhaps be some warning printed out which notifies the user which features (if any) were ignored? Do you think this makes sense?

wouellette:
Making it independent from saving makes sense. In such case, one must therefore perform a load of a filesystem eopatch before running the merge operation, which I guess is no problem.

Will work on that and take your other comments into account.

mlubej:
> Making it independent from saving makes sense. In such case, one must therefore perform a load of a filesystem eopatch before running the merge operation, which I guess is no problem.
> 
> Will work on that and take your other comments into account.

Yeah, my thinking was that you already have eopatches at your hand and you just apply the merge then. Otherwise you can prepare them in advance with lazy loading and it also shouldn't be a problem.

wouellette:
Currently reworking the PR, and I was wondering if implementing custom operators such as `time_dependent_op` and `timeless_op` in the fashion of geopandas would be useful.

This would allow the user to choose from operators such as `concatenate`, `mean`, `min`, `max`, `median` mapping respectively to their numpy counterparts. 

Because I can think of several use cases where being able to control how the eopatches are merged (both for `timeless` and `time_dependent` features) would be useful. Default behavior would remain the one in this PR, being `concatenate` for `time_dependent` features and `mean` for `timeless_features`.

Bear in mind a custom list of features would still be passable in order to completely bypass the merging of `timeless` or `time_dependent` features, would the user require that.

AleksMat:
Hi @wouellette,

I agree with the basic functionality of merging before save. Thanks for preparing this. :+1: 

I also agree that merging should be implemented independently of saving as it is also useful for merging 2 patches that are both in memory. In fact, something like that is already implemented in [`EOPatch.concatenate`](https://github.com/sentinel-hub/eo-learn/blob/v0.7.5/core/eolearn/core/eodata.py#L422) method. It's not well documented but what it does is the following:

- features that exist in only 1 of the 2 input patches are directly put into the output patch,
- if a timeless feature exists in both patches it has to have exactly the same values in both patches to be put into the output patch, otherwise, an error is raised,
- if a time-dependent feature exists in both patches we concatenate both arrays along the time axis and this becomes a feature in the output patch.

What is still missing:
- Instead of just concatenating time-dependent feature arrays they should be merged according to timestamps just like it is implemented in your `merge_eopatch` function. I think it is safe to make such change as the current "simple concatenation" is not very useful.
- Support for time-dependent vector features.

Therefore I suggest that the current functionality of `merge_eopatch` function is integrated into `EOPatch.concatenate` in order to have only one well-defined and well-functional way of joining patches. Or even better - we could join functionalities of `merge_eopatch` and `EOPatch.concatenate` into a new method `EOPatch.merge` and deprecate `EOPatch.concatenate`. This way we get a bit more accurate method name and make a more user-friendly transition.

Regarding operators `mean`, `min`, `max`, `median` that you described I think these are actually different ways how to join features that are overlapping. Currently a default one in `EOPatch.concatenate` is "if values are different, raise error", but maybe a user could be able to decide to rather join values with one of these operations. Therefore the new method could simply have an optional parameter to specify in which way to join values. This way there would be very little implementational overhead.

Let us know your opinion about this and if you would like any help with this. I might be able to find some time for this during next few weeks.

wouellette:
Hey Matej, 

Thanks for the comprehensive review. 
I agree with all of the above, and I agree that we should replace `EOPatch.concatenate` by `EOPatch.merge` with the functionalities in scope of this PR.

I'll add support for time-dependent vector features too.

Regarding the operators, what you describe is what I had in mind: add additional conditions depending on the operator parameter passed.

You didn't share your thought on merging timeless features, is that still desirable? I find it useful personally, especially if you are able to define which operator to apply to perform this merging of timeless features.

Therefore, I suggest the following two parameters: `time_dependent_op` and `timeless_op`, as one may want to have the flexibility of applying different operators for time-dependent and timeless respectively, with the following allowed operations:
    `allowed_timeless_ops = ["mean", "max", "min", "median"]`
    `allowed_time_dependent_ops = ["concatenate", "mean", "max", "min", "median"]`

I am still in doubt of allowing timeless operations, because they can get quick complex, and maybe its better to have the user perform this operation outside with regular `numpy` gynmnastics rather than as part of the `EOPatch` method. If you agree with that, we just keep it to a single `op` parameter which controls time-dependent features only.

AleksMat:
Excellent, I think we could have operations for both timeless and time-dependent features. To make sure we are on the same page here is a more accurate description of how I imagine merging could work:

- Merging of timeless features could happen if and only if both arrays have exactly the same shape. In that case, the possible ways of merging could be:
  * Both arrays have to be the same, else raise an error.
  * Apply min, max, or mean on pairs of corresponding values from both arrays. Now that I think about it, a median of 2 values is actually the same as the mean so I would exclude it. (These types of merging would however not be applied on vector features.)
  * Concatenate both arrays over the last dimension (i.e. bands dimension). In this case, arrays could actually have different size of the last dimension.

- Merging of time-dependent features would always happen along the time axis and shapes of arrays would have to match along other dimensions. Timestamps would be sorted in the correct order and that would also be the order in which time slices in the merged array would be sorted. Then there are different options on how to handle those time slices that have the same timestamp:
  * Both time slices have to have the same values to be joined into a single time slice, else raise an error.
  * Apply min, max or mean on pairs of corresponding values from both time slices to join them into a single one.
  * Leave both time slices as they are in the merged array with a duplicated timestamp (i.e. just concatenation).

Is that also how you imagined the merging would work?

wouellette:
A couple of comments:
- timeless concatenation on the `band` dimension: I don't know what the use case would be, but I guess I don't see the problem of doing it if the choice can be motivated.
- time-dependent concatenation: is duplicating timestamps a good practice? I personnally have never done it (same as above, what is use case?), and my initial idea was to limit the time-dependent concatenation to either an operation, either joining the duplicate time slices into a single one if they match.

Once I get clarification i'll finalize the PR.

AleksMat:
For now, I also don't see a clear use-case for either timeless or time-dependent concatenation. Therefore I agree not to implement "concatenation" operator in this iteration. We can keep it in mind and if a use-case ever appears the operator could always be added then without breaking anything that will be implemented now.

wouellette:
Concatenation remains useful for time-dependent features, as long as timestamps are not duplicated, then they basically get interleaved along the `time` dimension. So for concatenation, if there are two arrays of same timestamp which are not identical, I would make it fail.

wouellette:
I pushed a new implementation with some very light tests, just to check that our understanding of the functionality is on par.

mlubej:
Closes #80

core/eolearn/core/eodata_io.py
```diff
@@ -155,6 +156,136 @@ def walk_eopatch(eopatch, patch_location, features=...):
...
+    return eopatch
+
+
+def _perform_merge_operation(eopatch, eopatch_content, sorted_indices, time_dependent_op, timeless_op):
+    """ Performs the merging of duplicate timestamps of time-dependent features and of timeless features.
+    """
+    for feat, arrays in eopatch_content.items():
+        ftype, _ = feat
+        if ftype.is_time_dependent() and ftype != FeatureType.VECTOR and sorted_indices:
+            if time_dependent_op == 'mean':
```
AleksMat, line 254: To avoid all these if-else statements you can have a dictionary mapping operation strings to numpy functions: `{'mean': np.nanmean, ...}`.

core/eolearn/core/eodata_io.py
```diff
@@ -155,6 +156,136 @@ def walk_eopatch(eopatch, patch_location, features=...):
...
+
+def _check_duplicate_timestamps(eopatch_content, duplicate_timestamp):
+    """ Checks that no duplicate timestamps with different values exist
+    """
+    for dup_time in duplicate_timestamp:
+        for dup_feat in eopatch_content.keys():
+            arrays = [eopatch_content[feat][eopatch_content[(FeatureType.TIMESTAMP, ...)].index(dup_time)]
+                      for eop_idx, (feat, _) in enumerate(eopatch_content.items())
+                      if dup_time in eopatch_content[(FeatureType.TIMESTAMP, ...)] and dup_feat == feat]
+            if any(not deep_eq(x, y) for i, x in enumerate(arrays) for j, y in enumerate(arrays) if i != j):
```
AleksMat, line 285: These features checks should be done in the same place as merging operations are done (i.e. `_perform_merge_operation` in your case). After all, you can imagine "take 1 if all equal otherwise raise an error" as an operation, similar to min, max, mean, median.

wouellette:
I implemented the logic as you have described as part of `_perform_merge_operation`. 

Let me know if it matches what you had in mind, and I can make the tests more robust.

PR #220, opened 2020-07-03
Feat/docker images

AleksMat:
I prepared 2 `eo-learn` docker images and connected them with a (semi-)automatic build and deploy procedure at Docker Hub. I also documented the instructions.

Issues which I'm leaving for future improvements:
- [Notebook with cartopy visualizations](https://github.com/sentinel-hub/eo-learn/blob/master/examples/visualization/EOPatchVisualization.ipynb) doesn't work in the built docker image (kernel keeps dying).
- Build procedure at Docker Hub is not completely automatic as it doesn't allow build chaining. In our case image `latest-examples` depends on image `latest` but Docker Hub can only trigger their build at the same time. Hence `latest-examples` will have to be triggered manually.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git
+docs
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@
 [![License](https://img.shields.io/pypi/l/eo-learn.svg)](https://github.com/sentinel-hub/eo-learn/blob/master/LICENSE)
 [![Overall downloads](http://pepy.tech/badge/eo-learn)](https://pepy.tech/project/eo-learn)
 [![Last month downloads](https://pepy.tech/badge/eo-learn/month)](https://pepy.tech/project/eo-learn)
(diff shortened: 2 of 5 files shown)


batic:
👍

batic:
I've checked the issue with visualisation notebook; uninstalling shapely and reinstalling it with `--no-binary` fixed the kernel crashes.

After that, another issue was found:

```
/usr/local/lib/python3.8/site-packages/eolearn/visualization/xarray_utils.py in new_coordinates(data, crs, new_crs)
    230     bbox = BBox((x_values[0], y_values[0], x_values[-1], y_values[-1]), crs=crs)
    231     bbox = bbox.transform(new_crs)
--> 232     xmin, ymin = bbox.get_lower_left()
    233     xmax, ymax = bbox.get_upper_right()
    234     new_xs = np.linspace(xmin, xmax, len(x_values))

AttributeError: 'BBox' object has no attribute 'get_lower_left'
```

AleksMat:
Thanks @batic, I was able to fix visualization issues and the notebook now runs correctly. 

Note: we'll have to release new eo-learn version before images will build correctly. For testing purposes the following 2 lines have to be additionally used in build
```
COPY ./visualization ./visualization
RUN pip3 install ./visualization --upgrade
```

PR #225, opened 2020-07-07
Added crop type classification example notebook and folder structure

Gnilliw:


Gnilliw:
Hi Anze, thanks for your comments. I inserted almost all of them. 
I struggled fixing two things.
First: How did you get these information about the aoi polygon. If I use "aoi.crs" there is much less information.



> In section 1., cell that starts with `# define Area of interest` the reprojection to UTM is not needed:
> 
> ```python
> aoi = aoi.to_crs(crs={'init': CRS.ogc_string(crs)}) # assign coordinate system to 
> ```
> 
> Remove this line since the INPUT_FILE is already in UTM_33N. It is however strange that crs attribute of aoi is falsely set to WGS84.
> 
> ```python
> > aoi.crs
> 
> <Geographic 2D CRS: EPSG:4326>
> Name: WGS 84
> Axis Info [ellipsoidal]:
> - Lat[north]: Geodetic latitude (degree)
> - Lon[east]: Geodetic longitude (degree)
> Area of Use:
> - name: World
> - bounds: (-180.0, -90.0, 180.0, 90.0)
> Datum: World Geodetic System 1984
> - Ellipsoid: WGS 84
> - Prime Meridian: Greenwich
> ```
> 
> The coordinates are however in UTM:
> 
> ```python
> > print(aoi.loc[0].geometry)
> 
> POLYGON ((441526.2580528425 5350098.248442201, 439196.0130617045 5331945.929277981, 420862.3179607267 5329537.336315366, 422411.038396594 5347842.713809402, 441526.2580528425 5350098.248442201))
> ```
> 
> Please check and fix this inconsistency.


Second: I dedicated the task because I needed to get the individual amount of samples and the respective sampled lables for each EOPatch inside a workflow. I did not see a solution eo-learn is offering. 


> Why is dedicated sampling task needed? Could the one from eolearn be used intead? In what way would it have to be modified?

azupanc:
Frank, thank you very much for your contribution. I believe it will be very much appreciated by the community.

PR #230, opened 2020-07-20
Expose option to add `other_args` to SH input data object

mlubej:
- exposed parameter which is added to SH input data method
- added a test
- disabled a pylint for testing number of parameters (one to many :/)

I hope that's OK.

### DIFF
--- a/io/eolearn/io/processing_api.py
+++ b/io/eolearn/io/processing_api.py
@@ -164,11 +164,10 @@ class SentinelHubInputBase(EOTask):
 ProcApiType = collections.namedtuple('ProcApiType', 'id unit sample_type np_dtype feature_type')
 
 
-# pylint: disable=R0913
 class SentinelHubInputTask(SentinelHubInputBase):
     """ A processing API input task that loads 16bit integer data and converts it to a 32bit float feature.
     """
-
+    # pylint: disable=too-many-arguments
     PREDEFINED_BAND_TYPES = {
         ProcApiType("bool_mask", 'DN', 'UINT8', np.bool, FeatureType.MASK): [
             "dataMask"
--- a/io/eolearn/io/processing_api.py
+++ b/io/eolearn/io/processing_api.py
@@ -190,7 +190,7 @@ class SentinelHubInputTask(SentinelHubInputBase):
 
     def __init__(self, data_source, size=None, resolution=None, bands_feature=None, bands=None, additional_data=None,
                  maxcc=1.0, time_difference=None, cache_folder=None, max_threads=None, config=None,
-                 bands_dtype=np.float32, single_scene=False, mosaicking_order='mostRecent', aux_input_args=None):
+                 bands_dtype=np.float32, single_scene=False, mosaicking_order='mostRecent', aux_request_args=None):
         """
         :param data_source: Source of requested satellite data.
         :type data_source: DataSource
@@ -222,8 +222,8 @@ class SentinelHubInputTask(SentinelHubInputBase):
         :type single_scene: bool
         :param mosaicking_order: Mosaicking order, which has to be either 'mostRecent', 'leastRecent' or 'leastCC'.
         :type mosaicking_order: str
-        :param aux_input_args: a dictionary with auxiliary information for the input_data part of the SH request
-        :type aux_input_args: dict
+        :param aux_request_args: a dictionary with auxiliary information for the input_data part of the SH request
+        :type aux_request_args: dict
         """
         super().__init__(
             data_source=data_source, size=size, resolution=resolution, cache_folder=cache_folder, config=config,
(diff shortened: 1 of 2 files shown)


mlubej:
Not sure why Travis is showing incompleted tasks, they have finished successfully.

io/eolearn/io/processing_api.py
```diff
@@ -164,6 +164,7 @@ def _add_meta_info(self, eopatch):
 ProcApiType = collections.namedtuple('ProcApiType', 'id unit sample_type np_dtype feature_type')
 
 
+# pylint: disable=R0913
```
AleksMat, line 167: Could you add this into `__init__` method right after docs strings? If you leave it here outside I think it is applied to all functions and methods in this module.

I also recommend changing to `# pylint: disable=too-many-arguments`. This way it will be more human-readable than `R0913`.

io/eolearn/tests/test_processing_io.py
```diff
@@ -229,6 +229,24 @@ def test_additional_data(self):
...
+            data_source=DataSource.SENTINEL2_L1C,
+            max_threads=self.max_threads,
+            aux_input_args={'dataFilter': {'previewMode': 'PREVIEW'}}
+        )
+
+        eopatch = task.execute(bbox=self.bbox, time_interval=self.time_interval)
+        bands = eopatch[(FeatureType.DATA, 'BANDS')]
+
+        self.assertTrue(bands.shape == (4, 4, 4, 13))
+
```
AleksMat, line 249: Technically, this test has little to no effect whatsoever. I know this because changes in sentinelhub-py have not been released yet. Therefore CI is using the previous sentinelhub-py version, the preview parameter gets ignored but the test still passes.

Could we add a check that would make this test fail if the preview parameter is ignored?

io/eolearn/io/processing_api.py
```diff
@@ -190,7 +191,7 @@ class SentinelHubInputTask(SentinelHubInputBase):
 
     def __init__(self, data_source, size=None, resolution=None, bands_feature=None, bands=None, additional_data=None,
                  maxcc=1.0, time_difference=None, cache_folder=None, max_threads=None, config=None,
-                 bands_dtype=np.float32, single_scene=False, mosaicking_order='mostRecent'):
+                 bands_dtype=np.float32, single_scene=False, mosaicking_order='mostRecent', aux_input_args=None):
```
AleksMat, line 194: I just wonder if name `aux_input_args` is clear enough. What about `aux_data_source_args`? It's even more verbose but it tells you parameters are connected with a data source.

Feel free to disregard this comment if you don't agree.

io/eolearn/tests/test_processing_io.py
mlubej: > Could we add a check that would make this test fail if the preview parameter is ignored?

But this would then always fail and I don't see the benefit of checking for a raised exception which is there by design.

> Therefore CI is using the previous sentinelhub-py version, the preview parameter gets ignored but the test still passes.

I don't understand how this enables the test to pass. Can you elaborate? In any case, when the changes are released, this test will make sense, right?

io/eolearn/io/processing_api.py
mlubej: I'm not sure the `data_source` part makes it any clearer. Perhaps `aux_sh_request_args` in this case? In any case, hopefully users check the docstring if anything is unclear. I'm not sure you can explain the usage of this parameter with the name itself anyway.

io/eolearn/tests/test_processing_io.py
AleksMat: It passes because it doesn't 

> > Could we add a check that would make this test fail if the preview parameter is ignored?
> 
> But this would then always fail and I don't see the benefit of checking for a raised exception which is there by design.

I meant to add a check that would fail when running with the old sentinelhub-py version that ignores preview but will pass with the new version that actually includes the preview parameter in a request.

> 
> > Therefore CI is using the previous sentinelhub-py version, the preview parameter gets ignored but the test still passes.
> 
> I don't understand how this enables the test to pass. Can you elaborate? In any case, when the changes are released, this test will make sense, right?

The test passes because it only checks that it got something back from the service but not what. So whatever parameters you would specify the test would still pass. So does it really make sense to have a test that always passes?

io/eolearn/io/processing_api.py
AleksMat: A name `aux_request_args` seems clear for me. (`sh` in the name is sort of implicit since you are using `SentinelHubInputTask`).

io/eolearn/tests/test_processing_io.py
mlubej: Resolved in private.

AleksMat:
Changes for this in `sentinelhub-py` have been released, tests now pass.

PR #146, opened 2019-09-27
WIP: Added edge detection task from D4.6.

jO-Osko:
Added edge detection task from D4.6. Additional tests are needed for the Thresholding task.

PR #234, opened 2020-07-23
Edge extraction task from deliverable 4.7

bsircelj:
This is different edge task from the previous one. It has different approach (Canny) to computing edges and it computes a timeless mask of of edges from single patch based on multiple features during the whole year.

This mask was used for masking edges of parcels for LPIS classification to avoid any potential mixed class instances.

Mask is computed in several steps:
- Individual image edge calculation. 
    + Each image is firstly blurred with a Gaussian filter (`cv2.GaussianBlur`), then edges are computed using edge detector `cv2.Canny`), finally dilation and erosion are applied for filling potential holes.
- Individual image weight calculation
    + Each edge pixel's contribution is adjusted based on that feature's values in the vicinity. The weights are  calculated by normalizing and blurring image with a Gaussian filter (`cv2.GaussianBlur`).
- Yearly feature mask calculation by joining single weighted edge images for each feature
    + Weight mask is calculated by summing all weights for each pixel through the whole year. The pixels where do not contain an edge have a weight of 0, those who do, have a weight proportional to the weight calculated in the previous step. The idea of this is that we should prioritize edges which were calculated during the high vegetation period and ignore edges calculated during the off season eg. winter.
- Final temporal mask calculation by joining all the yearly feature masks
    + Pixels are included only, if total sum of all feature's weights for that pixel exceeds the weight threshold.


bsircelj:
Mask computed on the test patch with default parameters.
test_edges

mlubej:
@bsircelj is #146 still relevant or does this PR replace it?

bsircelj:
It is as it has different methods. Here is emphasis on a single mask computed from multiple features and times but uses just one method. I guess they could be joined so that different methods are also used here. The @jO-Osko's function could be called when the cv2.canny is called here. Although I'm not sure what is the point of all that thresholding options as canny is the standard edge detector that works fine. Also it doesn't seem that @jO-Osko will work on that PR anymore.

zigaLuksic:
Thank you for the contribution. Sadly, due to inactivity, it has become outdated.
I am closing this merge request for now, but feel free to reopen and update it if you wish for the feature to be integrated.

PR #239, opened 2020-07-31
Balanced class sampling D4.7

bsircelj:
A class that samples points from multiple patches and returns a balanced set depending on the class label. This is done by sampling the desired amount on each patch and then balancing the data based on the smallest class or amount. If the amount is provided and there are classes with less than that number of points, random point are duplicated to reach the necessary size. It also supports additional sampling around previously specified weak classes and ignoring certain classes.


bsircelj:
I forgot to consider multiprocessing with EOExecutor. As of now it doesn't work with it. After any reviews I will try to modify the code so it is compatible. Although I'm not sure how I would go about it. I could check the EOExecutor class or find a task that is similar to mine. The shared data between instances would be the sampled_data list where all the samples are stores. The sample() method could be called in parallel. I would be glad to receive any suggestions regarding this issue.

zigaLuksic:
Thank you for the contribution. Sadly, due to inactivity, it has become outdated.
I am closing this merge request for now, but feel free to reopen and update it if you wish for the feature to be integrated.

PR #241, opened 2020-07-31
Add example of ML workflow in custom script

devisperessutti:
This PR adds a notebook example on how to develop an end-to-end machine learning evalscript for water detection.


batic:
👍 
Approving and merging.

PR #245, opened 2020-08-19
Fixed value fillout bug

JernejPuc:
[This line](https://github.com/sentinel-hub/eo-learn/commit/4ad7bba060825c1047095885d420e296a5772e09#diff-43f01a19420680c6833c66a1a7366d22R165) in `ValueFilloutTask` contains a typo (the axis is wrong and `1` is added instead of subtracted), which causes problems in fillout in the backward direction. The result of chaining forward and backward fillout may also be corrupted by the latter operation.

The tests are a bit unfortunate in the sense that they all end up testing the function on data of the same particular form, which happens to work - specifically, where only the first element per row needs to be filled out, e.g.:
```python
>>> fill(np.array([[np.nan, 68., 62, 99., 67., 92., 22., 80.]]), operation='b')

array([[68., 68., 62., 99., 67., 92., 22., 80.]])  # Working as expected
```

However, a simple case such as this would still fail:
```python
>>> fill(np.array([[np.nan, 68., np.nan, 99., 67., 92., 22., 80.]]), operation='b')

array([[68., 68., nan, 99., 67., 92., 22., 80.]])  # The `NaN` in the middle is not filled out
```

After fixing the typo, the results are as expected. I've also simplified the code a bit and added a few more tests.

### DIFF
--- a/features/eolearn/features/feature_manipulation.py
+++ b/features/eolearn/features/feature_manipulation.py
@@ -169,20 +169,21 @@ class ValueFilloutTask(EOTask):
         if operation not in ['f', 'b']:
             raise ValueError("'operation' parameter should either be 'f' (forward) or 'b' (backward)!")
 
-        value_match_mask = np.isnan(data) if np.isnan(value) else (data == value)
-        value_match_true = 0 if operation == 'f' else (value_match_mask.shape[0] + 1)
-        accumulate_function = np.maximum.accumulate if operation == 'f' else np.minimum.accumulate
+        n_rows, n_frames = data.shape
 
-        idx = np.where(value_match_mask, value_match_true, np.arange(value_match_mask.shape[1]))
+        value_mask = np.isnan(data) if np.isnan(value) else (data == value)
+        init_index = 0 if operation == 'f' else (n_frames - 1)
 
-        idx = idx if operation == 'f' else idx[:, ::-1]
-        idx = accumulate_function(idx, axis=1)
-        idx = idx if operation == 'f' else idx[:, ::-1]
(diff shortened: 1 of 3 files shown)


AleksMat (commented):
Everything looks good in the code, only [some Travis tests](https://travis-ci.org/github/sentinel-hub/eo-learn/builds/719201558) fail. I see that you are using `equal_nan` parameter which has only been added recently in `numpy` `1.19.0`. I think for some reason Travis used cached older numpy version when it ran tests for 3.7 and 3.8. To fix this I recommend putting `numpy>=1.19.0` into `requirements-dev.txt` file so that Travis will be forced to upgrade `numpy` version.

I also checked what implications this bug has for our past work. It looks like the bug only appears if `ValueFilloutTask` is used with operations `'bf'` or `'b'`. We used it only with the default `'fb'` and therefore I think nothing went wrong...

JernejPuc:
Using `'fb'` rather than `'bf'` or `'b'` definitely helped - it filled out `NaN` values in between and at the end before the backward direction bug ever got to the data. 

But looking at the bug, I am still surprised that you had not seen errors before. A simple one would be if the 1st axis (in the 2-dimensional data that is operated on) were longer than the 2nd axis (you could get an `IndexError` on the [last line](https://github.com/sentinel-hub/eo-learn/commit/4ad7bba060825c1047095885d420e296a5772e09#diff-43f01a19420680c6833c66a1a7366d22R177) if `idx` went out of range for that axis).

Then again, if the shape of the data were just right (every time that fillout was used), it would not be as obvious. Quite unfortunate...

PR #247, opened 2020-09-07
Feat/data source abstraction

AleksMat:
Changes:
- improvements in `SentinelHubInputTask` to support any data source
- integration tests for more data sources
- pylint fixes

This PR depends on https://github.com/sentinel-hub/sentinelhub-py/pull/138 and should be merged after changes in `sentinelhub-py` will be released.


examples/features/BSplineInterpolation.ipynb
```diff
@@ -97,7 +97,7 @@
   {
    "cell_type": "markdown",
    "metadata": {},
-   "source": [
+   "collection": [
```
batic, line 100: this part should not be changed, as it changes metadata of the notebook, resulting in invalid notebook

examples/io/SentinelHubIO.ipynb
```diff
@@ -93,13 +93,13 @@
...
    "source": [
-    "## Request different types of layers and data sources to an eopatch"
+    "## Request different types of layers and data collections to an eopatch"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
-   "source": [
+   "collection": [
```
batic, line 102: Same here as with BSplineInterpolation notebook.

examples/water-monitor/WaterMonitorWorkflow.ipynb
```diff
@@ -193,7 +193,7 @@
   {
    "cell_type": "markdown",
    "metadata": {},
-   "source": [
+   "collection": [
```
batic, line 196: And here.

io/eolearn/io/processing_api.py
```diff
@@ -61,16 +64,17 @@ def __init__(self, data_source, size=None, resolution=None, cache_folder=None, c
...
-
         if (size is None) == (resolution is None):
             raise ValueError("Exactly one of the parameters 'size' and 'resolution' should be given.")
 
         self.size = size
         self.resolution = resolution
         self.config = config or SHConfig()
         self.max_threads = max_threads
-        self.data_source = data_source
+        self.data_collection = DataCollection(handle_deprecated_data_source(data_collection, data_source))
```
batic, line 77: There is a bit of mismatch how `handle_deprecated...` is used throughout the code. Sometimes it is passed to the  `DataCollection` constructor (as in this case), while some other times it is not. 

I suggest this is consolidated and handled in the utility function (so that it returns `DataCollection`).

io/eolearn/io/processing_api.py
```diff
@@ -218,40 +226,36 @@ def __init__(self, data_source, size=None, resolution=None, bands_feature=None,
...
         if bands_feature:
-            bands_feature = next(self._parse_features(bands_feature, allowed_feature_types=[FeatureType.DATA])())
-
-            if not bands and data_source in [DataSource.SENTINEL2_L1C, DataSource.SENTINEL2_L2A]:
-                bands = data_source.bands()
-            elif data_source.is_custom() and not bands:
-                raise ValueError("For custom data sources 'bands' must be provided as an argument.")
-
+            if not bands:
+                bands = data_collection.bands
```
batic, line 256: I don't think that this ensures `bands` to be empty list. Some collections have no default bands specified.

io/eolearn/io/sentinelhub_service.py
```diff
@@ -65,23 +66,28 @@ class SentinelHubOGCInput(EOTask):
...
+    :param data_source: A deprecated alternative to data_collection
+    :type data_source: DataCollection
     """
 
-    def __init__(self, layer, feature=None, valid_data_mask_feature='IS_DATA', service_type=None, data_source=None,
+    def __init__(self, layer, feature=None, valid_data_mask_feature='IS_DATA', service_type=None, data_collection=None,
                  size_x=None, size_y=None, maxcc=None, image_format=MimeType.TIFF_d32f, instance_id=None,
-                 custom_url_params=None, time_difference=None, raise_download_errors=True):
+                 custom_url_params=None, time_difference=None, raise_download_errors=True, data_source=None):
+        # pylint: disable=too-many-arguments
```
batic, line 76: It is a bad indicator that we have to disable pylint warnings like this :)

io/eolearn/tests/test_processing_io.py
```diff
@@ -1,28 +1,23 @@
...
-# logging.basicConfig(stream=sys.stdout)
-# logging.getLogger("eolearn.io.processing_api").setLevel(logging.DEBUG)
-# logging.getLogger("sentinelhub.sentinelhub_client").setLevel(logging.DEBUG)
-# logging.getLogger("sentinelhub.sentinelhub_rate_limit").setLevel(logging.DEBUG)
 
 def array_stats(array):
     time, height, width, _ = array.shape
     edge1 = np.mean(array[int(time/2):, 0, 0, :])
-    edge2 = np.mean(array[:int(time/2), -1, -1, :])
+    edge2 = np.mean(array[:max(int(time/2), 1), -1, -1, :])
```
batic, line 20: Could you shortly explain why `max` here?

examples/features/BSplineInterpolation.ipynb
AleksMat: fixed

examples/io/SentinelHubIO.ipynb
AleksMat: fixed

examples/water-monitor/WaterMonitorWorkflow.ipynb
AleksMat: fixed

io/eolearn/io/processing_api.py
AleksMat: I have made a small update regarding this in `cloud_mask.py` so that we now have `DataCollection` casting everywhere. In the future, `handle_deprecated_data_source` will be removed and `DataCollection` casting will remain. Because of that I wouldn't put `DataCollection` casting into `handle_deprecated_data_source`.

io/eolearn/io/processing_api.py
AleksMat: If bands are not defined for a given collection this will raise an error. Note that `DataCollection.bands` is actually a property -> [implementation](https://github.com/sentinel-hub/sentinelhub-py/blob/develop/sentinelhub/data_collections.py#L475). It is a bit sneaky but also very convenient to use.

io/eolearn/io/sentinelhub_service.py
AleksMat: Notice that `SentinelHubInputTask.__init__` is explicitly passing arguments to `SentinelHubInputBase.__init__`. Instead `SentinelHubInputTask` could also use and pass `**kwargs`. This way we would have `6` parameters less.

Usually I'm more inclined towards using `**kwargs` but in this case using explicit parameters also has a few advantages. Therefore I don't have a strong opinnion towards any option.

io/eolearn/tests/test_processing_io.py
AleksMat: If `time=1` then `array[:int(time/2), -1, -1, :]` would be an empty array and `np.mean` on an empty array returns NaN with a warning.

In the new tests such a case appears so I updated this in a way that wouldn't break previous test cases.

io/eolearn/io/sentinelhub_service.py
batic: Neither do I. The "proper" way would be passing some sort of (config) object, which is so much cleaner also from the perspective of hinting etc, but that would require quite a chunk of work.

batic:
👍 

This can be merged, either with squashing or without, however you prefer.

Issue #249, opened 2020-09-07
Add a template for issues


mlubej:
Add a template for GIT issues so that users provide issues of higher quality, helping us to be more efficient in solving them.

PR #250, opened 2020-09-09
Update issue templates

devisperessutti:
Add templates for bug reports, feature requests, and a generic help/support report.

Relates to #249


mlubej:
What are the same templates for the different bug templates? can you explain the dashes in the names?

AleksMat:
A couple of minor suggestions:
- Often people write just an error message instead of an entire stack trace. Because of that, it's hard to tell where the error actually happened. Maybe that could fit next to screenshots? The paragraph could even be called "Stack trace or screenshots".
- Sometimes someone includes some privat information, such as Sentinel Hub credentials. Maybe there could be a note with a recommendation to remove any credentials or personal data.

devisperessutti:
> What are the same templates for the different bug templates? can you explain the dashes in the names?

Removed them as they previous versions generated by the UI.

devisperessutti:
> A couple of minor suggestions:
> 
>     * Often people write just an error message instead of an entire stack trace. Because of that, it's hard to tell where the error actually happened. Maybe that could fit next to screenshots? The paragraph could even be called "Stack trace or screenshots".
> 
>     * Sometimes someone includes some privat information, such as Sentinel Hub credentials. Maybe there could be a note with a recommendation to remove any credentials or personal data.

Thanks, I've taken both into account and added a note in bug and help reports.

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,24 @@
+---
+name: "Feature request \U0001F30D"
+about: Contribute to eo-learn
```
mlubej, line 3: Perhaps this is more appropriate if this were a PR. I would suggest "What is missing?" or something. Otherwise, the templates look good and you can merge them. I would suggest using similar templates in sh-py.

PR #253, opened 2020-09-16
Add eo-learn with Batch Processing API example

maximlamare:
I am adding the latest working version of the workflow of LULC classification using Batch Processing API. Then I can link the blog post to the repository.

Note - I haven't accounted for the latest comments of Matej, due to lack of time. The code converting the Batch results to EOPatches could be improved (less nested statements and small lists of inputs in the class).


AleksMat:
I have just noticed - this PR contains a 88MB file. That is huge. This repository already has 500MB git history and by adding large dataset to the repository this only grows that much faster. In turn that slows down various development and CI processes.

Is so large file really needed? Could its size be reduced or could it be loaded from somewhere else (e.g. from a Geopedia layer)?

PR #254, opened 2020-09-21
[Docs] correct BlobTask docstrings

rpitonak:


AleksMat:
Hi @rpitonak, thanks for the fix. Merging to develop.

PR #256, opened 2020-10-02
eopatch returns boolean for mask values by default

Samshal:
For some reason, removing the type conversion causes this function to stop failing on the current version (0.7.7)


AleksMat:
Hi @Samshal, 

I don't see exactly how this change could fix a problem. Could you share a stack trace of an error that you get when applying a type conversion? I would like to understand the problem a bit better before we merge.

AleksMat:
Closing PR as everything seems to be working in new versions.

PR #257, opened 2020-10-06
Filesystem support for ExportToTiff and ImportFromTiff tasks

AleksMat:
This feature enables using both tasks to load data from s3 by simply specifying `folder='s3://my-bucket/path'`.

Additionally, I have also added a few minor filesystem-related improvements in the core subpackage.


devisperessutti (approved):
looked at the code and ran the tests, which cover very well the new added features

tried to replace existing code with this for ingesting tiffs from batch, but realised we deleted the data :/ 

no additional comments from my side :clap:

PR #266, opened 2021-01-07
Fix and improve coregistration

mlubej:
- a bug was discovered in the coregistration where the warp matrix used has not been inverted in order to correct for the misalignments, causing them to increase instead of fixing them
- additionally, the algorithm was improved by calculating the coregistration matrix from the normalized gradients of the input images
- ~~the reference image has been set to be the last frame~~

### Original

org_coreg_anim

#### Using translation + rotation

bug_tr_coreg_anim

Very weird effect

#### Using just translation

bug_t_coreg_anim

Also weird effect, probably due to the changing reference, where the reference itself was affected with the bug

#### Using just translation with last frame as reference

bug_t_lf_coreg_anim

The last frame was used as a reference and it was not altered at all, you can see the effects of increased misalignments, which  points into the direction of a non-inverted matrix

### After improvements

new_coreg_anim

Everything is OK after the matrix has been inverted

### DIFF
--- a/coregistration/eolearn/coregistration/coregistration.py
+++ b/coregistration/eolearn/coregistration/coregistration.py
@@ -71,6 +71,7 @@ class RegistrationTask(EOTask, ABC):
         :param params: Any other registration setting which will be passed to registration method
         :type params: object
     """
+
     def __init__(self, registration_feature, channel=0, valid_mask_feature=None, apply_to_features=...,
                  interpolation_type=InterpolationType.CUBIC, **params):
         self.registration_feature = self._parse_features(registration_feature, default_feature_type=FeatureType.DATA)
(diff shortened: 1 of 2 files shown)


batic:
😍

I also quickly checked the code an give a 👍.

mlubej:
- reverted back so that two neighbouring timestamps are used, otherwise time effects can cause divergence.
- also added a safety net in case gradients can't be calculated

mlubej:
@devisperessutti could I kindly ask you to give a final look and a blessing to merge?

devisperessutti (approved):

devisperessutti:
you have my blessing :) excellent work

I think this module would benefit from better tests and further evaluation wrt optimal working conditions. Hopefully we'll have time to explore this in the future

PR #272, opened 2021-03-01
Feat/manage windows paths

DomagojKorais:


core/eolearn/core/fs_utils.py
```diff
@@ -60,7 +60,8 @@ def get_base_filesystem_and_path(*path_parts, **kwargs):
         return get_filesystem(filesystem_path, **kwargs), relative_path
 
     entire_path = os.path.abspath(os.path.join(*path_parts))
-    posix_path = str(PurePosixPath(entire_path))
+    pure_path = PurePath(entire_path)
+    posix_path = str(pure_path.relative_to(pure_path.anchor).as_posix())
```
AleksMat, line 64: Since `as_posix` method already returns a string then additional casting with `str(...)` is not necessary.

AleksMat (commented):
Looks good, only `1` minor comment.

core/eolearn/core/fs_utils.py
DomagojKorais: casting to string removed.

PR #274, opened 2021-03-02
Feat/allow keyboard interrupt

DomagojKorais:
Now the `KeyboardInterrupt` exception is propagated, and this stops the execution of the program, instead of processing the next element in the `execution_args`.

### DIFF
--- a/core/eolearn/core/eoexecution.py
+++ b/core/eolearn/core/eoexecution.py
@@ -137,7 +137,6 @@ class EOExecutor:
 
         if workers == 1:
             processing_type = 'single process'
-            print(processing_type)
             self.execution_stats = list(tqdm(map(self._execute_workflow, processing_args), total=len(processing_args)))
         else:
             if multiprocess:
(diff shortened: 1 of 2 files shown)


core/eolearn/core/eoexecution.py
```diff
@@ -201,6 +201,8 @@ def _execute_workflow(self, process_args):
             if return_results:
                 stats[self.RESULTS] = results
 
+        except KeyboardInterrupt:
+            raise KeyboardInterrupt()
```
AleksMat, line 205: Comments:

- When you raise an exception without a custom message you don't have to initialize it. So instead of `raise KeyboardInterrupt()` you can do `raise KeyboardInterrupt`. It is a Python thing...
- Perhaps we want to propagate an entire stack trace to inform users where exactly in the code execution was interrupted. That can be achieved like this:

```Python
        except KeyboardInterrupt as exception:
            raise KeyboardInterrupt from exception
```

AleksMat (commented):
Perhaps it would be good to have a unit test to test this case so that in the future if we change this part of the code we don't lose this functionality. I would suggest something like this:

Create a dummy task that just raises a `KeyboardInterrupt`. Then use EOExecutor to run e.g. 10 times an EOWorkflow with this single task. The expected behavior would be that `KeyboardInterrupt` is propagated and execution stops. This should happen in single-worker mode, multithreaded mode, and multiprocessing mode.

core/eolearn/core/eoexecution.py
DomagojKorais: Added tests to check for `KeyboardInterrupt` propagation

AleksMat (approved):

PR #276, opened 2021-03-05
Fix dependecy updates

AleksMat:
This should fix all remaining failing tests and pylint issues. Note that CI will fail because it need new versions of sentinelhub-py and s2cloudless.

### DIFF
--- a/io/eolearn/io/local_io.py
+++ b/io/eolearn/io/local_io.py
@@ -294,11 +294,11 @@ class ExportToTiff(BaseLocalIo):
 
         channel_count, height, width = image_array.shape
 
-        src_crs = eopatch.bbox.crs.epsg
+        src_crs = {'init': eopatch.bbox.crs.ogc_string()}
         src_transform = rasterio.transform.from_bounds(*eopatch.bbox, width=width, height=height)
 
         if self.crs:
-            dst_crs = self.crs.epsg
+            dst_crs = {'init': self.crs.ogc_string()}
             dst_transform, dst_width, dst_height = rasterio.warp.calculate_default_transform(
                 src_crs, dst_crs, width, height, *eopatch.bbox
             )
--- a/mask/eolearn/mask/cloud_mask.py
+++ b/mask/eolearn/mask/cloud_mask.py
@@ -18,7 +18,7 @@ import joblib
 import numpy as np
 import cv2
 from skimage.morphology import disk
-from s2cloudless import S2PixelCloudDetector, MODEL_EVALSCRIPT
+from s2cloudless import S2PixelCloudDetector, get_s2_evalscript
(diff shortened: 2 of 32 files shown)


io/eolearn/io/local_io.py
```diff
@@ -294,11 +294,11 @@ def execute(self, eopatch, *, filename=None):
 
         channel_count, height, width = image_array.shape
 
-        src_crs = eopatch.bbox.crs.epsg
+        src_crs = {'init': eopatch.bbox.crs.ogc_string()}
```
batic, line 297: don't use 'init' here, to get rid of (at least some) warnings

io/eolearn/io/local_io.py
```diff
@@ -294,11 +294,11 @@ def execute(self, eopatch, *, filename=None):
...
         src_transform = rasterio.transform.from_bounds(*eopatch.bbox, width=width, height=height)
 
         if self.crs:
-            dst_crs = self.crs.epsg
+            dst_crs = {'init': self.crs.ogc_string()}
```
batic, line 301: same here

mask/eolearn/mask/cloud_mask.py
```diff
@@ -53,7 +53,7 @@ class AddCloudMaskTask(EOTask):
     """
     def __init__(self, classifier, data_feature, cm_size_x=None, cm_size_y=None, cmask_feature='CLM',
                  cprobs_feature=None, instance_id=None, data_collection=None,
-                 image_format=MimeType.TIFF, model_evalscript=MODEL_EVALSCRIPT, data_source=None):
+                 image_format=MimeType.TIFF, model_evalscript=None, data_source=None):
```
batic, line 56: We should remove these parameters (`image_format`, `model_evalscript`, `data_source`, `instance_id`, `data_collection`)? Either already available data (in eopatch) or (official) evalscript from the s2cloudless to download the data should be used. 

_If not now, when s2cloudless is updated..._

batic (commented):
Just a few notes. Otherwise looks good.

io/eolearn/io/local_io.py
AleksMat: Without this it doesn't work. Using epsg code failed unit tests. I suggest that this is resolved in the next release.

io/eolearn/io/local_io.py
AleksMat: Same answer

mask/eolearn/mask/cloud_mask.py
AleksMat: Yes, in the next release we will significantly refactor this.

io/eolearn/io/local_io.py
batic: Huh. Something worked for me, but I am now completely lost. I agree - let's take time to understand the changes going to rasterio 1.2...

PR #277, opened 2021-03-06
Feat/notebooks update

batic:
Updated some notebooks and other minor things.


mask/eolearn/mask/cloud_mask.py
```diff
@@ -371,8 +85,8 @@ def __init__(self,
                  mask_feature='CLM_INTERSSIM',
                  mono_threshold=0.4,
                  multi_threshold=0.5,
-                 average_over=1,
-                 dilation_size=1):
+                 average_over=4,
```
AleksMat, line 88: I know that `get_s2_pixel_cloud_detector` function had these default but `s2cloudless.S2PixelCloudDetector` has defaults `average_over=1, dilation_size=1` ([link](https://github.com/sentinel-hub/sentinel2-cloud-detector/blob/master/s2cloudless/cloud_detector.py#L50)).

I really don't know what is better here and I don't remember what was the reason for choosing certain parameters.

mask/eolearn/tests/test_cloud_mask.py
```diff
@@ -143,50 +46,43 @@ def _check_shape(self, output, data):
...
+        self._check_shape(eop_clm.data['CLP_TEST'], eop_clm.data['ALL_DATA'])
+        self.assertTrue(eop_clm.mask['CLM_TEST'].dtype == bool)
+        self.assertTrue(eop_clm.data['CLP_TEST'].dtype == np.float32)
 
         # Compare mean cloud coverage with provided reference
-        mean_clm_provided = np.mean(eop_clm.mask['CLM'])
-        mean_clp_provided = np.mean(eop_clm.data['CLP'])
-        self.assertAlmostEqual(np.mean(eop_clm.mask['CLM_INTERSSIM']), mean_clm_provided, places=2)
-        self.assertAlmostEqual(np.mean(eop_clm.data['CLP_S2C']), mean_clp_provided, places=1)
+        self.assertAlmostEqual(np.mean(eop_clm.mask['CLM_TEST']), np.mean(eop_clm.mask['CLM_S2C']), places=1)
```
AleksMat, line 81: Perhaps we could compare entire arrays, not just their mean values?

AleksMat (commented):
A couple of comments regarding cloud masks otherwise looks good.

By the way, I removed all warnings from notebooks. There is no need to show paths in our environments. I made pushed commit about this to this PR.

mask/eolearn/mask/cloud_mask.py
```diff
@@ -910,3 +624,9 @@ def execute(self, eopatch):
             eopatch.data[multi_proba_feature] = (multi_proba * is_data).astype(np.float32)
 
         return eopatch
+
+
+class AddMultiCloudMaskTask(CloudMaskTask):
```
AleksMat, line 629: Pylint complains that a class docstring is missing -> [CI report](https://travis-ci.org/github/sentinel-hub/eo-learn/jobs/763471991#L1569)

mask/eolearn/tests/test_cloud_mask.py
batic: I could do that if I rewrote the existing mask in the test eopatch. Should I?

mask/eolearn/mask/cloud_mask.py
batic: Neither do I. 

I chose this values because that meant the least changes in the existing example notebooks. We have also often seen people using s2cloudless with other values (not these), and they then claimed s2cloudless to be worse than it is. 

As for the reasons: I'd say they were selected empirically.

mask/eolearn/mask/cloud_mask.py
batic: Didn't think about that one as it is deprecated. I've added something just to make pylint happy.

mask/eolearn/mask/cloud_mask.py
AleksMat: Ok, that makes sense. We can leave it as it is now.

mask/eolearn/tests/test_cloud_mask.py
AleksMat: Ah, now I understand - these masks are not even the same shape because we are testing a downscaled version. But they do have the same mean values. Then everything is ok.

PR #282, opened 2021-03-24
Fixed documentation

AleksMat:
Changes:

- Documentation build was failing because the new Sphinx version wasn't compatible anymore with my fork of `m2r` package. Fortunately there exists now a stable and maintained fork `m2r2`. So I replaced a dependency and it worked.
- After that build was still failing and the issue was that Sphinx was parsing all example notebooks, even the ones that we don't include in the documentation. But then the new global timelapse notebook was using some extra package dependencies and because of that, the build would fail. I fixed this in a way that now only notebooks that we actually use in the documentation are being parsed and the other ones are left alone.
- After that, I decided to go through the entire documentation and fix all the errors about which Sphinx was complaining. This was quite a tedious process but I managed to fix everything. This has actually never been done before for eo-learn. Now Sphinx build process shows only a few warnings that are not problematic and would be hard to get rid of.

### DIFF
--- a/core/eolearn/core/utilities.py
+++ b/core/eolearn/core/utilities.py
@@ -49,8 +49,8 @@ class FeatureParser:
 
     General guidelines:
 
-    - Almost all EOTasks have take as a parameter some information about features. The purpose of this class is
-      to unite and generalize parsing of such parameter over entire eo-learn package
+    - Almost every `EOTask` requires an initialization parameter to define which features should be used by the task.
+      The purpose of this class is to unite and generalize parsing of such parameters over the entire eo-learn package.
     - The idea for this class is that it should support more or less any logical way how to describe a collection
       of features.
     - Parameter `...` is used as a contextual clue. In the supported formats it is used to describe the most obvious
--- a/core/eolearn/core/utilities.py
+++ b/core/eolearn/core/utilities.py
@@ -48,43 +48,60 @@ class FeatureParser:
     can be obtained by iterating over an instance of the class. An `EOPatch` is given as a parameter of the generator.
 
     General guidelines:
-        - Almost all `EOTask`s have take as a parameter some information about features. The purpose of this class is
-        to unite and generalize parsing of such parameter over entire eo-learn package
-        - The idea for this class is that it should support more or less any logical way how to describe a collection
-        of features.
-        - Parameter `...` is used as a contextual clue. In the supported formats it is used to describe the most obvious
-        way how to specify certain parts of feature collection.
-        - Supports formats defined with lists, tuples, sets and dictionaries.
+
+    - Almost all EOTasks have take as a parameter some information about features. The purpose of this class is
+      to unite and generalize parsing of such parameter over entire eo-learn package
+    - The idea for this class is that it should support more or less any logical way how to describe a collection
+      of features.
+    - Parameter `...` is used as a contextual clue. In the supported formats it is used to describe the most obvious
+      way how to specify certain parts of feature collection.
+    - Supports formats defined with lists, tuples, sets and dictionaries.
 
     Supported input formats:
-        - `...` - Anything that exists in a given `EOPatch`
-        - A feature type describing all features of that type. E.g. `FeatureType.DATA` or `FeatureType.BBOX`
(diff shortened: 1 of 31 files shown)


core/eolearn/core/utilities.py
```diff
@@ -48,43 +48,60 @@ class FeatureParser:
...
     General guidelines:
-        - Almost all `EOTask`s have take as a parameter some information about features. The purpose of this class is
-        to unite and generalize parsing of such parameter over entire eo-learn package
-        - The idea for this class is that it should support more or less any logical way how to describe a collection
-        of features.
-        - Parameter `...` is used as a contextual clue. In the supported formats it is used to describe the most obvious
-        way how to specify certain parts of feature collection.
-        - Supports formats defined with lists, tuples, sets and dictionaries.
+
+    - Almost all EOTasks have take as a parameter some information about features. The purpose of this class is
```
batic, line 52: Either "Almost all EOTasks have" or "Almost all EOTasks take", but not "have take".

core/eolearn/core/utilities.py
```diff
@@ -48,43 +48,60 @@ class FeatureParser:
...
-                'L8-BANDS': 'INTERPOLATED_L8_BANDS',
-                'NDVI': ...
-            },
-        }
+
+    - Anything that exists in a given `EOPatch` is defined with `...`
+    - A feature type describing all features of that type. Example: `FeatureType.DATA` or `FeatureType.BBOX`
+    - A single feature as a tuple. Example: `(FeatureType.DATA, 'BANDS')`
+    - A single feature as a tuple. Example: `(FeatureType.DATA, 'BANDS')`
+    - A single feature as a tuple with new name. Example `(FeatureType.DATA, 'BANDS', 'NEW_BANDS')`
```
batic, line 66: 🤔 Is this even true? 
If `new_names=False` such notation will result in two features of type `FeatureType.DATA` type, one named `'BANDS'` and another named `'NEW_BANDS'`. 

`(FeatureType.DATA: {'BANDS': 'NEW_BANDS'})` is much more clear notation, and possibly removes the need for `new_names` parameter altogether. The problem arises as the `FeatureParser` is typically used within an eo-learn task, and the user doesn't know if the `new_names` is used or not.

batic (commented):
😍 

Thanks for this cleanup. It will make a lot of difference to us and our users!

The comment about `FeatureParser` is something that cannot (and should not) be dealt with in this PR (apart from changing/improving the docs), but would require at least a chat how to deal with it.

core/eolearn/core/utilities.py
AleksMat: I rephrased the entire sentance. Should be better now.

core/eolearn/core/utilities.py
AleksMat: Yes, the case where you have `(ft, fn1, fn2)` is the problematic case that can be interpreted in 2 different ways depending on parameter `new_names`. It should be avoided when trying to define multiple features.

We noticed this problem a few years back but never had the time to fix it. Probably a good solution would be to remove support for providing multiple features with the notation `(ft, fn1, fn2, fn3, ...)` so that this notation can be used with the single purpose of defining `(ft, fn, new_fn)`. I'll open a new internal issue for this.

Notation `(FeatureType.DATA: {'BANDS': 'NEW_BANDS'}}` is for sure much cleaner but in some cases a bit less convenient. Besides that, you probably need `new_names` parameter so that users could be informed if their new names will have an effect or not. The information if a features parameter accepts new names or not should probably be written in docstrings for each parameter. At the moment, this is in many cases a hidden/undocumented extra functionality for which you have to read the code to understand it.

batic: Completely agree about the new issue.

I would personally opt for `(FeatureType.DATA: {'BANDS': 'NEW_BANDS'}}` notation, which is familiar from how pandas rename works (`rename={'old_name': 'new_name'}). Although it seems less convenient, it would allow for "everything in one go":
`[(FeatureType.DATA: [{'old_name': 'new_name}, 'name_1', 'name_2', {'old_name3' : 'new_name3'}]`

batic:
👍 

I suggest you "squash and merge".

PR #284, opened 2021-03-25
Fixed plotting boolean masks and minor CI update

AleksMat:
This PR fixes 2 unrelated things:

- Plotting features with boolean values by casting values to uint8
- running pytest in Travis CI procedure - module with SH integration tests has been renamed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,8 +49,7 @@ script:
     fi;
   else
     pytest \
-      --ignore io/eolearn/tests/test_io.py \
-      --ignore io/eolearn/tests/test_processing_io.py ;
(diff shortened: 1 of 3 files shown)


batic (approved):
Correct me if I am wrong, but the tests are failing because sh-py, right? Is there is a simple way to run tests (when merging to develop) on sh-py's develop branch? 

Otherwise, approving the merge.

AleksMat:
Yes, tests are failing only because of sh-py changes. The only way I see we could run tests on develop branch of sh-py is to modify CI procedure in a way that it would clone sh-py and install it from develop branch. But I'm not sure we want to do that.

PR #291, opened 2021-04-14
Tasks to import vectors from local/s3 files, Geopedia and GeoDB.

batic:
EOTask to import vectors from:

  * local file
  * s3 file (depends on AWS credentials specified in sentinelhub-py config)
  * Geopedia table (depends on geopedia credentials specified in sentinelhub-py config) 
  * GeoDB collection (to run test, client id and secret should be in the environment)


batic:
Should I add `git+git://github.com/dcs4cop/xcube-geodb.git` to a list of requirements?

AleksMat:
> Should I add `git+git://github.com/dcs4cop/xcube-geodb.git` to a list of requirements?

Hm, I'm leaning towards having this as a package extension dependency. That is because the package does not have a full/stable support for pip, only for conda. As you can see [here](https://github.com/dcs4cop/xcube-geodb/blob/master/setup.py#L29) installing it with pip won't install package dependencies so we also have to add those.

In either case, please add it in a form of

```
xcube-geodb @ git+git://github.com/dcs4cop/xcube-geodb.git
```

Only this way the setup will install it, otherwise it will be ignored.

io/eolearn/tests/test_geometry_io.py
```diff
@@ -0,0 +1,98 @@
...
+        (True, True, 193,
+         BBox([657089.4234095092, 5071037.944679743, 661093.7825844937, 5075039.833417523], crs=CRS.UTM_31N)),
+        (True, True, 125,
+         BBox([657690.073230332, 5071637.340226217, 660493.1246978994, 5074440.391780451], crs=CRS.UTM_31N))
+    ])
+class TestVectorImportTask:
+
+    @pytest.mark.parametrize('path', [
+        os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../../example_data/import-gpkg-test.gpkg'),
+        's3://eolearn-io/import-gpkg-test.gpkg'
```
AleksMat, line 36: I don't have a permission for this bucket, neither does CI. Perhaps we can instead mock this with `moto`? This way we could be used our locally saved `import-gpkg-test.gpkg` in the background.

io/eolearn/tests/test_geometry_io.py
```diff
@@ -0,0 +1,98 @@
...
+        feature=feature,
+        geodb_client=geodb_client,
+        geodb_collection='france_metropolitan_gsaa_2015',
+        geodb_db='lpis_iacs',
+        reproject=reproject,
+        clip=clip
+    )
+    eop = import_task.execute(bbox=bbox)
+    assert len(eop[feature]) == n_features, 'Wrong number of features!'
+    assert eop[feature].crs.to_epsg() == eop.bbox.crs.epsg
```
AleksMat, line 98: So far this can only run on your computer. How can we make it work anywhere else, e.g. on Travis CI?

- Package dependencies for `xcube-geodb` should be installed.
- OAuth credentials should be set as env variables in Travis
- Are we running our own GeoDB server or using an existing one, e.g. from EDC?

AleksMat (changes requested):

AleksMat:
One more thing: Please add documentation page for the new module.

io/eolearn/io/geometry_io.py
```diff
@@ -0,0 +1,261 @@
...
+
+This source code is licensed under the MIT license found in the LICENSE
+file in the root directory of this source tree.
+"""
+
+import logging
+
+import boto3
+import fiona
+import geopandas as gpd
```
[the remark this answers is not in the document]
batic: 👍

io/eolearn/tests/test_geometry_io.py
batic: My tries were unsuccessful. I didn't want to spend a lot of time on this, so I rather made it so that the test is skipped if the user cannot access the bucket.

io/eolearn/tests/test_geometry_io.py
batic: `pytest.importorskip` to the rescue

io/eolearn/io/geometry_io.py
```diff
@@ -13,16 +13,76 @@
 import boto3
 import fiona
 import geopandas as gpd
-from eolearn.core import EOPatch, EOTask, FeatureType
+from eolearn.core import EOPatch, EOTask, FeatureType, FeatureTypeSet
```
AleksMat, line 16: Pylint complains that `FeatureType` is now an unused import.

io/eolearn/io/geometry_io.py
```diff
@@ -31,26 +91,19 @@ def __init__(self, feature, path, reproject=True, clip=False, config=None, **kwa
...
-        :param reproject: Should the geometries be transformed to EOPatch coordinate reference system?
+        :param reproject: Should the geometries be transformed to coordinate reference system of the requested bbox?
         :type reproject: bool, default = True
-        :param clip: Should the geometries be clipped to EOPatch bbox, or should be geometries kept as they are?
+        :param clip: Should the geometries be clipped to the requested bbox, or should be geometries kept as they are?
         :type clip: bool, default = False
-        :param config: A configuration object containing AWS credentials
+        :param config: A configuration object with AWS credentials (if not provided, ~/.aws/credentials will be used)
         :type config: SHConfig
-        :param kwargs: Additional args that will be passed to `fiona.open` call (e.g. layer='GPKG_LAYER')
```
AleksMat: This line that documents `kwargs` has been removed. I would put it back.

io/eolearn/io/geometry_io.py
```diff
@@ -90,143 +143,101 @@ def dataset_crs(self):
...
-    def _prepare_bbox(self, bbox):
-        """ This makes sure that the bounding box is in correct CRS. The final bounds can be larger than given bbox
-        """
-        if bbox is None:
-            self._dataset_crs = CRS.WGS84  # geopedia always returns data in WGS84
-            return None
-
-        self._dataset_crs = bbox.crs
-        return bbox.transform_bounds(CRS.POP_WEB)
+        self.dataset_crs = CRS.WGS84
```
AleksMat, line 185: `self.dataset_crs` is actually not needed in this task at all. You already read CRS from responses.

io/eolearn/io/geometry_io.py
```diff
@@ -248,14 +258,13 @@ def dataset_crs(self):
     def _load_vector_data(self, bbox):
         """ Loads vector data from geoDB table
         """
+        prepared_bbox = bbox.transform_bounds(self.dataset_crs) if bbox else None
 
         return self.geodb_client.get_collection_by_bbox(
             collection=self.geodb_collection,
             database=self.geodb_db,
-            bbox=bbox.geometry.bounds,
+            bbox=prepared_bbox.geometry.bounds,
```
AleksMat, line 266: This will raise an error if `prepared_bbox` is `None`. I suggest changing:

```
bbox_bounds = bbox.transform_bounds(self.dataset_crs).geometry.bounds if bbox else None
```
a few lines above and passing `bbox_bounds` here.

io/eolearn/tests/test_geometry_io.py
AleksMat: I'll check if I can find a solution for this. Because if we have tests for which only you have credentials to run then we'll forget about that and the tests will be usually skipped.

io/eolearn/io/geometry_io.py
batic: True, fixed. 

PyCharm doesn't mark it as unused, as it is used to describe feature parameter type in the docs.

io/eolearn/io/geometry_io.py
batic: I would not. 

The base class doesn't need kwargs, and I'd rather that each new implementation of the vector import tasks take care of additional arguments. If - at some point - base task will need kwargs, we can add them.

io/eolearn/tests/test_geometry_io.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        eop = import_task.execute(bbox=bbox)
+        assert len(eop[feature]) == n_features, 'Wrong number of features!'
+        to_crs = crs or import_task.dataset_crs
+        assert eop[feature].crs == to_crs.pyproj_crs()
+
+
+def test_clipping_wrong_crs():
+    with pytest.raises(ValueError):
+        path = os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../../example_data/import-gpkg-test.gpkg')
```
AleksMat, line 67: You have `gpkg_file` fixture for this.

io/eolearn/io/geometry_io.py
batic: True.

io/eolearn/io/geometry_io.py
batic: Good catch.

io/eolearn/tests/test_geometry_io.py
batic: Missed, thanks.

io/eolearn/io/geometry_io.py
AleksMat: Sorry, I think I marked the wrong class - I meant to say that `VectorImportTask` is now missing documentation of `kwargs`, not `_BaseVectorImportTask`.

batic: 👍 Added.

io/eolearn/tests/test_geometry_io.py
AleksMat: I investigated this and it turns out:

- `fiona` uses `boto3` only to parse session information and not for interaction with S3 buckets. Instead it uses GDAL implementation of S3 client. Because of that `moto` cannot mock it.

- I checked how AWS tests are implemented in `fiona` ([here](https://github.com/Toblerity/Fiona/blob/master/tests/test_vfs.py#L185)). It turns out they also have a real bucket and credentials and they skip tests if credentials are missing.

Hence it is safe to say there is no easy way how to mock this and we can keep tests as they are now.

PR #297, opened 2021-04-28
Added sampling_fraction parameter

AleksMat:
This is a quick improvement of the existing `PointSamplingTask` that allows defining a fraction of data to be sampled.

For more I would prefer to refactor sampling functionalities into a new more general task.


zigaLuksic (approved):
Does the job (but i agree that we need a better task for the long run).

PR #303, opened 2021-05-21
Multiple minor improvements

AleksMat:
Every commit is a different minor fix or improvement, a couple of them reference the open GitHub issues.

### DIFF
--- a/core/eolearn/core/core_tasks.py
+++ b/core/eolearn/core/core_tasks.py
@@ -75,7 +75,7 @@ class IOTask(EOTask):
         """ A filesystem property that is being lazy-loaded the first time it is needed
         """
         if self._filesystem is None:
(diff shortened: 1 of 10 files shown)


core/eolearn/core/core_tasks.py
```diff
@@ -75,7 +75,7 @@ def filesystem(self):
         """ A filesystem property that is being lazy-loaded the first time it is needed
```
zigaLuksic, line 75: It would be nice to also update the description that it is no longer lazy-loaded

zigaLuksic (approved):

core/eolearn/core/core_tasks.py
AleksMat: I completely agree and fixed it.

PR #307, opened 2021-06-08
Implementation of Meteoblue IO tasks

AleksMat:
Implements tasks `MeteoblueVectorTask` and `MeteoblueRasterTask` that download data from [Meteoblue services](https://www.meteoblue.com/en/weather-api/dataset-api/index) and add it into an EOPatch either as a vector or a raster feature.


patrick-zippenfenig (approved):
Code and tests look good!

Issue #301, opened 2021-05-10
[FEAT] VectorToRaster time-dependent support


asylve:
**What is the problem? Please describe.**

I needed to convert time-dependent geopandas vector data to raster images for an image masking project. As noted in the docs for eolearn.geometry.transformations.VectorToRaster, rasterization of time-depended features is not yet supported. 

**Here's the solution**

In my project I solved this by looping over each time step in the vector data, using VectorToRaster to store the result in MASK_TIMELESS['TEMP'], retrieving this and storing in a numpy array, then finally adding the numpy array to MASK['TIMED_MASK']. See below:

```
times, length, width, channels = eopatch.data['BANDS'].shape #get the required shape of the masks
timed_mask = np.zeros(shape=(times, length, width, 1), dtype=np.int16) #this will hold the rasterized masks

for vector_mask in vector_masks:
    add_raster = VectorToRaster(vector_mask, (FeatureType.MASK_TIMELESS, 'TEMP'),
                                    values_column='CT', raster_shape=(FeatureType.DATA, 'BANDS'),
                                    raster_dtype=np.int16)
    add_raster(eopatch)
    timed_mask[count] = eopatch.mask_timeless['TEMP'] #append the resulting raster image to our numpy array
            
eopatch.mask['TIMED_MASK'] = timed_mask #store the numpy array in the eopatch mask feature
```

Would it make sense to extend this functionality into VectorToRaster, so that a geopandas dataframe with a 'time' column could be passed in, and the resulting set of raster features would be passed into the eopatch? I would be happy to give this a try if it makes sense.

AleksMat:
Hi @asylve,

I agree with the proposed extension of `VectorToRaster` task. We'll be happy to accept any pull request regarding that.

Each time-dependent vector feature already has to have a column named `'TIMESTAMP'`, therefore the task should group entries according to this column and rasterize them into a time-dependent raster feature. If an EOPatch already has a list of timestamps then the task should rasterize according to them, otherwise it should probably add them.

P.S. Sorry for the late answer.

asylve:
Thanks, I'm glad you agree with the concept. I will start working on it.

PR #312, opened 2021-07-14
Feat/vector to raster timed

asylve:
As per issue #301 

Added functionality to the VectorToRaster function so that it can take in a time dependent eopatch feature or geopandas dataframe with a 'TIMESTAMP' column. 

If there are no timestamps in the eopatch, they are added from the vector.
If there are timestamps in the eopatch, then timestamps in the vector that are not in the eopatch are ignored. 
Any timestamps in the eopatch that are not represented in the vector are rasterized into the no_data_value.

### DIFF
--- a/geometry/eolearn/geometry/transformations.py
+++ b/geometry/eolearn/geometry/transformations.py
@@ -13,6 +13,8 @@ file in the root directory of this source tree.
 import logging
 import warnings
 
+import itertools
+
 import numpy as np
 import pandas as pd
 import pyproj
@@ -113,9 +115,12 @@ class VectorToRaster(EOTask):
 
         if not VectorToRaster._is_geopandas_object(vector_input):
             vector_input = VectorToRaster._parse_features(vector_input,
-                                                          allowed_feature_types={FeatureType.VECTOR_TIMELESS, FeatureType.VECTOR})
+                                                          allowed_feature_types=
+                                                          {FeatureType.VECTOR_TIMELESS, FeatureType.VECTOR})
         raster_feature = next(iter(
-            VectorToRaster._parse_features(raster_feature, allowed_feature_types=FeatureTypeSet.RASTER_TYPES_3D.union(FeatureTypeSet.RASTER_TYPES_4D))
+            VectorToRaster._parse_features(raster_feature,
+                                           allowed_feature_types=
+                                           FeatureTypeSet.RASTER_TYPES_3D.union(FeatureTypeSet.RASTER_TYPES_4D))
         ))
 
         return vector_input, raster_feature
@@ -132,9 +137,9 @@ class VectorToRaster(EOTask):
         """
         if VectorToRaster._is_geopandas_object(vector_input):
             return 'TIMESTAMP' not in vector_input
-        else:
-            vector_type = next(vector_input._get_features())[0]
-            return vector_type.is_timeless()
+
+        vector_type = next(iter(vector_input))[0]
+        return vector_type.is_timeless()
 
     def _get_vector_data(self, eopatch):
         """ Provides vector data which will be rasterized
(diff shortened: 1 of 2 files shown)


geometry/eolearn/geometry/transformations.py
```diff
@@ -257,26 +294,49 @@ def execute(self, eopatch):
...
+            
+            #loop over timestamps
+            for idx, time, flag in timestamp_flags:
+                if flag==1:
+                    rasterization_shapes = self._get_rasterization_shapes(eopatch, group_classes=group_classes, timestamp=time)
 
-        affine_transform = rasterio.transform.from_bounds(*eopatch.bbox, width=width, height=height)
-        rasterize_args = dict(self.rasterio_params, transform=affine_transform, dtype=self.raster_dtype)
+                    if not rasterization_shapes:
+                        break
```
AleksMat, line 332: This will break the outer for loop. Instead it should be `continue`

geometry/eolearn/geometry/transformations.py
```diff
@@ -134,15 +145,41 @@ def _get_vector_data(self, eopatch):
...
+        if len(eopatch.timestamp )==0:
+            eopatch.timestamp = list(vector_timestamps)
+            warnings.warn('The EOPatch has no timestamp data. Timestamps from the vector data will be added '
+                          'to the EOPatch', RuntimeWarning)
+
+        if any([t_vec not in eopatch.timestamp for t_vec in vector_timestamps]):
+            warnings.warn('Some timestamps in the vector data are not found in the eopatch timestamps and will be '
+                          'ignored', RuntimeWarning)
+
+        timestamp_flags = [(idx, time, 1) if time in vector_timestamps else (idx, time, 0) for idx, time in enumerate(eopatch.timestamp)]
```
AleksMat, line 165: I would like to avoid undescriptive flags that have an effect in different part of the code. Instead, you can simply return a set of timestamps for which some data exists. Then in the execute method you can do something like:

```python
for timestamp in eopatch.timestamp:
    if timestamp not in timestamps_in_vector_data
        continue:
    ...
```

geometry/eolearn/geometry/transformations.py
```diff
@@ -257,26 +294,49 @@ def execute(self, eopatch):
...
 
-        affine_transform = rasterio.transform.from_bounds(*eopatch.bbox, width=width, height=height)
-        rasterize_args = dict(self.rasterio_params, transform=affine_transform, dtype=self.raster_dtype)
+                    if not rasterization_shapes:
+                        break
 
-        raster = self._get_raster(eopatch, height, width)
-        rasterize_func = rasterio.features.rasterize if self.overlap_value is None else self.rasterize_overlapped
-        rasterize_func(rasterization_shapes, out=raster, **rasterize_args)
+                    raster = self._get_raster(eopatch, height, width)
```
AleksMat, line 334: `_get_raster` method should be called only once and should return what is currently called `timed_raster`. Only this way it will be possible to rasterize more things into an existing raster feature. At this point, you can just slice the full temporal raster.

geometry/eolearn/geometry/transformations.py
```diff
@@ -134,15 +145,41 @@ def _get_vector_data(self, eopatch):
...
         :param group_classes: If true, function returns a zip that iterates through cascaded unions of polygons of the
         same class, otherwise zip iterates through all polygons regardless of their class.
         :type group_classes: boolean
+        :param timestamp: If supplied, only shapes associated with the given timestamp will be returned.
+        :type timestamp: datetime.datetime
         """
         vector_data = self._get_vector_data(eopatch)
+        
+        if timestamp:
+            vector_data = vector_data[vector_data['TIMESTAMP']==timestamp]
```
AleksMat, line 182: This can be quite inefficient if there are a lot of timestamps. This would require a bit more refactoring but I suggest something like this:

1. remove timestamps that are not in `eopatch.timestamp` (e.g. with `isin` method),
2. use `_get_rasterization_shapes` on the remaining dataframe (without splitting it by timestamp yet),
3. before rasterization split the final dataframe by timestamps. I think `pandas` `groupby` operation should be more efficient in splitting by timestamp than the current implementation.

AleksMat (changes requested):
Hi @asylve,

Thank you for preparing this PR. I like the new functionality but some code improvements and code style fixes will have to be done before we merge. I wrote the main comments in this review. 

I also recommend installing `pylint` package and running

```
pylint geometry/eolearn/geometry/transformations.py
```

from the main project folder. It will show you a list of code style issues that should also be fixed.

geometry/eolearn/geometry/transformations.py
```diff
@@ -134,15 +145,41 @@ def _get_vector_data(self, eopatch):
...
         return eopatch[feature_type][feature_name]
 
-    def _get_rasterization_shapes(self, eopatch, group_classes=False):
+    def _sort_timestamps(self, eopatch):
+        """ indexes all timestamps found in the vector input data according to the eopatch timestamps. Any 
+        timestamps present in the vector data but missing from the EOPatch are ignored. The exception to this
+        is if no timestamps are found in the EOPatch, they are added from the vector data. 
+        """
+        vectors = self._get_vector_data(eopatch)
+        vector_timestamps = set(vectors['TIMESTAMP'])
```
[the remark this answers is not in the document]
asylve: OK, I have added your suggestion instead

geometry/eolearn/geometry/transformations.py
```diff
@@ -134,15 +145,41 @@ def _get_vector_data(self, eopatch):
...
+
+        if len(eopatch.timestamp )==0:
```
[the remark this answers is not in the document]
asylve: Done

geometry/eolearn/geometry/transformations.py
```diff
@@ -134,15 +145,41 @@ def _get_vector_data(self, eopatch):
...
+            eopatch.timestamp = list(vector_timestamps)
+            warnings.warn('The EOPatch has no timestamp data. Timestamps from the vector data will be added '
+                          'to the EOPatch', RuntimeWarning)
+
+        if any([t_vec not in eopatch.timestamp for t_vec in vector_timestamps]):
+            warnings.warn('Some timestamps in the vector data are not found in the eopatch timestamps and will be '
```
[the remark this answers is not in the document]
asylve: OK I removed the warning

geometry/eolearn/geometry/transformations.py
asylve: Done

geometry/eolearn/geometry/transformations.py
asylve: Done

geometry/eolearn/geometry/transformations.py
asylve: I updated `_get_raster` so it now is only called once and returns the either the existing raster feature, an array with no time dimension, or an array with a time dimension depending on the case.

geometry/eolearn/geometry/transformations.py
asylve: I followed step 1 and 2. For step 3 I used `itertools` `groupby` instead of `pandas` since the `_get_rasterization_shapes` returns a zip object.

asylve:
Thanks @AleksMat. I updated based on your comments. When I run pylint it now gives a score of 10/10. Please let me know what you think.

AleksMat:
Hi @asylve,

I'm finally available again. I checked your changes and the code looks much better. I will now merge it but we plan to still make a few minor code updates in this part of the code in the near future. Thanks for the contribution. :+1: 

I also saw your [blog post about sea ice density detection](https://asylve.github.io/2021/04/28/sea-ice/). Really nice work. :clap:

AleksMat (approved):

PR #316, opened 2021-08-11
A few fixes and improvements

AleksMat:
This PR makes a few unrelated changes:

- solves the problem with handling `ffmpeg` package in one of the example notebooks,
  * some updates in dockerfiles according to this issue,
- some code refactoring in `eolearn.geometry.transformations` module,
- a minor documentation fix,
- a minor fix in `fiona` version in requirements,
- a few minor pylint fixes.

Sorry for putting everything into a single PR.

### DIFF
--- a/geometry/eolearn/geometry/transformations.py
+++ b/geometry/eolearn/geometry/transformations.py
@@ -11,7 +11,6 @@ file in the root directory of this source tree.
 """
 
 import functools
-import itertools
 import logging
 import warnings
 
--- a/features/eolearn/features/interpolation.py
+++ b/features/eolearn/features/interpolation.py
@@ -587,16 +587,16 @@ class ResamplingTask(InterpolationTask):
         new_data[time_mask] = interp_func(resampled_times[time_mask])
         return new_data
 
-    def get_interpolation_function(self, times, data):
+    def get_interpolation_function(self, times, series):
         """ Initializes interpolation model
 
         :param times: Array of reference times in second relative to the first timestamp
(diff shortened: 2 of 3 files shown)


geometry/eolearn/geometry/transformations.py
```diff
@@ -141,73 +143,69 @@ def _vector_is_timeless(vector_input):
...
 
-        vector_data = self._get_vector_data(eopatch)
+        if self._rasterize_per_timestamp:
+            vector_data = vector_data[vector_data.TIMESTAMP.isin(timestamps)]
 
-        if not timeless:
-            #remove timestamps that are not in the eopatch
-            vector_data = vector_data[vector_data['TIMESTAMP'].isin(eopatch.timestamp)]
+        if self.values_column is not None and self.values is not None:
+            values = [self.values] if isinstance(self.values, (int, float)) else self.values
```
zigaLuksic, line 188: Why is this needed?

zigaLuksic (commented):
I am only slightly concerned about one thing, the rest seems ok :)

geometry/eolearn/geometry/transformations.py
AleksMat: This `self.values` can either be a single value (i.e. int/float) or a list of values. Perhaps this can already be parsed in init?

zigaLuksic: Ah, my bad, i thought it was `values_column` twice, so it was odd that we were checking if its an integer or float.

PR #317, opened 2021-08-12
Update tests to use pytest

zigaLuksic:
I updated all the remaining tests that were using `unittest` to use `pytest`. The tests for `eolearn.core` are excluded from this pull request, since they have been updated on `develop-v1.0`.

I also fixed two outdated tests (an import error in `meteoblue` and a non-updated tests for the new `sh-py` collections). There are also some minor changes to other files (if i noticed a mistake in the docstrings).

The test content is almost entirely the same.

### DIFF
--- a/coregistration/eolearn/tests/test_coregistration.py
+++ b/coregistration/eolearn/tests/test_coregistration.py
@@ -9,8 +9,8 @@ file in the root directory of this source tree.
 """
 
 import logging
-import unittest
 
+import pytest
 import numpy as np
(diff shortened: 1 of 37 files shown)


AleksMat (commented):
Great work! I only found a few minor things, some of them have already been in the code before your changes. I found them mostly by running pylint on test files and ignoring errors about missing docstrings, invalid variable names or code duplications.

AleksMat (approved):

Issue #180, opened 2020-01-28
The eodata_io does not work with lazy loading.


bsircelj:
The method `save_eopatch` in `eodata_io.py` removes whole patch regardless if it was lazy loaded on line 32: `filesystem.removetree(patch_location)` thus producing an error when the patch is later tried to be saved:
`fs.errors.ResourceNotFound: resource 'During execution of task SaveTask: resource '/eopatch/data/NDVI.npy' not found' not found`

mlubej:
@AleksMat should we just make a try/except here with a `ResourceNotFound` exception? This would mean that the lazy-loaded feature would get skipped and would not be present on the disk.

I believe that if you lazy load and overwrite patch, it should not matter that you delete the features which were lazy-loaded. If you really care about the existing features, then you should use the overwrite features permission. 

On the other hand, if you care about such features, one could make sure that the eopatch is fully loaded into memory before the `removetree` is executed, but that brings potential overhead.

zigaLuksic:
This has been fixed in version `0.10`

PR #318, opened 2021-08-16
Fix eopatch saving w.r.t. compression and lazy loading

zigaLuksic:
Saving eopatch features at different compression levels produces an issue with old files not being properly overwritten (due to the added `.gz` file suffix). This leads to duplication of files as well as possible bugs when loading features.

This is now fixed with an additional step, where such files are found and removed (after the new ones are saved).

### DIFF
--- a/core/eolearn/core/eodata_io.py
+++ b/core/eolearn/core/eodata_io.py
@@ -29,11 +29,6 @@ def save_eopatch(eopatch, filesystem, patch_location, features=..., overwrite_pe
     """
     patch_exists = filesystem.exists(patch_location)
 
-    if overwrite_permission is OverwritePermission.OVERWRITE_PATCH and patch_exists:
-        filesystem.removetree(patch_location)
-        if patch_location != '/':  # avoid redundant filesystem.makedirs if the location is '/'
-            patch_exists = False
-
     if not patch_exists:
         filesystem.makedirs(patch_location, recreate=True)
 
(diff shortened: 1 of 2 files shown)


zigaLuksic:
Switched to WIP mark, because I'll try to resolve https://github.com/sentinel-hub/eo-learn/issues/180 as well.

core/eolearn/core/eodata_io.py
```diff
@@ -29,41 +29,84 @@ def save_eopatch(eopatch, filesystem, patch_location, features=..., overwrite_pe
...
+
+
+def remove_redundant_files(filesystem, eopatch_features, filesystem_features, current_compress_level):
+    """ Removes files that should have been overwriten but were not due to different compression levels
+    """
+    files_to_remove = []
+    saved_features = {(ftype, fname) for ftype, fname, _ in eopatch_features}
+    for ftype, fname, path in filesystem_features:
+        if fname is ... and not ftype.is_meta():
+            continue
```
AleksMat, line 82: I'm  quite sure this case cannot happen because of the last if-sentence in `walk_filesystem` function.

AleksMat (commented):
Looks good, I only have 2 minor comments.

core/eolearn/core/eodata_io.py
```diff
@@ -29,41 +29,84 @@ def save_eopatch(eopatch, filesystem, patch_location, features=..., overwrite_pe
...
+
+        different_compression = path.endswith(FileFormat.GZIP.extension()) != (current_compress_level > 0)
+        if (ftype, fname) in saved_features and different_compression:
+            files_to_remove.append(path)
+
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        # The following is intentionally wrapped in a list in order to get back potential exceptions
+        list(executor.map(filesystem.remove, files_to_remove))
```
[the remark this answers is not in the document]
zigaLuksic: Oh no, i forgot to throw this one out :open_mouth:

core/eolearn/core/eodata_io.py
zigaLuksic: will check and update accordingly

AleksMat (approved):

PR #321, opened 2021-08-31
EOWorkflow makes shallow copies of EOPatches

AleksMat:
Before each task in an EOWorkflow is executed the workflow will make sure that all input EOPatches are shallow-copied. This ensures the correct logic in a complicated workflow. This PR also adds `EOPatch.copy` method.

Tests for checking the copy functionality of EOWorkflow will be added in another PR into develop-v1.0.

### DIFF
--- a/core/eolearn/tests/test_eodata.py
+++ b/core/eolearn/tests/test_eodata.py
@@ -172,6 +172,41 @@ class TestEOPatch(unittest.TestCase):
 
         self.assertTrue(np.array_equal(eop_bands, bands), msg="Data numpy array not returned properly")
 
+    def test_shallow_copy(self):
+        path = os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../../example_data/TestEOPatch')
+        eop = EOPatch.load(path)
+
+        eop_copy = eop.copy()
+        assert eop == eop_copy
+        assert eop is not eop_copy
+
+        eop_copy.mask['CLM'] += 1
+        assert eop == eop_copy
+
+        eop_copy.timestamp.pop()
+        assert eop != eop_copy
+
+    def test_deep_copy(self):
+        path = os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../../example_data/TestEOPatch')
+        eop = EOPatch.load(path)
+
+        eop_copy = eop.copy(deep=True)
+        assert eop == eop_copy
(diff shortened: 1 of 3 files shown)


core/eolearn/tests/test_eodata.py
```diff
@@ -172,6 +172,41 @@ def test_get_feature(self):
 
         self.assertTrue(np.array_equal(eop_bands, bands), msg="Data numpy array not returned properly")
 
+    def test_shallow_copy(self):
+        path = os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../../example_data/TestEOPatch')
```
zigaLuksic, line 176: this should be defined as a constant at the start of file. Since these are the outdated 0.9 core tests i guess there is no fixture for it yet?

core/eolearn/tests/test_eodata.py
```diff
@@ -172,6 +172,41 @@ def test_get_feature(self):
...
+        eop = EOPatch.load(path)
+
+        eop_copy = eop.copy()
+        assert eop == eop_copy
+        assert eop is not eop_copy
+
+        eop_copy.mask['CLM'] += 1
+        assert eop == eop_copy
```
zigaLuksic, line 184: Maybe you can also compare id's of objects to make sure it's shallow?

AleksMat: The line `assert eop is not eop_copy` is meant to compare ids. Do you think it would be nicer to do `id(eop) != id(eop_copy)`?

core/eolearn/tests/test_eodata.py
AleksMat: Yes, I intentionally wrote like this for now and then improved it for the other PR for `develop-v1.0`. I hope it is ok if we leave this copy-pasta here as it has already been fixed in the new version.

core/eolearn/tests/test_eodata.py
zigaLuksic: I would like the comparison on feature level, e.g. `eop.mask['CLM'] is eop_copy.mask['CLM']`

AleksMat: Makes sense. Added.

(After this is merged, I'll merge `develop` into `develop-v1.0` so that this change will also be applied there.)

PR #322, opened 2021-08-31
EOWorkflow makes shallow copies of EOPatches for v1.0

AleksMat:
This is a continuation of https://github.com/sentinel-hub/eo-learn/pull/321 with additional tests and minor updates. This will be merged into develop-v1.0.


core/eolearn/tests/test_eodata.py
```diff
@@ -167,6 +167,41 @@ def test_get_feature():
     assert np.array_equal(eop_bands, bands), 'Data numpy array not returned properly'
 
 
+def test_shallow_copy(test_eopatch_path):
```
zigaLuksic, line 170: If we don't have a fixture for `test_eopatch` (that also loads it) yet, it should be added

AleksMat: You are right, I realized we already have this fixture. Fixed.

PR #323, opened 2021-09-01
Fixed merging of time-dependent features

AleksMat:
The process is now a bit refactored and more strict. I also added more unit tests.

Additionally, I also added `_FeatureDict.__eq__` method.


zigaLuksic (approved):
It looks fine

PR #327, opened 2021-09-08
Update example notebooks that appear in docs

zigaLuksic:
Minimal updates just to get the notebooks to work without any kind of deprecation errors.

I marked it as a draft, because if we add new sampling methods, the LULC pipeline notebook needs to be updated again.


AleksMat:
Looks good so far.

PR #330, opened 2021-09-13
Switched from Travis to GitHub actions and updated tests

AleksMat:
This is how the new testing setup looks like: https://github.com/sentinel-hub/eo-learn/actions/runs/1230856183


zigaLuksic (approved):
Looks great, lets hope this solution lasts for a while :D

PR #333, opened 2021-09-19
CI fixes

AleksMat:
Fixes a few issues in CI:

- Switches scheduled (i.e. cron job) testing from master to develop branch. This way we don't have to push to master whenever any minor tests start failing.
- Sets pylint to a fixed older version because clearly `2.11` has some issues.
- Updates Makefile pylint command to correctly propagate a status code.
- Suppresses info about setting SH credentials during Gitlab CI runs.
- Updates a failing integration test.

### DIFF
--- a/.github/workflows/ci_action.yml
+++ b/.github/workflows/ci_action.yml
@@ -15,7 +15,16 @@ jobs:
           - '3.8'
           - '3.9'
     steps:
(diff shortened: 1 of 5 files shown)


zigaLuksic (approved):
I was wondering what was going on with the pylint checks, it really messes up the imports. Thanks for fixing

PR #347, opened 2022-01-04
New feature serialization

AleksMat:
In this PR we get rid of saving EOPatch features into pickle format. It also updates some tests and resolves some warnings in eo-learn-features.

The only issue with the new format is that saving vector features into Geopackage format is sometimes a bit lossy - either a dtype of a certain pandas column gets changed or CRS definition gets changed. I'm trying to partially solve this at load but it is not 100% restorable. In the future we might have to handle possible additional cases regarding that.


core/eolearn/core/eodata_io.py
```diff
@@ -221,18 +229,30 @@ def _to_lowercase(ftype, fname, *_):
...
+    if feature_type.is_vector():
+        return FileFormat.GPKG
+    return FileFormat.JSON
+
+
 class FeatureIO:
     """ A class that handles the saving and loading process of a single feature at a given location
     """
-    def __init__(self, filesystem, path):
+    def __init__(self, filesystem: fs.base.FS, path: str, feature_type: FeatureType):
```
batic, line 247: As above: I'd prefer signature `feature_type: FeatureType, filesystem: fs.base.FS, path: str`. Even better would be that `FeatureIO` doesn't have to know its feature type, but that could only be achieved if `FeatureIO` would be a part fo feature itself. Unfortunately, that cannot be achieved without a substantial refactoring (and code breaking changes).

core/eolearn/core/eodata_io.py
```diff
@@ -281,20 +303,69 @@ def _save(self, filesystem, data, path, file_format, compress_level=0):
...
         """ Writes to a file
         """
         if file_format is FileFormat.NPY:
-            np.save(file, data)
-        elif file_format is FileFormat.PICKLE:
-            pickle.dump(data, file)
+            return np.save(file, data)
+
+        if file_format is FileFormat.GPKG:
+            return data.to_file(file, driver="GPKG", encoding="utf-8", layer=fs.path.basename(self.path))
```
batic, line 313: Do we want to allow the user to store several layers into the gpkg? 

Additionally, although `geopandas` doesn't require `driver="GPKG"` anymore, I'd keep it: explicit > implicit.

core/eolearn/core/eodata_io.py
```diff
@@ -281,20 +303,69 @@ def _save(self, filesystem, data, path, file_format, compress_level=0):
...
-    def _decode(file, path):
+            if self.feature_type is FeatureType.TIMESTAMP:
+                data = [timestamp.isoformat() for timestamp in data]
+
+            try:
+                json_data = json.dumps(data, indent=2)
+            except TypeError as exception:
+                raise TypeError(
+                    f'Failed to serialize {self.feature_type} into a JSON file. Make sure that this feature type '
+                    f'contains only basic Python types before attempting to serialize it.'
```
batic, line 327: I'd write "JSON serializable Python types" instead of "basic Python types".

core/eolearn/core/eodata_io.py
```diff
@@ -281,20 +303,69 @@ def _save(self, filesystem, data, path, file_format, compress_level=0):
...
         """ Loads from a file and decodes content
         """
-        if FileFormat.PICKLE.extension() in path:
+        if path.endswith(FileFormat.NPY.extension()):
+            return np.load(file)
+
+        if path.endswith(FileFormat.GPKG.extension()):
+            dataframe = gpd.read_file(file)
+
+            if dataframe.crs is not None:
```
batic, line 343: Why this wrapping into `CRS` and back?

core/eolearn/core/eodata_io.py
```diff
@@ -281,20 +303,69 @@ def _save(self, filesystem, data, path, file_format, compress_level=0):
...
+                with warnings.catch_warnings():
+                    warnings.simplefilter("ignore", category=SHUserWarning)
+                    dataframe.crs = CRS(dataframe.crs).pyproj_crs()
+
+            if 'TIMESTAMP' in dataframe:
```
batic, line 348: Yes, this is the drawback of the gpkg (compared to pickle). :(

core/eolearn/core/eodata_io.py
```diff
@@ -281,20 +303,69 @@ def _save(self, filesystem, data, path, file_format, compress_level=0):
...
+                dataframe.TIMESTAMP = pd.to_datetime(dataframe.TIMESTAMP)
+
+            return dataframe
+
+        if path.endswith(FileFormat.JSON.extension()) or path.endswith(FileFormat.GEOJSON.extension()):
+            json_data = json.load(file)
+
+            if self.feature_type is FeatureType.BBOX:
+                return Geometry.from_geojson(json_data).bbox
+
```
batic, line 358: Did I miss the decoding for `FeatureType.TIMESTAMP`? 🤔

core/eolearn/core/eodata_io.py
```diff
@@ -249,9 +269,11 @@ def load(self):
             return self.loaded_value
 
         with self.filesystem.openbin(self.path, 'r') as file_handle:
-            if self.path.endswith(FileFormat.GZIP.extension()):
+            gzip_extension = FileFormat.GZIP.extension()
+            if self.path.endswith(gzip_extension):
+                path = self.path[:-len(gzip_extension)]
```
zigaLuksic, line 274: consider `os.path.splitext`, might be cleaner

core/eolearn/core/eodata_io.py
```diff
@@ -281,20 +303,69 @@ def _save(self, filesystem, data, path, file_format, compress_level=0):
...
+            if self.feature_type is FeatureType.BBOX:
+                return Geometry.from_geojson(json_data).bbox
+
+            return json_data
+
+        if path.endswith(FileFormat.PICKLE.extension()):
+            warnings.warn(
+                f"File {self.path} with data of type {self.feature_type} is in pickle format which is deprecated "
+                f"since eo-learn version 1.0. Please re-save this EOPatch with the new eo-learn version to "
+                f"update the format. In newer versions this backward compatibility will be removed.",
```
zigaLuksic, line 365: these two lines dont need to be `f` strings (my linters complain)

core/eolearn/core/eodata_io.py
```diff
@@ -8,20 +8,28 @@
...
+import geopandas as gpd
 from geopandas import GeoDataFrame, GeoSeries
+from sentinelhub import Geometry, CRS
+from sentinelhub.exceptions import SHUserWarning
 from sentinelhub.os_utils import sys_is_windows
 
 from .constants import FeatureType, FileFormat, OverwritePermission
 from .utilities import FeatureParser
 
+warnings.simplefilter('default', DeprecationWarning)
```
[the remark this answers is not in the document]
AleksMat: Actually we are not trying to hide it - we are trying to show it. Without this it turns out that warnings would not show in all cases. But I also agree that this is not the cleanest as we are changing behavior for all possible deprecation warnings. A cleaner way would be to have custom deprecation warnings like we have [here](https://github.com/sentinel-hub/sentinelhub-py/blob/v3.4.2/sentinelhub/exceptions.py#L43) in sh-py. If you agree I would add `EODeprecationWarning` and raise it throughout eo-learn package.

core/eolearn/core/eodata_io.py
```diff
@@ -47,9 +55,9 @@ def save_eopatch(eopatch, filesystem, patch_location, features=..., overwrite_pe
 
     features_to_save = []
     for ftype, fname, path in eopatch_features:
-        feature_io = FeatureIO(filesystem, path)
+        feature_io = FeatureIO(filesystem, path, ftype)
         data = eopatch[(ftype, fname)]
-        file_format = FileFormat.NPY if ftype.is_raster() else FileFormat.PICKLE
+        file_format = _get_file_format(ftype)
```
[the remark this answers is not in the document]
AleksMat: I considered this but I don't want to tie `FileFormat` to `FeatureType`. In the future certain feature types might support multiple file formats. So I would prefer that `FileFormat` and `FeatureType` remain independent from each other.

core/eolearn/core/eodata_io.py
```diff
@@ -221,18 +229,30 @@ def _to_lowercase(ftype, fname, *_):
     return ftype, fname if fname is ... else fname.lower()
 
 
+def _get_file_format(feature_type: FeatureType) -> FileFormat:
```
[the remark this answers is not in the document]
AleksMat: I would prefer not to. I would only consider making this a method of `FeatureIO` class, which joins both `FeatureType` and `FileFormat` logic, but even that I don't find necessary for now.

core/eolearn/core/eodata_io.py
AleksMat: I agree about the order (or at least I don't have a strong preference).

I wanted to avoid adding `feature_type` parameter here but it was necessary. As you can see it is required for:

- additional serialization/deserialization steps for certain feature types,
- nicer error messages.

core/eolearn/core/eodata_io.py
AleksMat: No, it is not intended that we would have multiple layers. But it turns out that if you don't specify layer name then geopandas does the following:

- if `file` would be a filename then it uses its name for layer name,
- if `file` is a file object then it creates a random hash name for layer name which is not very nice.

So by setting `layer` parameter I'm reconstructing the `geopandas` functionality of nice names even though we are passing it a file object.

Yes, I also think it's better to be explicit by specifying `driver="GPKG"`.

core/eolearn/core/eodata_io.py
AleksMat: It's because of the thing you also noticed yourself - if you save Geopackage and then load it back the CRS definition will not necessarily remain completely the same. It turns out that this happens not only for WGS84 but even for Popular Web Mercator and others. So by casting it to `sentinelhub.CRS` class and back we are correcting these definitions to the ones used in our code base and achieving more consistency in the process of saving and loading.

core/eolearn/core/eodata_io.py
AleksMat: I expect that in the future we'll want even further tweak this part but for now I cannot predict all potential issues.

core/eolearn/core/eodata_io.py
AleksMat: That already happens by default when setting the list of timestamps in EOPatch. This part of code is [here](https://github.com/sentinel-hub/eo-learn/blob/v0.10.1/core/eolearn/core/eodata.py#L104). It is convenient beyond just loading data so I would leave it like that.

core/eolearn/core/eodata_io.py
AleksMat: Nice find! Although I'll use `fs.path.splitext` as we are working with OS-agnostic paths.

core/eolearn/core/eodata_io.py
AleksMat: Thanks, I'm surprised pylint doesn't :thinking:.

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
batic: I agree, but perhaps in another issue/MR?

core/eolearn/core/eodata_io.py
```diff
@@ -47,9 +55,9 @@ def save_eopatch(eopatch, filesystem, patch_location, features=..., overwrite_pe
 
     features_to_save = []
     for ftype, fname, path in eopatch_features:
-        feature_io = FeatureIO(filesystem, path)
+        feature_io = FeatureIO(filesystem, path, ftype)
```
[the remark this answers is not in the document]
AleksMat: Ok, I made the change.

core/eolearn/core/eodata_io.py
AleksMat: Changed.

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
batic: I agree, but even in case that `FeatureType` will support several types (e.g. `FeatureType.DATA` in `npy`, `tiff`, `zarr`...), one will have to be the default, and the code will have to be changed so that the user specifies his preferred file type. In both cases, this will have to be tied to the `feature` itself. 

At the moment, when we only have the default file type, it still makes sense to me to have this tied to the `FeatureType`, as they are not really independent.

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
batic: Feels wrong that we have `FeatureType`, `FileFormat`, and `FeatureIO` dealing with that (not to mention `sentinelhub.constants.MimeType` which is also similar to some extent). I believe things can be (and should) be unified and improved.

core/eolearn/core/eodata_io.py
batic: > I wanted to avoid adding feature_type parameter here but it was necessary.

Indeed; `FeatureIO`, `FeatureType` and `FileFormat` are very intertwined. I believe there is better architectural construct to tie them together where adding parameters won't be necessary.

core/eolearn/core/eodata_io.py
batic: 👍

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
AleksMat: Yes, I'll prepare this in the next PR.

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
AleksMat: Hm, I kinda hoped that `FeatureType` would contain only the in-memory logic for features and `FileFormat` only file format logic. That is because nowhere else in eo-learn you need to know about serialization formats of certain feature types.

But if you insist I'll add a method /property about this to `FeatureType` class.

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
AleksMat: Ok, we can get rid of `FileFormat` and use the existing `MimeType` class. I will have to add a few extra non-standard mime types though (e.g. pickle, geojson, geopackage) so there might be some issues about that.

I'll prepare a PR for sh-py about this and update this PR.

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
batic: They are different architectural approaches, I guess. I don't insist; if you feel more comfortable this way, keep it as is. 

Personally I find the delegation of responsibilities a bit weird.

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
batic: Can (should?) happen in another PR.

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
AleksMat: Done in https://github.com/sentinel-hub/eo-learn/pull/348

core/eolearn/core/eodata_io.py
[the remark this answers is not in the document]
AleksMat: Done in https://github.com/sentinel-hub/eo-learn/pull/348

PR #349, opened 2022-01-05
Custom warnings and cleaning deprecations

AleksMat:
Changes:

- introduced custom eo-learn warning classes which were discussed in https://github.com/sentinel-hub/eo-learn/pull/347#discussion_r778708790
- cleaned up code that was deprecated for a while and updated tests.


zigaLuksic (approved):
I like the more expressive warnings/errors. I guess having specialized errors is fine as long as reasonable `ŧry/except` blocks catch them (should be taken care of by subclassing). I noticed a lot of warnings were transfered from logging to warnings, we do log warnings right?

PR #350, opened 2022-01-06
Support for aws_profile parameter

AleksMat:
A minor PR to support loading AWS credentials from profile. 

So far I haven't added any unit test for this as it would be a bit tricky. I might have to mock `boto3.Session` object or something like that. Let me know if you think we need a test for this.

### DIFF
--- a/core/eolearn/core/fs_utils.py
+++ b/core/eolearn/core/fs_utils.py
@@ -10,7 +10,7 @@ file in the root directory of this source tree.
 """
 import os
 from pathlib import Path, PurePath
-from typing import Optional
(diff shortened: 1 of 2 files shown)


zigaLuksic (commented):
If it's too difficult i think we can manage without a unit test. On a different note, since you added all types of parameters, it would be neat if you added return types of functions as well (if there is a sensible type for it)

AleksMat:
Added return types. I also decided to practice mocking and prepared a unit test for this.

zigaLuksic (approved):

PR #354, opened 2022-01-10
Update code for projecting to 2D in rasterization

zigaLuksic:
Fixes the issue with Shapely 1.8 removing the `to_wkt` method we used to project 3D -> 2D in rasterization task.

### DIFF
--- a/geometry/eolearn/geometry/transformations.py
+++ b/geometry/eolearn/geometry/transformations.py
@@ -10,7 +10,7 @@ This source code is licensed under the MIT license found in the LICENSE
 file in the root directory of this source tree.
 """
 
-import functools
+from functools import partial
 import logging
 import warnings
 
@@ -191,7 +191,7 @@ class VectorToRasterTask(EOTask):
         if vector_data.geometry.has_z.any():
             warnings.warn('Given vector polygons contain some 3D geometries, they will be projected to 2D',
                           EORuntimeWarning)
-            vector_data.geometry = vector_data.geometry.apply(lambda geo: shapely.wkt.loads(geo.to_wkt()))
+            vector_data.geometry = vector_data.geometry.map(partial(shapely.ops.transform, lambda *args: args[:2]))
 
         return vector_data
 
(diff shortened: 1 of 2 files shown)


geometry/eolearn/geometry/transformations.py
```diff
@@ -10,7 +10,7 @@
 file in the root directory of this source tree.
 """
 
-import functools
+from functools import partial
```
AleksMat, line 13: [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) recommends `import functools` and then calling `functools.partial`. That is because the name `partial` by itself can easily be mistaken for a variable or some custom function.

geometry/eolearn/geometry/transformations.py
```diff
@@ -207,7 +207,7 @@ def _vector_data_to_shape_iterator(self, vector_data, join_per_value):
         if join_per_value:
             classes = np.unique(vector_data[self.values_column])
             grouped = (vector_data.geometry[vector_data[self.values_column] == cl] for cl in classes)
-            grouped = (shapely.ops.cascaded_union(group) for group in grouped)
+            grouped = (shapely.ops.unary_union(group) for group in grouped)
```
AleksMat, line 210: Should we then add `shapely>=1.8.0` into requirements (this is a few months old version), or should we perhaps here do something like we have in [here in sh-py](https://github.com/sentinel-hub/sentinelhub-py/blob/v3.4.3/sentinelhub/areas.py#L105)?

geometry/eolearn/geometry/transformations.py
zigaLuksic: but then i have to break the code up in 3 lines :(
(will fix)

geometry/eolearn/geometry/transformations.py
zigaLuksic: Not a big fan of checking versions at runtime, but also not a big fan of requirements restrictions. We can use the same trick yeah.

AleksMat (approved):

PR #360, opened 2022-01-17
EOExecutor improvements, part 3

AleksMat:
Implements various improvements in `EOExecutor` class and especially in the generated report. Changes are documented in commit messages.

### DIFF
--- a/core/eolearn/core/eoexecution.py
+++ b/core/eolearn/core/eoexecution.py
@@ -21,6 +21,7 @@ import multiprocessing
 import warnings
 from enum import Enum
 from logging import Logger, Handler, Filter
+from dataclasses import dataclass
 from typing import Sequence, List, Tuple, Dict, Optional, Callable, Iterable, TypeVar
 
 import fs
@@ -37,10 +38,7 @@ MULTIPROCESSING_LOCK = None
 
 _InputType = TypeVar('_InputType')
 _OutputType = TypeVar('_OutputType')
-_LogsHandlerFactoryType = Callable[[str], Handler]
-_ExecutorProcessingArgsType = Tuple[
-    EOWorkflow, Dict[EONode, Dict[str, object]], Optional[str], bool, Filter, _LogsHandlerFactoryType
-]
+_HandlerFactoryType = Callable[[str], Handler]
 
 
 class _ProcessingType(Enum):
@@ -50,6 +48,18 @@ class _ProcessingType(Enum):
     MULTIPROCESSING = 'multiprocessing'
     MULTITHREADING = 'multithreading'
     RAY = 'ray'
+
+
+@dataclass(frozen=True)
+class _ProcessingData:
+    """Data to be used in EOExecutor processing. This will be passed to a process pool, so everything has to be
+    serializable with pickle."""
+    workflow: EOWorkflow
+    workflow_kwargs: Dict[EONode, Dict[str, object]]
+    log_path: Optional[str]
+    filter_logs_by_thread: bool
+    logs_filter: Optional[Filter]
+    logs_handler_factory: _HandlerFactoryType
 
 
 class EOExecutor:
@@ -64,7 +74,7 @@ class EOExecutor:
     def __init__(self, workflow: EOWorkflow, execution_kwargs: Sequence[Dict[EONode, Dict[str, object]]], *,
                  execution_names: Optional[List[str]] = None, save_logs: bool = False, logs_folder: str = '.',
                  filesystem: Optional[fs.base.FS] = None, logs_filter: Optional[Filter] = None,
-                 logs_handler_factory: _LogsHandlerFactoryType = logging.FileHandler):
+                 logs_handler_factory: _HandlerFactoryType = logging.FileHandler):
         """
         :param workflow: A prepared instance of EOWorkflow class
         :param execution_kwargs: A list of dictionaries where each dictionary represents execution inputs for the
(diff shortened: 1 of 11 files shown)


core/eolearn/core/eoexecution.py
```diff
@@ -58,59 +61,72 @@ class EOExecutor:
...
                              "execution arguments")
         return execution_names
 
+    @staticmethod
+    def _parse_logs_filesystem(filesystem: Optional[fs.base.FS], logs_folder: str) -> Tuple[fs.base.FS, str]:
+        """ Ensures a filesystem and a file path relative to it.
+        """
+        if filesystem is None:
+            return get_base_filesystem_and_path(logs_folder)
+        return filesystem, logs_folder
```
zigaLuksic, line 128: So if filesystem is `None` but `logs_folder` is something, the path gets ignored? Is this the desired behaviour?

core/eolearn/core/eoexecution.py
```diff
@@ -129,14 +145,15 @@ def run(self, workers: int = 1, multiprocess: bool = True) -> List[WorkflowResul
...
+            self.filesystem.makedirs(self.report_folder, recreate=True)
 
-        log_paths = self.get_log_paths() if self.save_logs else [None] * len(self.execution_args)
+        log_paths = self.get_log_paths(full_path=True) if self.save_logs else [None] * len(self.execution_kwargs)
 
         filter_logs_by_thread = not multiprocess and workers > 1
-        processing_args = [(self.workflow, init_args, log_path, filter_logs_by_thread, self.logs_filter)
-                           for init_args, log_path in zip(self.execution_args, log_paths)]
         processing_type = self._get_processing_type(workers, multiprocess)
+        processing_args = [
```
zigaLuksic, line 154: This is really becoming slightly unwieldy. I would again suggest having a dataclass, either for all parameters or just for the ones pertaining to logging

As a further point, mypy is detecting an error, because some of the arguments are allegedly `Optional` but the function signatures does not have them listed as optional

core/eolearn/core/eoexecution.py
```diff
@@ -176,15 +193,15 @@ def _run_execution(self, processing_args: List[_ExecutorProcessingArgsType], wor
...
+    def _try_add_logging(cls, log_path: Optional[str], filter_logs_by_thread: bool, logs_filter: Optional[Filter],
+                         logs_handler_factory: _LogsHandlerFactoryType) -> Tuple[Optional[Logger], Optional[Handler]]:
         """ Adds a handler to a logger and returns them both. In case this fails it shows a warning.
         """
         if log_path:
             try:
                 logger = logging.getLogger()
                 logger.setLevel(logging.DEBUG)
-                handler = cls._get_log_handler(log_path, filter_logs_by_thread, logs_filter)
+                handler = cls._get_log_handler(log_path, filter_logs_by_thread, logs_filter, logs_handler_factory)
```
zigaLuksic, line 204: i would change this to `_build_log_handler` or something like that. Getting a handler would be the call to the factory, no?

core/eolearn/core/eonode.py
```diff
@@ -42,7 +42,7 @@ def __post_init__(self):
         if self.name is None:
             super().__setattr__('name', self.task.__class__.__name__)
 
-        super().__setattr__('uid', generate_uid(self.name))
+        super().__setattr__('uid', generate_uid(self.task.__class__.__name__))
```
zigaLuksic, line 45: This ignores custom node names. IMO it would be better to use `self.get_custom_name()`

visualization/eolearn/visualization/eoexecutor_visualization.py
```diff
@@ -86,17 +90,47 @@ def make_report(self, include_logs: bool = True):
...
+            )
+            exception_stats[error_node.node_uid][exception_str] += 1
+
+        sorted_exception_stats = []
+        for node in self.eoexecutor.workflow.get_nodes():
+            if node.uid not in exception_stats:
+                continue
+
+            node_stats = exception_stats[node.uid]
+            sorted_exception_stats.append((node.name, node.uid, sorted(node_stats.items(), key=lambda item: -item[1])))
```
zigaLuksic, line 130: It is not clear what this line does at first glance, a comment would be nice

zigaLuksic (commented):
Carrying the handler generator around makes the code a bit less readable, but I see no better way

core/eolearn/core/eoexecution.py
AleksMat: No, In that case we assume `logs_folder` is an absolute path, hence `get_base_filesystem_and_path` will split it into a filesystem object and a path relative to the object.

core/eolearn/core/eoexecution.py
```diff
@@ -30,14 +29,18 @@
...
+from .fs_utils import get_base_filesystem_and_path, get_full_path
 from .utilities import LogFileFilter
 
 LOGGER = logging.getLogger(__name__)
 MULTIPROCESSING_LOCK = None
 
 _InputType = TypeVar('_InputType')
 _OutputType = TypeVar('_OutputType')
-_ExecutorProcessingArgsType = Tuple[EOWorkflow, Dict[EONode, Dict[str, object]], Optional[str], bool, Filter]
+_LogsHandlerFactoryType = Callable[[str], Handler]
```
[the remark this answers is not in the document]
AleksMat: Changed to `_HandlerFactoryType`. I still like that all types end with the name `Type`.

core/eolearn/core/eoexecution.py
AleksMat: Done

core/eolearn/core/eonode.py
AleksMat: With task class name it looks much better in reports where we then write e.g.:

```
Load features (LoadTask-c2cabaa677a311ec8d2e-ae4c34840284) 
```

(We show both custom name and uid.)

visualization/eolearn/visualization/eoexecutor_visualization.py
AleksMat: I split this part into a new method where I wrote more description into a docstring.

core/eolearn/core/eoexecution.py
AleksMat: Done - changed to a dataclass.

core/eolearn/core/eoexecution.py
zigaLuksic: ah, i missed the part where it is passed to `get_base_filesystem_and_path`

core/eolearn/core/eonode.py
zigaLuksic: fine, let's leave it at that for now. Not likely to cause issues i hope

zigaLuksic (approved):

PR #362, opened 2022-01-20
Reduce amount of mypy warnings in eo-learn.core

zigaLuksic:
Since we recently started annotating our code mypy is starting to complain. This removes a fair amount of errors from the code without changing functionality (and without cluttering the code too much). There are also some small documentation changes.

### DIFF
--- a/core/eolearn/core/eoworkflow.py
+++ b/core/eolearn/core/eoworkflow.py
@@ -315,9 +315,8 @@ class EOWorkflow:
 
     def get_node_with_uid(self, uid: Optional[str], fail_if_missing: bool = False) -> Optional[EONode]:
         """Returns node with give uid, if it exists in the workflow."""
-        relevant_nodes = [node for node in self._nodes if node.uid == uid]
-        if relevant_nodes:
-            return relevant_nodes[0]  # uniqueness of uid within a workflow
+        if uid in self._uid_dict:
+            return self._uid_dict[uid]
         if fail_if_missing:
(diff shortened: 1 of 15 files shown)


zigaLuksic:
I also added the `get_node_with_uid` method

core/eolearn/core/eoworkflow.py
```diff
@@ -308,6 +313,15 @@ def get_nodes(self) -> List[EONode]:
         """
         return self._nodes[:]
 
+    def get_node_with_uid(self, uid: Optional[str], fail_if_missing: bool = False) -> Optional[EONode]:
+        """Returns node with give uid, if it exists in the workflow."""
+        relevant_nodes = [node for node in self._nodes if node.uid == uid]
```
AleksMat, line 318: Why not just checking if `uid` exists in `self._uid_dict`? This way you achieve O(1) time complexity instead of O(n) and uniqueness is ensured.

core/eolearn/tests/test_eoworkflow.py
```diff
@@ -9,6 +9,7 @@
 """
 import concurrent.futures
 import datetime as dt
+import pylint
```
AleksMat, line 12: Unnecessary import

core/eolearn/core/eoworkflow.py
zigaLuksic: oh yes, ofcourse, I now have the `_uid_dict` at disposal. Will change

AleksMat (approved):

PR #364, opened 2022-01-26
Package updates

AleksMat:
Various package-related updates:

- package versions,
- handling extra dependencies and imports,
- module naming changes,
- docs updates and fixes.


io/eolearn/io/geometry_io.py
```diff
@@ -40,6 +41,10 @@ def __init__(self, feature, reproject=True, clip=False, config=None):
         self.reproject = reproject
         self.clip = clip
 
+    @abc.abstractmethod
```
zigaLuksic, line 44: If you want this to be a true abstract base class don't forget to set the meta class

zigaLuksic (approved):
Looks good

PR #365, opened 2022-01-28
EOPatch.__repr__ changes

AleksMat:
EOPatch representation is now only showing feature types which have at least one defined feature.

### DIFF
--- a/core/eolearn/tests/test_eodata.py
+++ b/core/eolearn/tests/test_eodata.py
@@ -19,13 +19,6 @@ from eolearn.core import EOPatch, FeatureType, FeatureTypeSet
 from eolearn.core.eodata_io import FeatureIO
 
 
-def test_loading_valid(test_eopatch_path):
-    eop = EOPatch.load(test_eopatch_path)
-
-    repr_str = eop.__repr__()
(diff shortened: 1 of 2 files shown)


core/eolearn/tests/test_eodata.py
```diff
@@ -112,10 +105,26 @@ def test_invalid_characters():
...
 
 
+def test_repr(test_eopatch_path):
+    test_eopatch = EOPatch.load(test_eopatch_path)
+
+    for eopatch in [test_eopatch, EOPatch()]:
+        repr_str = repr(eopatch)
+
+        assert isinstance(repr_str, str)
+        assert repr_str == eopatch.__repr__()
```
zigaLuksic, line 115: is this not the same thing?

zigaLuksic (approved):

core/eolearn/tests/test_eodata.py
AleksMat: Hm, I though that `repr` function falls back to `__str__` if `__repr__` is not defined but now I see that is actually the default definition of `__repr__`. Will fix.

PR #370, opened 2022-01-31
SH input tasks for cases with no data

AleksMat:
In case of no available data `SentinelHubInputTask` and `SentinelHubEvalscriptTask` now create arrays with time dimension 0 instead of raising an error.

Additionally, I've a few of minor improvements in `get_available_timestamps` and in tests.


zigaLuksic (approved):
Does such an EOPatch have the timestamp correctly set to `[]` and not `None`? Looks great

AleksMat:
> Does such an EOPatch have the timestamp correctly set to [] and not None?

It should have. We only set timestamps in case timestamps are not None. Also the default value of `EOPatch.timestamp` is `[]`.

PR #371, opened 2022-01-31
Cleanup of core module

zigaLuksic:
- Adds a lot of `.` to ends of sentences
- Removes some outdated methods from EOPatches
- Restructures utilities into their own subfolder
- Other minor things


core/eolearn/core/__init__.py
```diff
@@ -28,8 +28,8 @@
     MergeEOPatchesTask,
 )
 
-from .fs_utils import get_filesystem, load_s3_filesystem
-from .utilities import deep_eq, negate_mask, constant_pad, get_common_timestamps, bgr_to_rgb, FeatureParser
-
+from .utilities.filesystem import get_filesystem, load_s3_filesystem
```
AleksMat, line 31: I suggest renaming:

- folder `utilities` -> `utils` (we have same in `eo-grow`),
- module `filesystem` -> `fs` (we have same in `eo-grow` and its shorter),
- module `feature_parser` -> `parsing`
- module `other` -> `common`

core/eolearn/core/utilities/other.py
```diff
@@ -0,0 +1,199 @@
...
+import geopandas as gpd
+from geopandas.testing import assert_geodataframe_equal
+
+from ..exceptions import EODeprecationWarning
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class LogFileFilter(Filter):
```
AleksMat, line 32: Perhaps this class could be in a separate utils module `logging.py`? It is very specific compared to other things in this module and I imagine that in the future we might have more logging utils.

core/eolearn/core/utilities/other.py
```diff
@@ -0,0 +1,199 @@
...
+
+        for key in fst_obj:
+            if not deep_eq(fst_obj[key], snd_obj[key]):
+                return False
+        return True
+
+    return fst_obj == snd_obj
+
+
+def constant_pad(X, multiple_of, up_down_rule="even", left_right_rule="even", pad_value=0):
```
AleksMat, line 104: Maybe we can rename this `X` to `array` or `image` or something similar? This way we get rid of `# pylint: disable=invalid-name`.

core/eolearn/core/utilities/other.py
```diff
@@ -0,0 +1,199 @@
...
+
+    The ID is composed from the prefix, a hexadecimal string obtained from the current time and a random hexadecimal
+    string. This makes the uid sufficiently unique.
+    """
+    time_uid = uuid.uuid1(node=0).hex[:-12]
+    random_uid = uuid.uuid4().hex[:12]
+    return f"{prefix}-{time_uid}-{random_uid}"
+
+
+def renamed_and_deprecated(deprecated_class):
```
AleksMat, line 176: I think this function would belong to `exceptions` module.

AleksMat (changes requested):
Now that some modules in the core were moved also please update rst documentation pages accordingly.

core/eolearn/core/__init__.py
zigaLuksic: I used `filesystem` because `fs` was flagged by my linters as a self import (since we use the `fs` package)

AleksMat (approved):

PR #374, opened 2022-02-02
Remove and update ml_tools files

zigaLuksic:


ml_tools/eolearn/ml_tools/train_test_split.py
```diff
@@ -27,8 +27,10 @@ class TrainTestSplitType(Enum):
 class TrainTestSplitTask(EOTask):
     """Randomly assign each pixel or groups of pixels to multiple subsets (e.g., test/train/validate).
 
-    Input pixels are defined by an input feature (e.g., MASK_TIMELESS with polygon ids, connected component ids, or
-    similar), that groups together pixels with similar properties.
+    When sampling PER_PIXEL the input feature defines the shape of the output feature. For PER_CLASS and PER_VALUE the
```
AleksMat, line 30: > When sampling PER_PIXEL the input feature defines the shape of the output feature. 

Is that not true for all 3 strategies?

ml_tools/eolearn/ml_tools/train_test_split.py
```diff
@@ -92,30 +99,29 @@ def execute(self, eopatch, *, seed=None):
         :param eopatch: input EOPatch
         :type eopatch: EOPatch
         :param seed: An argument to be passed to numpy.random.seed function.
-        :type seed: numpy.int64
+        :type seed: numpy.int64 or None
```
AleksMat, line 102: I would go with `int or None`. because it is just a number.

AleksMat (changes requested):
One important thing: make sure to add `ml_tools/eolearn/ml_tools/extra/__init__.py` file otherwise packaging procedure will fail.

ml_tools/eolearn/ml_tools/train_test_split.py
zigaLuksic: will specify that in that case it ONLY defines the shape

PR #375, opened 2022-02-04
New EOPatch plotting

AleksMat:
Moves old plotting functionalities into an extension and instead adds new ones with basic `matplotlib`.


visualization/eolearn/visualization/eopatch.py
```diff
@@ -1,355 +1,263 @@
...
-        out = data_gpd.loc[data_gpd[self.timestamp_column] == timestamp]
-        return gv.Polygons(out, crs=ccrs.epsg(int(crs.value)))
-
-    def plot_vector_timeless(self, feature_name):
-        """Plot FeatureType.VECTOR_TIMELESS data
+        config = config or PlotConfig()
+        super().__init__(eopatch, feature, config=config, **kwargs)
+        self.config = cast(PlotConfig, self.config)
+
+        if axes is not None and not isinstance(axes, np.ndarray):
```
zigaLuksic, line 102: what if you have only one line or one column? then i think you get a 1D array of axes. Is this ok? I think you almost always flatten beforehand so it shouldn't be an issue.

AleksMat: We actually use `squeeze=False` when creating axes. Because of that we always get a 2D grid of subplots. Flattening is a precaution if a user would give only an 1D array of axes.

core/eolearn/core/eodata.py
```diff
@@ -537,56 +546,49 @@ def consolidate_timestamps(self, timestamps):
...
+            indices will be used.
+        :param channels: A list or a slice of indices on channels axis to be used for plotting. If not provided all
+            indices will be used.
+        :param channel_names: Names of channels of the last dimension in the given raster feature.
+        :param rgb: If provided, it should be a list of 3 indices of RGB channels to be plotted. It will plot only RGB
+            images with these channels. This only works for raster features with spatial dimension.
+        :param backend: A type of plotting backend.
+        :param config: A configuration object with advanced plotting parameters.
+        :param kwargs: Parameters that are specific to a specified plotting backend.
+        :return: A plot object
```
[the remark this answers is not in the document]
AleksMat: You are right. Fixed.

zigaLuksic (approved):

PR #385, opened 2022-02-16
Support AWS session tokens

theirix:
Use a config parameter `aws_session_token` introduced with [sentinelhub-py 3.4.4](https://github.com/sentinel-hub/sentinelhub-py/releases/tag/v3.4.4) that allows configuring libraries with SHConfig instances with temporary IAM credentials that are usually used at EC2 machines (access id, key and STS token are all required to initialize boto3).


theirix:
Need to rebase to 1.0

PR #393, opened 2022-02-24
Memory and speed optimization in simple eopatch merging procedures

zigaLuksic:
When merging eopatches with temporal features this optimization recognizes when the merge operation would not change anything for a given feature, saving time and (more importantly) memory.

### DIFF
--- a/core/eolearn/core/eodata_merge.py
+++ b/core/eolearn/core/eodata_merge.py
@@ -156,13 +156,9 @@ def _merge_time_dependent_raster_feature(eopatches, feature, operation, order_ma
     """Merges numpy arrays of a time-dependent raster feature with a given operation and masks on how to order and join
     time raster's time slices.
     """
-    if optimize:
-        relevant_features = [eopatch[feature] for eopatch in eopatches if feature in eopatch]
-        if _all_equal(relevant_features):
-            return relevant_features[0]
 
     merged_array, merged_order_mask = _extract_and_join_time_dependent_feature_values(
-        eopatches, feature, order_mask_per_eopatch
+        eopatches, feature, order_mask_per_eopatch, optimize
     )
 
     # Case where feature array is already in the correct order and doesn't need splitting, which includes a case
@@ -193,7 +189,7 @@ def _merge_time_dependent_raster_feature(eopatches, feature, operation, order_ma
     return np.array(split_arrays)
 
 
-def _extract_and_join_time_dependent_feature_values(eopatches, feature, order_mask_per_eopatch):
+def _extract_and_join_time_dependent_feature_values(eopatches, feature, order_mask_per_eopatch, optimize):
     """Collects feature arrays from EOPatches that have them and joins them together. It also joins together
     corresponding order masks.
     """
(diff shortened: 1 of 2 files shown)


core/eolearn/core/eodata_merge.py
```diff
@@ -143,10 +144,23 @@ def _merge_timestamps(eopatches, reduce_timestamps):
...
+
+
+def _merge_time_dependent_raster_feature(eopatches, feature, operation, order_mask_per_eopatch, optimize):
     """Merges numpy arrays of a time-dependent raster feature with a given operation and masks on how to order and join
     time raster's time slices.
     """
+    if optimize:
+        relevant_features = [eopatch[feature] for eopatch in eopatches if feature in eopatch]
+        if _all_equal(relevant_features):
+            return relevant_features[0]
```
AleksMat, line 162: The problem here is that some EOPatches might not necessarily have the feature. In those cases this will fail. The correct place where to check this is actually inside `_extract_and_join_time_dependent_feature_values` just before arrays are concatenated.

Perhaps we can even add a test where we merge 2 EOPatches that have a feature and 1 EOPatch that doesn't have it?

AleksMat (commented):
Only 1 thing to change, otherwise I agree with the fix. :+1:

core/eolearn/core/eodata_merge.py
zigaLuksic: That is done in line 160 and I am sure there are tests for it already. It was also tested in eo-grow, because each of the indices was contained in exactly one eopatch.

AleksMat: I see I missed that. Ok, then there is also the question of how we handle the case where timestamps are not sorted in an increasing order. What I think currently happens:

- timestamps in EOPatch get sorted,
- time dimension of temporal feature doesn't get modified.

Hence we get a discrepancy. But if this handling would be inside `_extract_and_join_time_dependent_feature_values` then the code would still reach the middle part of `_merge_time_dependent_raster_feature` where time dimension would be sorted for this case.

zigaLuksic: this merging does waaay too much :joy: 
The issue is that `_extract_and_join_time_dependent_feature_values` also does a lot (makes masks, merges arrays and all that) so there is no neat way to inject this shortcircuit. I'll see if I can re-order things and make it work.

AleksMat: Timestamp sorting is sadly a necessary step of merging. Otherwise there is no way how to define a merge over temporal dimension.

AleksMat: Note that `_extract_and_join_time_dependent_feature_values` doesn't make masks, it just takes the according pre-prepared masks. I don't see why you couldn't simply change:

```Python
    if len(arrays) == 1:
        return arrays[0], order_masks[0]
```

into

```Python
    if len(arrays) == 1 or (optimize and _all_equal(arrays)):
        return arrays[0], order_masks[0]
```

It should be as simple as that.

AleksMat (approved):

PR #397, opened 2022-03-24
Moved code coverage config to pyproject.toml to solve config clashes

AleksMat:
Same solution as in `sentinelhub-py`. Without this we were getting warnings:

```
/home/runner/work/eo-learn/eo-learn/io/eolearn/tests/test_sentinelhub_process.py:71: PytestUnknownMarkWarning: Unknown pytest.mark.sh_integration - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
```

when running pytest with code coverage. That is because pytest can only look at 1 configuration file so it had to look at `.coverage` and didn't see markers defined in `pyproject.toml`. Now it can only look at `pyproject.toml`.


pyproject.toml
```diff
@@ -13,3 +13,51 @@ line_length = 120
...
+    "mask/eolearn/__init__.py",
+    "mask/eolearn/mask/__init__.py",
+    "ml_tools/setup.py",
+    "ml_tools/eolearn/tests/*",
+    "ml_tools/eolearn/__init__.py",
+    "ml_tools/eolearn/ml_tools/__init__.py",
+    "visualization/setup.py",
+    "visualization/eolearn/tests/*",
+    "visualization/eolearn/__init__.py",
+    "visualization/eolearn/visualization/__init__.py"
```
zigaLuksic, line 62: is there no pattern to tell it to omit all __init__ files and test folders?

AleksMat: Ah, of course! :facepalm: 

Now it is much better. :)

PR #405, opened 2022-05-02
Extended is_discrete_type function and wrote unit tests for it

AleksMat:
I extended `is_discrete_type` function so that it works both with `type` and `numpy.dtype` objects (because `numpy` also works with both). Then I wrote precise unit tests for it.

More info about type hierarchy can be found in [`numpy` docs](https://numpy.org/doc/stable/reference/arrays.scalars.html).

### DIFF
--- a/core/eolearn/core/eodata_io.py
+++ b/core/eolearn/core/eodata_io.py
@@ -17,9 +17,12 @@ import warnings
 from collections import defaultdict
 
 import fs
+import fs.move
 import geopandas as gpd
 import numpy as np
 import pandas as pd
+from fs.base import FS
+from fs.osfs import OSFS
 from fs.tempfs import TempFS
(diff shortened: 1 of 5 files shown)


AleksMat:
Today a new version of `fs` package has been released and it broke a couple of things. The easy-to-fix one was the imports. The following doesn't work anymore

```Python
import fs

fs.move.move_file
```
Instead you have to call `import fs.move`. I fixed this in a commit in this branch. Additionally I updated a few other imports from `fs` because PyCharm was also warning me about missing import paths.

The 2nd issue is more problematic and potentially a bug in `fs` package. It will be addressed in a separate PR. So I propose that we ignore failing tests in this PR because failures are not related to these changes.

zigaLuksic (approved):

PR #411, opened 2022-05-30
Enable session sharing in sentinelhub tasks via execute arguments

zigaLuksic:
This is a suggestion on how we can do session sharing in `eo-learn`.

My first idea was to add it as an `init` parameter, but it is easier to add it as an execution argument. It should be no different were it an `__init__` parameter, so I am willing to adjust that.

I checked that it works with multiprocessing:
```python
thread = SessionSharingThread(session)
thread.start() 

try:
    with ProcessPoolExecutor(max_workers=3) as executor:
        futures = [executor.submit(task.execute, None, bbox, time_interval, collect_shared_session) for _ in range(3)]

finally:
    thread.join()
```

And it also works with `Ray`:
```python
@ray.remote
class RaySessionActor:
    def __init__(self, session):
        self.session = session

    def get_valid_session(self):
        token = self.session.token
        return SentinelHubSession.from_token(token)

@ray.remote
def remote_execution(actor):
    task.execute(None, bbox, time_interval, lambda: ray.get(actor.get_valid_session.remote()))

actor = RaySessionActor.remote(session)

futures = [remote_function.remote(actor) for _ in range(3)]
ray.get(futures)
```


io/eolearn/io/sentinelhub_process.py
```diff
@@ -66,7 +75,13 @@ def __init__(self, data_collection, size=None, resolution=None, cache_folder=Non
         self.data_collection = DataCollection(data_collection)
         self.cache_folder = cache_folder
 
-    def execute(self, eopatch=None, bbox=None, time_interval=None):
+    def execute(
+        self,
+        eopatch=None,
+        bbox=None,
+        time_interval=None,
+        session_obtainer: Optional[Callable[[], SentinelHubSession]] = None,
```
AleksMat, line 83: As discussed, I have 2 comments here:

- Conceptually it might be better that `session_obtainer` is an `__init__` parameter than an `execute` parameter. This is mainly because I don't see a use case where you would have to provide a different session obtainer for different spatial and temporal parameters. It seems that a task in a workflow will always require the same object here.

- The word "obtainer" sounds a bit weird to me. Maybe name `session_loader` would be more natural?

AleksMat:
Another question - would it perhaps make sense to put the implementation of `RaySessionActor` also into `eo-learn`?

The best place for it would actually be `eolearn.io.extra.ray`, which means we would have to add a Ray extension to `eo-learn-io` package. However, since this actor class really has only a few lines of code I'm not sure if it justifies creating a package extension. :thinking:

io/eolearn/io/sentinelhub_process.py
zigaLuksic: For the first point, the only reason I see why `exec_args` are better is that you can initialize the session sharing object after the workflow is built. In case of ray, one must first create the actor before one can create a `session_obtainer`

Loader has a I/O feel to me, but honestly I don't mind. Will change

zigaLuksic:
> Another question - would it perhaps make sense to put the implementation of `RaySessionActor` also into `eo-learn`?
> 
> The best place for it would actually be `eolearn.io.extra.ray`, which means we would have to add a Ray extension to `eo-learn-io` package. However, since this actor class really has only a few lines of code I'm not sure if it justifies creating a package extension. thinking

I would wait until we are approached about it. It described in `sh-py` and easy to do. The thread solution is more complex so it makes sense to provide it for users.

io/eolearn/io/sentinelhub_process.py
AleksMat: Looks good.

PR #412, opened 2022-06-01
Faster node hash

zigaLuksic:
Because EONode is a frozen dataclass the built-in hashing is done via structural recursion (apparently in a non-memoized way). The slowdown seems to be exponential and in workflows with over 120 nodes the time required becomes an enormous issue.

Since we use EONodes as keys in execution arguments (and in `from_endnodes` and possible elsewhere) this needs to be improved.

The suggested improvement instead hashes the EONodes uid, for which we already assume it's unique (and is often used as a key internally).

### DIFF
--- a/core/eolearn/core/eonode.py
+++ b/core/eolearn/core/eonode.py
@@ -50,7 +50,7 @@ class EONode:
         super().__setattr__("uid", generate_uid(self.task.__class__.__name__))
 
     def __hash__(self) -> int:
-        return self.uid.encode("utf-8").__hash__()
+        return self.uid.__hash__()
 
     def get_name(self, suffix_number: int = 0) -> str:
         """Provides node name according to the class of the contained task and a given number."""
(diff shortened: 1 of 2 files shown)


AleksMat (changes requested):
I completely agree with the fix. However, given the complicated nature of the issue, this PR also requires a new unit test that will properly test this. E.g. a test that creates a dummy workflow with 10K nodes (or whatever the appropriate number for a fast test would be). The test could then measure that the workflow is created in about a second and not in few minutes (as in the code before the fix). 

Without such test we won't be able to know if ever in the future the issue won't reappear because of some other change. Hence, this is the only way to ensure that the code doesn't degrade.

zigaLuksic:
> I completely agree with the fix. However, given the complicated nature of the issue, this PR also requires a new unit test that will properly test this. E.g. a test that creates a dummy workflow with 10K nodes (or whatever the appropriate number for a fast test would be). The test could then measure that the workflow is created in about a second and not in few minutes (as in the code before the fix).
> 
> Without such test we won't be able to know if ever in the future the issue won't reappear because of some other change. Hence, this is the only way to ensure that the code doesn't degrade.

Fair point. I'll try to conjure up something.

AleksMat (approved):

PR #417, opened 2022-06-07
Minor pytest updates

AleksMat:
- In `pytest>=7.0.0` a parameter of `pytest.skip` was renamed - this fixes it.
- Tests now make sure that EO warnings are being raised wherever expected and not raised otherwise.

These updates reduced a bit the number of warnings we get in tests. We still get a ton of warnings because of an issue in `geopandas` that will be fixed in the next `geopandas` version. We also get quite a number of warnings because a utility function in core intentionally tests deprecated numpy types.

### DIFF
--- a/io/eolearn/tests/test_local_io.py
+++ b/io/eolearn/tests/test_local_io.py
@@ -14,7 +14,8 @@ import datetime
 import logging
 import os
 import tempfile
-from typing import Optional, Union
+import warnings
+from typing import Any, Optional, Type, Union
 
 import boto3
 import numpy as np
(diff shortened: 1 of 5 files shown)


io/eolearn/tests/test_local_io.py
```diff
@@ -170,6 +179,19 @@ def test_export_import(test_case, test_eopatch):
         ), "Tiff imported into new EOPatch has different dtype as expected"
 
 
+def _execute_with_warning_control(
```
zigaLuksic, line 182: since you use this (inlined) in `test_vector_to_raster_result`, would it make sense to move it to test_utils?

AleksMat: Hm, not sure. So far we don't have any module for test utils in eo-learn, only in sentinelhub-py, right? So I'm not sure if we should create one only for this.

PR #419, opened 2022-06-09
Add image resizing utility function with better support

zigaLuksic:
The new utility function supports both the CV2 and PIL backend. It tries to clarify certain CV2 errors by analyzing methods and dtypes, since CV2 errors are terrible.

The PIL backend is the default because the antialiasing ensures images are closer to expectations. It also works with all dtypes. For higher dimensional data (e.g. temporal features) using PIL is still about 3-4x slower than CV2, so it is available as the alternative.

It is hard to write precise tests due to antialiasing, but I included a rather basic test to check that they work properly.

### DIFF
--- a/mask/eolearn/mask/utils.py
+++ b/mask/eolearn/mask/utils.py
@@ -61,6 +61,7 @@ class ResizeMethod(Enum):
 
 
 class ResizeLib(Enum):
+    """Backends available for spatial resizing of data."""
     PIL = "PIL"
     CV2 = "cv2"
 
@@ -192,7 +193,7 @@ def resize_images(
                           One of 'nearest', 'linear', 'cubic'. Default is 'linear'.
     """
     warnings.warn(
-        f"The function `resize_images` is deprecated and will be removed. Please switch to `spatially_resize_image`.",
+        "The function `resize_images` is deprecated and will be removed. Please switch to `spatially_resize_image`.",
         EODeprecationWarning,
     )
     inter_methods = {"nearest": cv2.INTER_NEAREST, "linear": cv2.INTER_LINEAR, "cubic": cv2.INTER_CUBIC}
--- a/mask/eolearn/mask/utils.py
+++ b/mask/eolearn/mask/utils.py
@@ -10,23 +10,70 @@ Copyright (c) 2017-2019 Blaž Sovdat, Andrej Burja (Sinergise)
 This source code is licensed under the MIT license found in the LICENSE
 file in the root directory of this source tree.
 """
+import warnings
+from enum import Enum
+from functools import partial
+from typing import Any, Callable, Optional, Tuple
 
 import cv2
 import numpy as np
+from PIL import Image
+
+from eolearn.core.exceptions import EODeprecationWarning
 
 
-def map_over_axis(data, func, axis=0):
+class ResizeMethod(Enum):
+    """Methods available for spatial resizing of data."""
+
+    NEAREST = "nearest"
+    LINEAR = "linear"
+    CUBIC = "cubic"
+
+    def get_cv2_method(self, dtype: Any) -> int:
+        """Obtain the constant specifying the interpolation method for the CV2 library."""
+        if np.issubdtype(dtype, np.floating):
+            choices = {
+                ResizeMethod.NEAREST: cv2.INTER_NEAREST,
+                ResizeMethod.LINEAR: cv2.INTER_LINEAR,
+                ResizeMethod.CUBIC: cv2.INTER_CUBIC,
+            }
+            return choices[self]
+        if np.issubdtype(dtype, np.integer):
+            choices = {
+                ResizeMethod.NEAREST: cv2.INTER_NEAREST_EXACT,
+                ResizeMethod.LINEAR: cv2.INTER_LINEAR_EXACT,
+            }
+            if self not in choices:
+                raise ValueError(
+                    f"The {self.value} interpolation method cannot be used for integers with the CV2 backend."
+                )
+            return choices[self]
+        raise ValueError(f"The dtype {dtype} cannot be processed with the CV2 backend.")
+
+    def get_pil_method(self) -> Image.Resampling:
+        """Obtain the constant specifying the interpolation method for the PIL library."""
+        choices = {
+            ResizeMethod.NEAREST: Image.Resampling.NEAREST,
+            ResizeMethod.LINEAR: Image.Resampling.BILINEAR,
+            ResizeMethod.CUBIC: Image.Resampling.BICUBIC,
+        }
+        return choices[self]
+
+
+class ResizeLib(Enum):
+    PIL = "PIL"
+    CV2 = "cv2"
+
+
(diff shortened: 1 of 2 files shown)


mask/eolearn/mask/utils.py
```diff
@@ -10,23 +10,72 @@
...
+class ResizeMethod(Enum):
+    """Methods available for spatial resizing of data."""
+
+    NEAREST = "nearest"
+    LINEAR = "linear"
+    CUBIC = "cubic"
+
+    def get_cv2_method(self, dtype: Any) -> int:
+        """Obtain the constant specifying the interpolation method for the CV2 library."""
+        if np.issubdtype(dtype, np.floating):
```
AleksMat, line 34: Before this I recommend that you do `dtype` casting like we use [here](https://github.com/sentinel-hub/eo-learn/blob/v1.0.2/core/eolearn/core/utils/common.py#L155). That is because e.g. `np.float32` and `np.dtype('float32')` are not the same things.

mask/eolearn/mask/utils.py
```diff
@@ -10,23 +10,72 @@
...
+
+
+class ResizeMethod(Enum):
+    """Methods available for spatial resizing of data."""
+
+    NEAREST = "nearest"
+    LINEAR = "linear"
+    CUBIC = "cubic"
+
+    def get_cv2_method(self, dtype: Any) -> int:
```
AleksMat, line 32: You can set `dtype: Union[np.dtype, type]` like we do [here](https://github.com/sentinel-hub/eo-learn/blob/v1.0.2/core/eolearn/core/utils/common.py#L153).

mask/eolearn/mask/utils.py
```diff
@@ -38,49 +87,127 @@ def map_over_axis(data, func, axis=0):
...
-    :type new_size: (int, int)
     :param scale_factors: Factors (fy,fx) by which to resize the image
-    :type scale_factors: (float, float)
     :param anti_alias: Use anti aliasing smoothing operation when downsampling. Default is True.
-    :type anti_alias: bool
     :param interpolation: Interpolation method used for resampling.
                           One of 'nearest', 'linear', 'cubic'. Default is 'linear'.
-    :type interpolation: string
     """
+    warnings.warn(
```
AleksMat, line 196: `resize_images` is also used by `CloudMaskTask`. So if we deprecate it we should start using `spatially_resize_image` there so that users won't be getting deprecation warnings. But then the problem will be that this one supports anti-aliasing while the new one doesn't. Hence the results of `CloudMaskTask` will change which we don't want.

I'm wondering if we shouldn't rather just join functionalities of `spatially_resize_image` and `resize_image` so that we don't have code duplications?

mask/eolearn/tests/test_utils.py
```diff
@@ -0,0 +1,79 @@
...
+    """Test that all methods and backends are able to downscale and upscale images of various dtypes."""
+    if library is ResizeLib.CV2:
+        if dtype == np.int32 and method is ResizeMethod.CUBIC or dtype == bool:
+            return
+
+    old_shape = (111, 111)
+    data_2d = np.arange(np.prod(old_shape)).astype(dtype).reshape(old_shape)
+    result = spatially_resize_image(data_2d, new_size, resize_method=method, resize_library=library)
+    assert result.shape == new_size
+
```
AleksMat, line 45: Additionally you could also check that dtype is preserved.

mask/eolearn/mask/utils.py
zigaLuksic: I recalled that we had some weird shenanigans with dtypes. Didn't feel like it's necessary since the intended use is to pass the dtype of the array, but you're right, its a simple improvement.

mask/eolearn/tests/test_utils.py
zigaLuksic: good call, added

mask/eolearn/mask/utils.py
zigaLuksic: added

mask/eolearn/mask/utils.py
zigaLuksic: I thought i forgot something...
Actually the new function does have anti-aliasing, as long as you use the PIL backend (and this is actually 3x faster than the old function). However the anti-aliasing is a bit different, so there will be some differences if I switch.

features/eolearn/features/utils.py
```diff
@@ -0,0 +1,152 @@
...
+
+    To specify the new scale use one of `new_size` or `scale_factors` parameters.
+
+    :param data: input image array
+    :param new_size: New size of the data (height, width)
+    :param scale_factors: Factors (f_height, f_width) by which to resize the image
+    :param spatial_axes: Which two axes of input data represent height and width. If left as `None` they are selected
+        according to standards of eo-learn features.
+    :param resize_method: Interpolation method used for resizing.
+    :param resize_library: Which Pyhon library to use for resizing. Default is PIL, as it supports all dtypes and
```
AleksMat, line 90: Typo "Pyhon"

(Caught by PyCharm spellchecker :wink:)

features/eolearn/tests/test_features_utils.py
```diff
@@ -0,0 +1,82 @@
...
+    example[0, 50:, 50:, 2] = 30
+    example[0, 50:, 50:, 3] = 40
+    return example.astype(np.float32)
+
+
+@pytest.mark.parametrize("method", ResizeMethod)
+@pytest.mark.parametrize("library", ResizeLib)
+@pytest.mark.parametrize("dtype", (np.float32, np.int32, bool))
+@pytest.mark.parametrize("new_size", [(50, 50), (35, 39), (271, 271)])
+def test_spatially_resize_image_new_size(method, library, dtype, new_size):
```
AleksMat, line 25: Usually we don't do this but would you mind setting types for input parameters in all test functions in this module? The problem is that PyCharm somehow doesn't understand types from decorators correctly and then marks a lot of code in these functions as "problematic".

AleksMat (commented):
Just 2 minor things, otherwise looks good.

features/eolearn/features/feature_manipulation.py
```diff
@@ -229,3 +232,40 @@ def map_method(self, feature: np.ndarray, slope: float, intercept: float, dtype:
...
+        resize_library: ResizeLib = ResizeLib.PIL,
+    ):
+        """
+        :param features: The specification of feature for which to perform resizing. Must be supported by the
+            :class:`FeatureParser<eolearn.core.utilities.FeatureParser>`. Features can be renamed, see `FeatureParser`
+            documentation.
+        :param new_size: New size of the data (height, width)
+        :param scale_factors: Factors (f_height, f_width) by which to resize the image
+        :param resize_method: Interpolation method used for resizing.
+        :param resize_library: Which Pyhon library to use for resizing. Default is PIL, as it supports all dtypes and
```
AleksMat, line 255: Same "Pyhon" typo, different place.

AleksMat (approved):

PR #420, opened 2022-06-10
Minor eoworkflow visualization update

AleksMat:
The main idea is that in case we have multiple nodes with the same name they get shown in report as "name_1", "name_2", ... instead of "name", "name_1", ... Only in case a node name is unique it gets shown as "name".

I also fixed some types and improved unit tests.


zigaLuksic (approved):
That was actually bothering me, great :D

PR #431, opened 2022-07-20
Filesystem upgrade and serialization

AleksMat:
The main use case for this is that we'll be able to first define a filesystem object with all parameters (e.g. credentials, ACL settings, ... ) and then pass it as an init parameter to IO tasks. This was solved by custom pickling and unpickling functions for filesystem objects.

Notes: 

- we still have to update IO tasks from `eolearn.io` subpackage. This will be done in next PRs.
- CI is still failing but only because of an unrelated `rasterio` issue.

### DIFF
--- a/core/eolearn/core/core_tasks.py
+++ b/core/eolearn/core/core_tasks.py
@@ -19,7 +19,7 @@ import numpy as np
 
 from .eodata import EOPatch
 from .eotask import EOTask
-from .utils.fs import get_filesystem
+from .utils.fs import get_filesystem, pickle_fs, unpickle_fs
 
 
 class CopyTask(EOTask):
@@ -54,7 +54,7 @@ class IOTask(EOTask, metaclass=ABCMeta):
         :param path: root path where all EOPatches are saved
         :type path: str
         :param filesystem: An existing filesystem object. If not given it will be initialized according to the EOPatch
-            path. If you intend to run this task in multiprocessing mode you shouldn't specify this parameter.
+            path.
         :type filesystem: fs.base.FS or None
         :param create: If the filesystem path doesn't exist this flag indicates to either create it or raise an error
         :type create: bool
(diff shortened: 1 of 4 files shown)


core/eolearn/core/core_tasks.py
```diff
@@ -63,19 +63,20 @@ def __init__(self, path, filesystem=None, create=False, config=None):
...
-        """A filesystem property that either initializes a new object or returns an existing one."""
-        if self._filesystem is None:
-            return get_filesystem(self.path, create=self._create, config=self.config)
+        """A filesystem property that unpickles an existing filesystem definition or creates a new one."""
+        if self._pickled_filesystem is None:
+            filesystem = get_filesystem(self.path, create=self._create_path, config=self.config)
+            self._pickled_filesystem = pickle_fs(filesystem)
 
-        return self._filesystem
+        return unpickle_fs(self._pickled_filesystem)
```
mlubej, line 79: Is unpickling the fs costly? if yes, then you could add `return filesystem` under the if-statement to reduce unnecessary unpickling operations.

core/eolearn/tests/test_utils/test_fs.py
```diff
@@ -66,6 +71,20 @@ def test_s3_filesystem(aws_session_token):
         assert filesystem.aws_session_token == aws_session_token
 
 
+@pytest.mark.parametrize("s3fs_function", [get_filesystem, load_s3_filesystem])
+def test_s3fs_keyword_arguments(s3fs_function):
+    filesystem = s3fs_function("s3://dummy-bucket/", acl="bucket-owner-full-control")
```
mlubej, line 76: Interesting, I didn't know the S3FS already provides an `acl` argument. Perhaps you could also add a test where the `acl` is set via the `upload_args`? Or is that outside of this scope and is the responsibility of S3FS?

core/eolearn/core/utils/fs.py
```diff
@@ -118,6 +126,58 @@ def get_aws_credentials(aws_profile: str, config: Optional[SHConfig] = None) ->
     return config
 
 
+def pickle_fs(filesystem: FS) -> bytes:
+    """By default, filesystem objects cannot be pickled because they contain thread locks and similar unserializable
+    objects. This function removes the before pickling the filesystem object.
```
mlubej, line 131: Probably something wrong in the sentence - weird grammar.

mlubej (commented):
I managed to review what I could, but I'm not so familiar with the details of the FS (un)pickling, so it's possible that I couldn't give a thorough review.

core/eolearn/tests/test_utils/test_fs.py
AleksMat: Actually, from the [`S3FS` implementation](https://github.com/PyFilesystem/s3fs/blob/v1.1.1/fs_s3fs/_s3fs.py#L302) you can see it is the other way around. Parameter `acl` is just a helper parameter and its value gets put into `upload_args` dictionary. This unit tests checks this.

core/eolearn/core/utils/fs.py
AleksMat: Thanks, it should be better now.

core/eolearn/core/core_tasks.py
AleksMat: My benchmarks show `6 µs` which is nothing. But I still decided to add the return statement.

mlubej (approved):
No more comments from my side.

PR #435, opened 2022-07-24
Updated logs_handler_factory parameter to support filesystem changes

AleksMat:
This should be the last PR in the series of filesystem-handling changes. It is about passing a filesystem object to log handlers of each workflow execution.

### DIFF
--- a/core/eolearn/core/eoexecution.py
+++ b/core/eolearn/core/eoexecution.py
@@ -20,7 +20,7 @@ import logging
 import threading
 import warnings
 from dataclasses import dataclass
-from logging import Filter, Handler, Logger
+from logging import FileHandler, Filter, Handler, Logger
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple
 
 import fs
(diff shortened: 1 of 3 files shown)


zigaLuksic (approved):

core/eolearn/core/eoexecution.py
```diff
@@ -16,25 +16,34 @@
...
 from .eonode import EONode
 from .eoworkflow import EOWorkflow, WorkflowResults
 from .exceptions import EORuntimeWarning
-from .utils.fs import get_base_filesystem_and_path, get_full_path
+from .utils.fs import get_base_filesystem_and_path, get_full_path, pickle_fs, unpickle_fs
 from .utils.logging import LogFileFilter
 from .utils.parallelize import _decide_processing_type, _ProcessingType, parallelize
 
+
+class _FilesystemHandlerFactoryType(Protocol):
```
zigaLuksic, line 38: my brain reads this as **filesystem-handler** factory type
perhaps `_HandlerWithFsFactoryType`? I would even be inclined to drop the `Type` suffix, since I dont see it used that often in python.

core/eolearn/core/eoexecution.py
```diff
@@ -93,8 +103,14 @@ def __init__(
         :param filesystem: A filesystem object for saving logs and a report.
         :param logs_filter: An instance of a custom filter object that will filter certain logs from being written into
             logs. It works only if save_logs parameter is set to True.
-        :param logs_handler_factory: A callable class or function that takes logging path as its only input parameter
-            and creates an instance of logging handler object
+        :param logs_handler_factory: A callable class or function that initializes an instance of a logging `Handler`
+            object. Its signature should support one of the following options:
+
+            - A single parameter accepting a full path to the log file.
```
zigaLuksic, line 109:
```suggestion
            - A single parameter describing a full path to the log file.
```

zigaLuksic (approved):

core/eolearn/core/eoexecution.py
AleksMat: Ok, renamed to `_HandlerWithFsFactoryType`. I would keep `Type` suffix otherwise it can be confused with normal classes.

core/eolearn/core/eoexecution.py
AleksMat: Changed :+1:

PR #436, opened 2022-07-26
Extra S3 filesystem pickling test

AleksMat:
Changes:
- Added a test that makes sure `S3FS` is fully usable after being unpickled.
- Reduced code duplication by using a package `pytest-lazy-fixture`. If this works well I think we have a few more cases where such solution would be great.

### DIFF
--- a/core/eolearn/tests/test_utils/test_fs.py
+++ b/core/eolearn/tests/test_utils/test_fs.py
@@ -13,7 +13,6 @@ from _thread import RLock
 from pathlib import Path
 from typing import List
 
-import boto3
 import pytest
 from botocore.credentials import Credentials
 from fs.base import FS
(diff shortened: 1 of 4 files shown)


core/eolearn/tests/test_utils/test_fs.py
```diff
@@ -135,6 +136,30 @@ def test_tempfs_serialization():
     assert not unpickled_filesystem.exists("/")
 
 
+@mock_s3
```
zigaLuksic, line 139: is this decorator needed despite it being used in the fixture?

AleksMat: Good point but it turns out it is. If I remove `@mock_s3` for any test it fails saying that the bucket doesn't exist. But I don't have enough understanding how `moto` works to be able to explain why this is so.

PR #438, opened 2022-07-27
Updated SimpleFilterTask and wrote additional unit tests

AleksMat:
There are a few minor changes in the functionality:

- In case all timestamps are filtered the old version raised an error but the new version returns empty numpy arrays (i.e. shape (0, height, width, bands)) and empty GeoDataFrames.
- The new version returns a shallow copy of an input EOPatch. That is because otherwise we would have a mismatch in time dimension.
- I removed `_update_other_data` method because it was useless. Instead, any such update can be performed in a separate task.
- The new version now also supports filtering vector features.

Note: I'm not sure if I correctly handled type of `filter_func` but it is a bit tricky because it depends on feature type of `feature` parameter.

### DIFF
--- a/features/eolearn/features/feature_manipulation.py
+++ b/features/eolearn/features/feature_manipulation.py
@@ -15,15 +15,18 @@ file in the root directory of this source tree.
 import datetime as dt
 import logging
 from functools import partial
-from typing import Any, Optional, Tuple, Union
+from typing import Any, Callable, Iterable, List, Optional, Tuple, TypeVar, Union
 
 import numpy as np
+from geopandas import GeoDataFrame
 
-from eolearn.core import EOPatch, EOTask, FeatureType, MapFeatureTask
+from eolearn.core import EOPatch, EOTask, FeatureType, FeatureTypeSet, MapFeatureTask
 
 from .utils import ResizeLib, ResizeMethod, spatially_resize_image
 
 LOGGER = logging.getLogger(__name__)
+
+_T = TypeVar("_T", np.ndarray, dt.datetime)
 
 
 class SimpleFilterTask(EOTask):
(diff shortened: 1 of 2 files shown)


features/eolearn/features/feature_manipulation.py
```diff
@@ -34,51 +37,55 @@ class SimpleFilterTask(EOTask):
     A filter_func is a callable which takes a numpy array and returns a bool.
     """
 
-    def __init__(self, feature, filter_func, filter_features=...):
+    def __init__(self, feature, filter_func: Callable[[_T], bool], filter_features=...):
         """
         :param feature: Feature in the EOPatch , e.g. feature=(FeatureType.DATA, 'bands')
-        :type feature: (FeatureType, str)
+        :type feature: (FeatureType, str) or FeatureType.TIMESTAMP
```
zigaLuksic, line 43: why not add this to signature as well? then please also remove types from docstrings so we dont confuse docbuilders

zigaLuksic: i know that the last argument cannot be typed properly yet, just slap `Any` on it for now i guess...

features/eolearn/features/feature_manipulation.py
```diff
@@ -34,51 +37,55 @@ class SimpleFilterTask(EOTask):
...
+        for feature in self.filter_features_parser.get_features(eopatch):
+            feature_type, _ = feature
+            data = eopatch[feature]
+
+            if feature_type is FeatureType.TIMESTAMP:
+                data = [data[idx] for idx in good_idxs]
+
+            elif feature_type.is_temporal():
+                if feature_type.is_raster():
+                    data = data[good_idxs, ...]
```
zigaLuksic, line 82: are the `...` necessary?

zigaLuksic (commented):
I am not sure i understand the 'returns a shallow copy'
what exactly is shallowly copied here?

features/eolearn/features/feature_manipulation.py
AleksMat: I think not because by default filtering should already happen on the 1st axis. But I find it more explicit to show that we filtering on the 1st axis.

features/eolearn/features/feature_manipulation.py
AleksMat: Added types. Btw, for Sphinx docs types in docstrings take priority over type hints so it would just take those.

features/eolearn/features/feature_manipulation.py
AleksMat: Removed `...`

AleksMat:
> I am not sure i understand the 'returns a shallow copy'
> what exactly is shallowly copied here?

We return a new instance of an EOPatch and by default all features are filtered and moved into it. So it's basically a shallow copy.

zigaLuksic:
> > I am not sure i understand the 'returns a shallow copy'
> > what exactly is shallowly copied here?
> 
> We return a new instance of an EOPatch and by default all features are filtered and moved into it. So it's basically a shallow copy.

Not entirely sure I would call it a copy, it's more of a **view** (since we are returning views of numpy features), but I see your reasoning now.

features/eolearn/features/feature_manipulation.py
```diff
@@ -29,27 +29,29 @@
 _T = TypeVar("_T", np.ndarray, dt.datetime)
 
 
-class SimpleFilterTask(EOTask):
+class SimpleFilterTask(EOTask, Generic[_T]):
```
zigaLuksic, line 32: wouldnt it be possible to omit all the generic stuff by using 
```python
FilterFuncType = Union[Callable[[np.ndarray], bool], Callable[[dt.datetime], bool]]
```
While i'm a big fan of using advanced type theory, I'm worried that this migh class with future improvements to EOTasks since they would need to be generics (and i'm not sure how generics clash in python).

AleksMat: Ah, too bad. Ok, I removed `Generic` and `TypeVar` and used the normal union. Then I had to set a type `Iterable` to the parameter in `_get_filtered_indices` method and mypy is also fine with that.

PR #441, opened 2022-08-01
separate FeatureIO into subclasses

zigaLuksic:
FeatureIO seemed to be a 'swiss army knife' class, so I extracted the common behaviour into a generic class that is then specialized in subclasses. This also greatly improves type information. Sadly we currently cannot 100% guarantee type safety for JSON-like data (so timestamp and metainfo) and geodataframes (due to deprecated pickle functionality). I am considering further specializing json-like featureIO into specific ones for timestamps and metainfo, but it might cause some code duplication, since both are handled very similarly. Open to ideas.

### DIFF
--- a/core/eolearn/core/eodata_io.py
+++ b/core/eolearn/core/eodata_io.py
@@ -10,11 +10,14 @@ This source code is licensed under the MIT license found in the LICENSE
 file in the root directory of this source tree.
 """
 import concurrent.futures
+import datetime
 import gzip
 import json
 import pickle
 import warnings
+from abc import ABCMeta, abstractmethod
 from collections import defaultdict
+from typing import Any, BinaryIO, Generic, Iterable, Optional, TypeVar, Union
 
 import fs
 import fs.move
@@ -26,14 +29,16 @@ from fs.osfs import OSFS
 from fs.tempfs import TempFS
 from geopandas import GeoDataFrame, GeoSeries
 
-from sentinelhub import CRS, Geometry, MimeType
+from sentinelhub import CRS, BBox, Geometry, MimeType
 from sentinelhub.exceptions import SHUserWarning
 from sentinelhub.os_utils import sys_is_windows
 
-from .constants import FeatureType, OverwritePermission
+from .constants import FeatureType, FeatureTypeSet, OverwritePermission
 from .exceptions import EODeprecationWarning
 from .utils.parsing import FeatureParser
 from .utils.vector_io import infer_schema
+
+_T = TypeVar("_T")
 
 
 def save_eopatch(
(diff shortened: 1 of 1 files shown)


core/eolearn/core/eodata_io.py
```diff
@@ -106,7 +111,7 @@ def remove_redundant_files(filesystem, eopatch_features, filesystem_features, cu
 def load_eopatch(eopatch, filesystem, patch_location, features=..., lazy_loading=False):
     """A utility function used by `EOPatch.load` method."""
     features = list(walk_filesystem(filesystem, patch_location, features))
-    loading_data = [FeatureIO(ftype, path, filesystem) for ftype, _, path in features]
+    loading_data: Iterable[Any] = [create_feature_io(ftype, path, filesystem) for ftype, _, path in features]
```
AleksMat, line 114: Couldn't a type here be `List[FeatureIO]` and wouldn't this be automatically implied?

core/eolearn/core/eodata_io.py
```diff
@@ -262,83 +270,69 @@ def load(self):
...
-
-        raise ValueError(f"Unsupported file format {file_format} for feature type {self.feature_type}.")
-
-    def _decode(self, file, path):
-        """Loads from a file and decodes content."""
-        file_format = MimeType(fs.path.splitext(path)[1].strip("."))
 
-        if file_format is MimeType.NPY:
-            return np.load(file)
+class FeatureIOnumpy(FeatureIO[np.ndarray]):
```
AleksMat, line 317: Perhaps the name could be `FeatureIONumpy` to better follow naming conventions? (The other `3` classes could also be renamed accordingly.)

core/eolearn/core/eodata_io.py
```diff
@@ -382,4 +368,74 @@ def _decode(self, file, path):
...
+
+
+def jsonify_timestamp(param: object) -> str:
+    """Adds the option to serialize datetime.date objects via isoformat."""
+    if isinstance(param, datetime.date):
+        return param.isoformat()
+    raise TypeError(f"Object of type {type(param)} is not yet supported in jsonify utility function")
+
+
+def create_feature_io(ftype: FeatureType, path: str, filesystem: FS) -> FeatureIO:
```
AleksMat, line 433: Maybe `jsonify_timestamp` and `create_feature_io` could start with an underscore so that we know they are private for this module?

core/eolearn/core/eodata_io.py
zigaLuksic: sadly no, because a few lines down we overwrite it with the values inside FeatureIOs (unless the load is lazy), so we cannot be more precise than `Any`. And since we do it via `map` it cant even be `List`.

core/eolearn/core/eodata_io.py
zigaLuksic: done, please check if further corrections are desired

core/eolearn/core/eodata_io.py
zigaLuksic: good call, undrescore added

core/eolearn/core/eodata_io.py
AleksMat: The only remaining one is `FeatureIOgeodf`. I would rename it to `FeatureIOGeoDf`.

zigaLuksic: I could've sworn i've renamed it. Guess I didnt confirm the renaming... Done now

core/eolearn/core/eodata_io.py
zigaLuksic: Well the problem is because this is later iterated over and fed into the EOPatch so they need to be the same variable. If you look at the whole block you'll see that it's a bit of a hassle. I guess we could do the loading in the middle of assigning values, but that was not part of this update, so I didn't refactor to that extent.

AleksMat: I see now :facepalm: 
Ok, let's leave it like this.

PR #443, opened 2022-08-17
Typed feature dicts

zigaLuksic:
Similar to `FeatureIO` the only proper way to bring some notion of types to `EOPatches` begins at making a generic `_FeatureDict` which is then properly sub-classed. Now `eopatch.data["name"]` is recognized as a numpy array. Can't do much about `eopatch[FeatureType.DATA, "name"]` without a massive amount of overloads.

### DIFF
--- a/core/eolearn/core/eodata.py
+++ b/core/eolearn/core/eodata.py
@@ -14,9 +14,10 @@ file in the root directory of this source tree.
 from __future__ import annotations
 
 import copy
-import datetime
+import datetime as dt
 import logging
-from typing import TYPE_CHECKING, List, Optional, Tuple, Union
+from abc import ABCMeta, abstractmethod
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Literal, Optional, Tuple, Type, TypeVar, Union, overload
 
 import attr
 import dateutil.parser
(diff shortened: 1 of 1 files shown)


core/eolearn/core/eodata.py
```diff
@@ -44,6 +54,169 @@
...
+    """_FeatureDict object specialized for GeoDataFrames."""
+
+    def __init__(self, feature_dict: Dict[str, gpd.GeoDataFrame], feature_type: FeatureType):
+        if not feature_type.is_vector:
+            raise ValueError(f"Feature type {feature_type} does not represent a vector feature.")
+        super().__init__(feature_dict, feature_type)
+
+    def _parse_feature_value(self, value: object, feature_name: str) -> gpd.GeoDataFrame:
+        if isinstance(value, gpd.GeoSeries):
+            value = gpd.GeoDataFrame(dict(geometry=value), crs=value.crs)
```
AleksMat, line 187: It turns out using `dict` here is unnecessary, we can simply do `gpd.GeoDataFrame(geometry=value, crs=value.crs)`. Sorry, this is my old mistake :see_no_evil:.

core/eolearn/core/eodata.py
```diff
@@ -44,6 +54,169 @@
...
+    def __setitem__(self, feature_name: str, value: Union[_T, FeatureIO[_T]]) -> None:
+        """Before setting value to the dictionary it checks that value is of correct type and dimension and tries to
+        transform value in correct form.
+        """
+        if not isinstance(value, FeatureIO):
+            value = self._parse_feature_value(value, feature_name)
+        self._check_feature_name(feature_name)
+        super().__setitem__(feature_name, value)
+
+    def _check_feature_name(self, feature_name: str) -> None:
```
AleksMat, line 96: Maybe a short docstring could be added.

AleksMat (commented):
Really nice improvement. I added just a couple of minor comments.

PR #445, opened 2022-08-22
Re-adds pickle reading to other data types

zigaLuksic:
In the refactoring of FeatureIO we lost the ability to read pickle files. This fixes it (for a while, since it will be removed soon).

### DIFF
--- a/core/eolearn/core/eodata_io.py
+++ b/core/eolearn/core/eodata_io.py
@@ -17,7 +17,8 @@ import pickle
 import warnings
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


MustafaAlahmid:
I have a dataset that have bbox, metadata, and timestamp all in pickle format 
the EOPatch is not able to read them and show that the format should be .geojson 

is there any version that i can use to read this data or to convert them into .geojson format?
thanks

zigaLuksic:
Hi @MustafaAlahmid , something similar was discussed in https://github.com/sentinel-hub/eo-learn/issues/582, maybe this will solve your problem.

PR #447, opened 2022-08-30
fix: meteoblue y-axis orientation for MeteoblueRasterTask

fred-sch:
MeteoblueRasterTask was previously using northward facing y-axis in contrast to data coming from sentinelhub


batic:
Hi @fred-sch, 

Could you please edit the pull request to merge to `develop` branch? 

Thanks in advance!

zigaLuksic:
Hi @fred-sch, thanks for your contribution.
Currently one of the tests fails with the fix, could you fix the test as well?
Also the comment could probably fit into two lines (while respecting the 120 character line width)

fred-sch:
Hey @zigaLuksic, 
sorry I've somehow overseen this test. I now adapted it to check the last element in the flipped y-axis is the same value as in the original test. 
Unfortunately, the four line comment is not fitting in two lines while respecting the 120 characters line width. Therefore, I just left it as is because I did not want to shorten the comment for comprehensibility.

PR #448, opened 2022-08-30
Fix image in core example notebook

zigaLuksic:
The process now explicitly copies the image for the core overview notebook.

Read-the-docs started crashing due to conda being too slow (over 300 seconds to build the env :scream_cat: ) so as suggested by ReadTheDocs I switched over to using Mamba. This moves the python version of docbuilding from 3.7 to a special mamba 3.10, but it now takes 200s to build the env and the docs seem the same.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -243,9 +243,9 @@ def copy_documentation_examples(source_folder, target_folder):
             if line.startswith("examples/"):
                 files_to_include.append(line.split("/", 1)[1])
 
-    for notebook in files_to_include:
-        source_path = os.path.join(source_folder, notebook)
-        target_path = os.path.join(target_folder, notebook)
+    for file in files_to_include:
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -241,9 +241,9 @@ def copy_documentation_examples(source_folder, target_folder):
         for line in content.split("\n"):
             line = line.strip(" \t")
             if line.startswith("examples/"):
-                notebooks_to_include.append(line.split("/", 1)[1])
+                files_to_include.append(line.split("/", 1)[1])
 
-    for notebook in notebooks_to_include:
+    for notebook in files_to_include:
```
batic, line 246: Perhaps just rename to `for file in files_to_include`, as they are not strictly just notebooks anymore.

zigaLuksic: done

PR #452, opened 2022-09-07
Feat/s3fs strict

batic:
As we are having some issues in eo-grow when dealing with S3FS, this PR ensures that the S3FS created in eo-learn is not `strict`.

Additionally, updated .gitignore with ".DS_Store".

### DIFF
--- a/core/eolearn/tests/test_utils/test_fs.py
+++ b/core/eolearn/tests/test_utils/test_fs.py
@@ -57,7 +57,7 @@ def test_s3_filesystem(aws_session_token):
     filesystem = get_filesystem(s3_url)
     assert isinstance(filesystem, S3FS)
     assert filesystem.dir_path == folder_name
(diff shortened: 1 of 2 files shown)


nkadunc (approved):

core/eolearn/tests/test_utils/test_fs.py
```diff
@@ -57,6 +57,7 @@ def test_s3_filesystem(aws_session_token):
     filesystem = get_filesystem(s3_url)
     assert isinstance(filesystem, S3FS)
     assert filesystem.dir_path == folder_name
+    assert not filesystem.strict, "By default, resulting S3FS is not strict."
```
zigaLuksic, line 60:
```suggestion
    assert not filesystem.strict, "Other use-cases assume that the returned S3FS is non-strict by default."
```

PR #453, opened 2022-09-07
Adds an option to pass resolution to the SpatialResizeTask.

veseln:
Also refactored the way the parameters are passed to the task, breaking the existing API.

### DIFF
--- a/features/eolearn/features/feature_manipulation.py
+++ b/features/eolearn/features/feature_manipulation.py
@@ -265,13 +265,11 @@ class SpatialResizeTask(EOTask):
         resize_type, resize_type_param = resize_parameters
         resize_type = ResizeParam(resize_type)
 
-        if resize_type == ResizeParam.NEW_SIZE:
-            return {ResizeParam.NEW_SIZE.value: resize_type_param}
-        if resize_type == ResizeParam.SCALE_FACTORS:
-            return {ResizeParam.SCALE_FACTORS.value: resize_type_param}
         if resize_type == ResizeParam.RESOLUTION:
             new_size = bbox_to_dimensions(bbox, resize_type_param)
             return {ResizeParam.NEW_SIZE.value: new_size}
+
+        return {resize_type.value: resize_type_param}
 
     def __init__(
         self,
--- a/features/eolearn/features/feature_manipulation.py
+++ b/features/eolearn/features/feature_manipulation.py
@@ -15,14 +15,16 @@ file in the root directory of this source tree.
 import datetime as dt
 import logging
 from functools import partial
-from typing import Any, Callable, Iterable, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, List, Tuple, Union
 
 import numpy as np
 from geopandas import GeoDataFrame
 
+from sentinelhub import BBox, bbox_to_dimensions
+
 from eolearn.core import EOPatch, EOTask, FeatureType, FeatureTypeSet, MapFeatureTask
 
-from .utils import ResizeLib, ResizeMethod, spatially_resize_image
+from .utils import ResizeLib, ResizeMethod, ResizeParam, spatially_resize_image
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


features/eolearn/features/utils.py
```diff
@@ -22,6 +22,14 @@
 )
 
 
+class ResizeParam(Enum):
+    """Possible parameters (options) for passing the new dimensions to the spatial resizing of data."""
```
zigaLuksic, line 26:
```suggestion
    """Descriptors of spatial-resizing parameter options."""
```

features/eolearn/tests/test_feature_manipulation.py
```diff
@@ -232,14 +233,38 @@ def test_linear_function_task():
 @pytest.mark.filterwarnings("ignore::RuntimeWarning")
 def test_spatial_resize_task(example_eopatch):
```
zigaLuksic, line 234: the test is getting kinda long, would it be possible to parametrize it?
So to figure out how to get an eopatch of twice the size and half the size (for each of the input modes) and then write all the possible parameters expressing that. The test for faulty parameters is then a separate one. Also test that strings are correctly parsed to enums (so a test case should use `"resolution"`) if we are keen on keeping that shortcut for users.

features/eolearn/features/feature_manipulation.py
```diff
@@ -257,11 +259,23 @@ def map_method(self, feature: np.ndarray, slope: float, intercept: float) -> np.
...
+            new_size = bbox_to_dimensions(bbox, resize_type_param)
+            return {ResizeParam.NEW_SIZE.value: new_size}
+
+        return {resize_type.value: resize_type_param}
+
     def __init__(
         self,
+        *,
+        resize_parameters: Tuple[ResizeParam, Tuple[float, float]],
         features: Any = ...,
```
zigaLuksic, line 278: if we don't want to lock arguments to keyword-only (but keep `features=...`) the value `(ResizeParam.SCALE_FACTORS, (1, 1))` is technically a viable default. Not saying it's a good option, but maybe it's worth to be considered.

features/eolearn/tests/test_feature_manipulation.py
jgersak: done

features/eolearn/features/feature_manipulation.py
zigaLuksic: agreed

features/eolearn/tests/test_feature_manipulation.py
```diff
@@ -230,41 +230,30 @@ def test_linear_function_task():
...
 
+@pytest.mark.parametrize(
+    ["resize_parameters", "features_call", "features_check", "outputs"],
+    [
+        ((ResizeParam.NEW_SIZE, (50, 70)), ("data", "CLP"), ("data", "CLP"), (68, 50, 70, 1)),
+        ((ResizeParam.NEW_SIZE, (50, 70)), ("data", "CLP"), ("mask", "CLM"), (68, 101, 100, 1)),
+        ((ResizeParam.NEW_SIZE, (50, 70)), ..., ("data", "CLP"), (68, 50, 70, 1)),
+        ((ResizeParam.NEW_SIZE, (50, 70)), ..., ("mask", "CLM"), (68, 50, 70, 1)),
+        ((ResizeParam.NEW_SIZE, (50, 70)), ("data", "CLP", "CLP_small"), ("data", "CLP_small"), (68, 50, 70, 1)),
+        ((ResizeParam.NEW_SIZE, (50, 70)), ("data", "CLP", "CLP_small"), ("data", "CLP"), (68, 101, 100, 1)),
```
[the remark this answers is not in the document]
veseln: I am not sure I understand your comment. The previous tests seemed to check two things: 

1) That the feature you were calling got resized
2) That some other feature did not get resized 

This is parameterized in such a way to allow for this.  

There is then a separate test for the invalid inputs.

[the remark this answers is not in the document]
zigaLuksic: I was thinking that the part 2. should perhaps not be in this parametrization, but rather in the other test (or a completely new, third test). But perhaps we can refactor that once we venture onto the journey of microtests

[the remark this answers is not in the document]
veseln: To be fair, I would leave it as is for now and leave it for future wholesale refactoring of tests. I understand that checking that something has changed and something hasn't changed could be decoupled, but on the other hand I don't see it hurting that  much if left as is.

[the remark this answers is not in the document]
zigaLuksic: agreed

PR #456, opened 2022-09-07
Feat/resampling

nkadunc:
The SH tasks now expose the upsampling and downsampling parameters, which previously needed to be passed in aux_parameters

### DIFF
--- a/io/eolearn/io/sentinelhub_process.py
+++ b/io/eolearn/io/sentinelhub_process.py
@@ -12,7 +12,7 @@ file in the root directory of this source tree.
 
 import datetime as dt
 import logging
-from typing import Callable, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


zigaLuksic:
Please add pylint ignore comments where appropriate, so that all the tests complete.

devisperessutti:
> Please add pylint ignore comments where appropriate, so that all the tests complete.

Done. Any better way of dealing with these issues otherwise?

zigaLuksic:
> > Please add pylint ignore comments where appropriate, so that all the tests complete.
> 
> Done. Any better way of dealing with these issues otherwise?

Apart from refactoring the task (which will need to happen eventually but will be a very tricky business due to code-breaking) I'm afraid not. In the past we just increased the values in the pylint configuration but that is even worse

jgersak (approved):

PR #457, opened 2022-09-07
Feat/exploding bands task

jgersak:
test commit


zigaLuksic:
Please re-base the MR to the develop branch so that we don't merge commits pertaining to the eopatch plotting. It might be that the easiest way to achieve that is to close the MR and open a new one

PR #458, opened 2022-09-08
Feat/exploding task

jgersak:
new MR

### DIFF
--- a/core/eolearn/core/core_tasks.py
+++ b/core/eolearn/core/core_tasks.py
@@ -18,7 +18,7 @@ from typing import Iterable, List, Tuple
 import fs
 import numpy as np
 
-from . import FeatureType
+from .constants import FeatureType
 from .eodata import EOPatch
 from .eotask import EOTask
 from .utils.fs import get_filesystem, pickle_fs, unpickle_fs
--- a/core/eolearn/core/core_tasks.py
+++ b/core/eolearn/core/core_tasks.py
@@ -576,7 +576,6 @@ class ExplodeBandsTask(EOTask):
         """
         self.input_feature = input_feature
         self.output_features = output_features
-
         self.bands = bands
 
     def execute(self, eopatch):
--- a/core/eolearn/core/core_tasks.py
+++ b/core/eolearn/core/core_tasks.py
@@ -13,7 +13,7 @@ file in the root directory of this source tree.
 """
 import copy
 from abc import ABCMeta, abstractmethod
-from typing import Dict, List, Tuple
+from typing import Dict, Iterable, Tuple
 
 import fs
 import numpy as np
@@ -563,7 +563,7 @@ class ExplodeBandsTask(EOTask):
     def __init__(
         self,
         input_feature: Tuple[FeatureType, str],
-        output_mapping: Dict[Tuple[FeatureType, str], List[int]],
+        output_mapping: Dict[Tuple[FeatureType, str], Iterable[int]],
     ):
         """
         :param input_feature: A source feature from which to take the subset of bands.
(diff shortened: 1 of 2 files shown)


core/eolearn/core/core_tasks.py
```diff
@@ -555,6 +557,32 @@ def map_method(self, feature):
...
 
 
+class ExplodeBandsTask(EOTask):
+    """Explode a subset of bands from one feature to multiple new features."""
+
+    def __init__(
+        self,
+        input_feature: Tuple[FeatureType, str],
+        output_features: Iterable[Tuple[FeatureType, str]],
+        bands: Iterable[List[int]],
```
zigaLuksic, line 567: this way of specifying input is kind of problematic. What if the two iterables don't match? What if the user forgets an element (and the rest are off by one) etc.

I think it would make more sense to have a mapping (via dict or similar) of band indices to output features.

core/eolearn/tests/test_core_tasks.py
```diff
@@ -387,6 +388,29 @@ def test_map_features(test_eopatch):
...
+    bands = [[0], [1], [1, 2]]
+    move_bands = ExplodeBandsTask(
+        (FeatureType.DATA, "REFERENCE_SCENES"),
+        [(FeatureType.DATA, "B01"), (FeatureType.DATA, "B02"), (FeatureType.DATA, "B02 & B03")],
+        bands,
+    )
+    patch = move_bands(test_eopatch)
+    assert "B01" in patch.data
+    assert "B02" in patch.data
+    assert "B02 & B03" in patch.data
```
zigaLuksic, line 411: you should also verify that the correct data was extracted

core/eolearn/core/core_tasks.py
zigaLuksic, line 567: another notice here, what do you do if the list of bands to extract is empty? do we delegate handling to `ExtractBandTask`?

zigaLuksic: you can switch to using tuples for indices

zigaLuksic: but having it the other way around is also completely fine (names to indices)

veseln: We'll have it names to indices.

core/eolearn/core/core_tasks.py
```diff
@@ -555,6 +557,29 @@ def map_method(self, feature):
         return feature[..., self.bands]
 
 
+class ExplodeBandsTask(EOTask):
+    """Explode a subset of bands from one feature to multiple new features."""
+
+    def __init__(
+        self,
+        input_feature: Tuple[FeatureType, str],
+        output_mapping: Dict[Tuple[FeatureType, str], Iterable[int]],
```
zigaLuksic, line 566: as mentioned before, it would be nice to have the options of `Dict[Tuple[FeatureType, str], Union[int, Iterable[int]]]`, so that one can do `{"B04": 3, "B03": 2}` without the pesky brackets

core/eolearn/tests/test_core_tasks.py
```diff
@@ -387,6 +390,31 @@ def test_map_features(test_eopatch):
...
+    ],
+)
+def test_explode_bands(
+    test_eopatch: EOPatch,
+    feature: FeatureType,
+    task_input: Dict[Tuple[FeatureType, str], Iterable[int]],
+):
+    move_bands = ExplodeBandsTask(feature, task_input)
+    patch = move_bands(test_eopatch)
+    assert all(new_feature in patch for new_feature in task_input.keys())
```
zigaLuksic, line 412: no need to use `.keys()` when using iteration over lists (keys are iterated over by default)

PR #459, opened 2022-09-08
Feat/no data value

tsuligoj:
Default value of the parameter `no_data_value`  was set to `None`. In `ExportToTiffTask` this is the NoData value of pixels in exported images. In `ImportFromTiffTask` this value is assigned to the pixels with NoData.

Added test for different `no_data_values`  input to both `ExportToTiffTask` and `ImportFromTiffTask` .

### DIFF
--- a/io/eolearn/tests/test_raster_io.py
+++ b/io/eolearn/tests/test_raster_io.py
@@ -26,7 +26,7 @@ from fs.errors import ResourceNotFound
 from moto import mock_s3
 from numpy.testing import assert_array_equal
 
-from sentinelhub import read_data
+from sentinelhub import CRS, BBox, read_data
 from sentinelhub.time_utils import serialize_time
 
 from eolearn.core import EOPatch, FeatureType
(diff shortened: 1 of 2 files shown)


io/eolearn/tests/test_raster_io.py
```diff
@@ -360,3 +360,33 @@ def test_time_dependent_feature_with_timestamps(test_eopatch):
     new_eopatch = import_task(test_eopatch, filename=filename)
 
     assert_array_equal(new_eopatch[feature], test_eopatch[feature])
+
+
+@pytest.mark.parametrize(
+    "no_data_value, data_type", [(None, np.uint16), (np.nan, float), (0, int), (None, float), (1, np.byte)]
+)
+def test_export_import_sequence(no_data_value, data_type):
```
zigaLuksic, line 368: a quick docstring describing the tests would be nice. Doesn't have to be much, just to guide whoever tries to adjust the tests.

io/eolearn/tests/test_raster_io.py
```diff
@@ -360,3 +360,33 @@ def test_time_dependent_feature_with_timestamps(test_eopatch):
...
+        filename = "test_seq.tiff"
+        file_path = os.path.join(tmp_dir_name, filename)
+        export_task = ExportToTiffTask(
+            feature=feature, folder=tmp_dir_name, band_indices=[0], no_data_value=no_data_value
+        )
+        export_task.execute(eopatch=eopatch, filename=filename)
+
+        with rasterio.open(file_path) as src:
+            tif_array = src.read(masked=True)
+            assert np.sum(tif_array.mask) == np.sum(np_arr == no_data_value)
```
zigaLuksic, line 387: could we just compare `tif_array.mask` with `np_arr == no_data_value`?

io/eolearn/tests/test_raster_io.py
```diff
@@ -360,3 +360,33 @@ def test_time_dependent_feature_with_timestamps(test_eopatch):
...
+@pytest.mark.parametrize(
+    "no_data_value, data_type", [(None, np.uint16), (np.nan, float), (0, int), (None, float), (1, np.byte)]
+)
+def test_export_import_sequence(no_data_value, data_type):
+    eopatch = EOPatch()
+    eopatch.bbox = BBox((0, 0, 1, 1), crs=CRS.WGS84)
+    feature = (FeatureType.DATA_TIMELESS, "DATA")
+
+    np_arr = np.zeros((10, 10, 1), dtype=data_type)
+    np_arr[:5, :5, :] = 1
```
zigaLuksic, line 374: should this always be set to 1? Do the `np.nan` even play a role in the tests then?

io/eolearn/tests/test_raster_io.py
zigaLuksic: if you add a comment describing the difference in dimensions then i think the `squeeze` approach is fine. Another option is to use `tif_array[0]` and `np_arr[..., 0]` (or whichever way it has to be), but perhaps squeeze is better here

io/eolearn/tests/test_raster_io.py
```diff
@@ -383,8 +383,15 @@ def test_export_import_sequence(no_data_value, data_type):
...
         with rasterio.open(file_path) as src:
-            tif_array = src.read(masked=True)
-            assert np.sum(tif_array.mask) == np.sum(np_arr == no_data_value)
+            # when reading, move axis to have the tif_array in the EOPatch.data_timeless dimension structure
+            tif_array = np.moveaxis(src.read(masked=True), 0, -1)
+
+            if no_data_value is not np.nan:
+                no_data_arr = np_arr == no_data_value
+            else:
+                no_data_arr = np.isnan(np_arr)
```
zigaLuksic, line 392: would `no_data_arr = np.isnan(np_arr) if no_data_value is np.nan else np_arr == no_data_value` also work?

zigaLuksic: Then I would just change the order of the branches, so that the condition would be `no_data_value is np.nan`.

zigaLuksic (approved):

PR #461, opened 2022-09-08
Add geom param to sh input and evalscript task

devisperessutti:
This PR exposes a `geometry` argument that can be passed to the `SentinelHubInputTask` and `SentinelHubEvalscriptTask`, which effectively allows to download only pixels within the specified geometry.

At the moment, `bbox` needs to be specified as well.

### DIFF
--- a/io/eolearn/io/sentinelhub_process.py
+++ b/io/eolearn/io/sentinelhub_process.py
@@ -572,7 +572,6 @@ class SentinelHubInputTask(SentinelHubInputBaseTask):
             size=(size_x, size_y),
             data_folder=self.cache_folder,
             config=self.config,
-            geometry=geometry
         )
 
     def _extract_data(self, eopatch, images, shape):
--- a/io/eolearn/io/sentinelhub_process.py
+++ b/io/eolearn/io/sentinelhub_process.py
@@ -12,7 +12,7 @@ file in the root directory of this source tree.
 
 import datetime as dt
 import logging
-from typing import Callable, List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple, Union
 
 import numpy as np
 
@@ -22,6 +22,7 @@ from sentinelhub import (
     DataCollection,
     MimeType,
     MosaickingOrder,
+    ResamplingType,
     SentinelHubCatalog,
     SentinelHubDownloadClient,
     SentinelHubRequest,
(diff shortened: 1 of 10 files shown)


io/eolearn/tests/test_sentinelhub_process.py
```diff
@@ -164,6 +167,47 @@ def test_upsampling_downsampling(self, resampling_type: ResamplingType, stats: L
...
+            (
+                Geometry(
+                    BBox(
+                        [easting - 20, northing - 20, easting + (size[0] + 10) * 10, northing + (size[1] + 10) * 10],
+                        crs=crs,
+                    ).geometry,
+                    crs=crs,
+                ),
+                [0.0836, 0.1547, 0.0794],
+            ),
```
zigaLuksic, line 189: If i understand correctly, this test requests the same area but with two different geometries?

io/eolearn/io/sentinelhub_process.py
```diff
@@ -83,11 +85,13 @@ def __init__(
...
-        bbox=None,
-        time_interval=None,
+        eopatch: Optional[EOPatch] = None,
+        bbox: Optional[BBox] = None,
+        time_interval: Optional[RawTimeIntervalType] = None,
+        geometry: Optional[Geometry] = None,
     ):
-        """Main execute method for the Process API tasks"""
+        """Main execute method for the Process API tasks.
+        The geometry is used only if also the bbox is provided."""
```
zigaLuksic, line 94:
```suggestion
        The `geometry` is used only in conjunction with the `bbox` and does not act as a replacement."""
```

zigaLuksic (commented):
I see that the history is a bit off (due to us merging MRs with squash commits). It would be nice if you could fix that (probably easiest if you open a new MR) but It's not that big of an issue so if you're short on time we can leave it as is.

io/eolearn/tests/test_sentinelhub_process.py
devisperessutti: yes, the first geometry is smaller than the bounding box, while the second is larger than the bounding box.

We wanted to test that in the second case you get the same results as requesting only the bounding box.

io/eolearn/io/sentinelhub_process.py
nkadunc: Changed as suggested

io/eolearn/tests/test_sentinelhub_process.py
```diff
@@ -164,6 +167,47 @@ def test_upsampling_downsampling(self, resampling_type: ResamplingType, stats: L
...
+            max_threads=self.max_threads,
+        )
+
+        eopatch = task.execute(bbox=self.bbox, time_interval=self.time_interval, geometry=geometry)
+        bands = eopatch[(FeatureType.DATA, "BANDS")]
+
+        assert calculate_stats(bands) == approx(stats)
+
+        width, height = self.size
+        assert bands.shape == (4, height, width, 1)
```
jgersak, line 209: Is there a reason why "all" bands are $(4,x,y,1)$ ? or $4$ and $1$ shouldn't be "hard coded"

devisperessutti: I guess given the bbox and time interval we know there will be 4 timestamps, while given which bands we require in the test we set the last dimension, i.e. number of channels

Surely these tests could be beautified and simplified

PR #467, opened 2022-09-20
Add test for extract bands metod

jgersak:
Test checking deep copy of Map Feature Task

### DIFF
--- a/core/eolearn/tests/test_core_tasks.py
+++ b/core/eolearn/tests/test_core_tasks.py
@@ -12,6 +12,7 @@ file in the root directory of this source tree.
 
 import copy
 import datetime
+import math
 import pickle
 from typing import Dict, Iterable, Tuple, Union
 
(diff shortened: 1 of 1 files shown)


core/eolearn/tests/test_core_tasks.py
```diff
@@ -425,6 +426,11 @@ def test_extract_bands(test_eopatch):
     patch = move_bands(test_eopatch)
     assert np.array_equal(patch.data["MOVED_BANDS"], patch.data["REFERENCE_SCENES"][..., bands])
 
+    patch.data["MOVED_BANDS"][0, 0, 0, 0] += 1.0
+    assert math.isclose(
+        patch.data["REFERENCE_SCENES"][0, 0, 0, bands[0]] + 1.0, patch.data["MOVED_BANDS"][0, 0, 0, 0], rel_tol=0.2
+    )
+
```
zigaLuksic, line 433: use `pytest`s `approx` function, since it plays nicely with the asserts.

I'd also prefer to memorize the old value, change it to a new one, then check that it did change in the moved bands but not in the reference scenes (where it should match the old value).

core/eolearn/tests/test_core_tasks.py
```diff
@@ -426,10 +426,10 @@ def test_extract_bands(test_eopatch):
     patch = move_bands(test_eopatch)
     assert np.array_equal(patch.data["MOVED_BANDS"], patch.data["REFERENCE_SCENES"][..., bands])
 
+    tmp = patch.data["MOVED_BANDS"][0, 0, 0, 0]
     patch.data["MOVED_BANDS"][0, 0, 0, 0] += 1.0
-    assert math.isclose(
-        patch.data["REFERENCE_SCENES"][0, 0, 0, bands[0]] + 1.0, patch.data["MOVED_BANDS"][0, 0, 0, 0], rel_tol=0.2
-    )
+    assert_equal(patch.data["REFERENCE_SCENES"][0, 0, 0, bands[0]], tmp)
```
zigaLuksic, line 431: these should be identical values, so a simple `==` comparison should suffice, no?

core/eolearn/tests/test_core_tasks.py
```diff
@@ -426,10 +426,10 @@ def test_extract_bands(test_eopatch):
     patch = move_bands(test_eopatch)
     assert np.array_equal(patch.data["MOVED_BANDS"], patch.data["REFERENCE_SCENES"][..., bands])
 
+    tmp = patch.data["MOVED_BANDS"][0, 0, 0, 0]
```
zigaLuksic, line 429: `old_value` would probably be a better name than `tmp`

zigaLuksic (approved):
Great, thanks :D

PR #469, opened 2022-09-22
Feat/types

zigaLuksic:
This adds types to the feature parsers (without changing their functionality)

1. Adds a `eolearn.core.utils.types` module which has the `EllipsisType` handled for pre-3.10 versions
2. Adds type annotations to parsers (any suggestions on how to organize the huge types would be welcome)
3. Change defaults in `allowed_feature_types` to be more in-line with the philosophy of using `...`. The change was made in a way that it should be non-codebreaking (so passing in a None would count as `...`).

I am also not entirely satisfied by names.

### DIFF
--- a/core/eolearn/core/utils/parsing.py
+++ b/core/eolearn/core/utils/parsing.py
@@ -29,14 +29,21 @@ if TYPE_CHECKING:
     from ..eodata import EOPatch
 
 
-SingleFeatureSpec = Union[
-    Literal[FeatureType.TIMESTAMP, FeatureType.BBOX], Tuple[FeatureType, str], Tuple[FeatureType, str, str]
+FeatureSpec = Union[Tuple[Literal[FeatureType.BBOX, FeatureType.TIMESTAMP], None], Tuple[FeatureType, str]]
+FeatureRenameSpec = Union[
+    Tuple[Literal[FeatureType.BBOX, FeatureType.TIMESTAMP], None, None], Tuple[FeatureType, str, str]
 ]
-SequenceFeatureSpec = Sequence[Union[SingleFeatureSpec, Tuple[FeatureType, EllipsisType]]]
+SingleFeatureSpec = Union[FeatureSpec, FeatureRenameSpec]
 
+SequenceFeatureSpec = Sequence[Union[SingleFeatureSpec, FeatureType, Tuple[FeatureType, Optional[EllipsisType]]]]
 DictFeatureSpec = Dict[FeatureType, Union[None, EllipsisType, Iterable[Union[str, Tuple[str, str]]]]]
+MultiFeatureSpec = Union[
+    EllipsisType, FeatureType, Tuple[FeatureType, EllipsisType], SequenceFeatureSpec, DictFeatureSpec
+]
 
-MultiFeatureSpec = Union[EllipsisType, Tuple[FeatureType, EllipsisType], SequenceFeatureSpec, DictFeatureSpec]
+FeaturesSpecification = Union[SingleFeatureSpec, MultiFeatureSpec]
+
+_ParserFeaturesSpec = Union[Tuple[FeatureType, None, None], Tuple[FeatureType, str, str]]
 
 
 class FeatureParser:
@@ -100,7 +107,7 @@ class FeatureParser:
 
     def __init__(
         self,
-        features: Union[SingleFeatureSpec, MultiFeatureSpec],
+        features: FeaturesSpecification,
         allowed_feature_types: Union[Iterable[FeatureType], EllipsisType] = ...,
     ):
         """
@@ -114,9 +121,7 @@ class FeatureParser:
         )
         self._feature_specs = self._parse_features(features)
 
-    def _parse_features(
-        self, features: Union[SingleFeatureSpec, MultiFeatureSpec]
-    ) -> List[Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]]:
+    def _parse_features(self, features: FeaturesSpecification) -> List[_ParserFeaturesSpec]:
         """This method parses and validates input, returning a list of `(ftype, old_name, new_name)` triples.
 
         Due to typing issues the all-features requests are transformed from `(ftype, ...)` to `(ftype, None, None)`.
@@ -124,7 +129,7 @@ class FeatureParser:
         either by processing the request or by substituting ellipses back (case of `get_feature_specifications`).
         """
 
-        if features is FeatureType.BBOX or features is FeatureType.TIMESTAMP:
+        if isinstance(features, FeatureType):
             return [(features, None, None)]
 
         if isinstance(features, dict):
@@ -143,10 +148,10 @@ class FeatureParser:
     def _parse_dict(
         self,
         features: DictFeatureSpec,
-    ) -> List[Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]]:
+    ) -> List[_ParserFeaturesSpec]:
         """Implements parsing and validation in case the input is a dictionary."""
 
-        feature_specs: List[Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]] = []
+        feature_specs: List[_ParserFeaturesSpec] = []
 
         for feature_type, feature_names in features.items():
             feature_type = self._parse_feature_type(feature_type, message_about_position="keys of the dictionary")
@@ -170,10 +175,10 @@ class FeatureParser:
         features: Union[
             Tuple[FeatureType, str], Tuple[FeatureType, str, str], Tuple[FeatureType, EllipsisType], SequenceFeatureSpec
         ],
-    ) -> List[Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]]:
+    ) -> List[_ParserFeaturesSpec]:
         """Implements parsing and validation in case the input is a tuple describing a single feature or a sequence."""
 
-        feature_specs: List[Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]] = []
+        feature_specs: List[_ParserFeaturesSpec] = []
 
         # Check for possible singleton
         if 2 <= len(features) <= 3:
@@ -198,9 +203,7 @@ class FeatureParser:
 
         return feature_specs
 
-    def _parse_singleton(
-        self, feature: Sequence
-    ) -> Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]:
+    def _parse_singleton(self, feature: Sequence) -> FeatureRenameSpec:
         """Parses a pair or triple specifying a single feature or a get-all request."""
         feature_type, *feature_name = feature
         feature_type = self._parse_feature_type(feature_type, message_about_position="first elements of tuples")
@@ -261,25 +264,6 @@ class FeatureParser:
                 f" {specification} instead."
             )
 
-    @staticmethod
-    def _validate_parsing_request(feature_type: FeatureType, name: Optional[str], eopatch: Optional[EOPatch]) -> None:
-        """Checks if the parsing request is viable with current arguments.
-
-        This means checking that `eopatch` is provided if the request is an all-features request and in the case
-        where an EOPatch is provided, that the feature exists in the EOPatch.
-        """
-        if not feature_type.has_dict():
-            return
-
-        if name is None and eopatch is None:
-            raise ValueError(
-                f"Input specifies that for feature type {feature_type} all existing features are parsed, but the "
-                "`eopatch` parameter was not provided."
-            )
-
-        if eopatch is not None and name is not None and (feature_type, name) not in eopatch:
-            raise ValueError(f"Requested feature {(feature_type, name)} not part of eopatch.")
-
     def get_feature_specifications(self) -> List[Tuple[FeatureType, Union[str, EllipsisType]]]:
         """Returns the feature specifications in a more streamlined fashion.
 
@@ -287,7 +271,7 @@ class FeatureParser:
         """
         return [(ftype, ... if fname is None else fname) for ftype, fname, _ in self._feature_specs]
 
-    def get_features(self, eopatch: Optional[EOPatch] = None) -> List[Tuple[FeatureType, Optional[str]]]:
+    def get_features(self, eopatch: Optional[EOPatch] = None) -> List[FeatureSpec]:
         """Returns a list of `(feature_type, feature_name)` pairs.
 
         For features that specify renaming, the new name of the feature is ignored.
@@ -297,20 +281,10 @@ class FeatureParser:
 
         If `eopatch` is not provided the method fails if an all-feature request is in the specification.
         """
-        feature_names: List[Tuple[FeatureType, Optional[str]]] = []
-        for ftype, name, _ in self._feature_specs:
-            self._validate_parsing_request(ftype, name, eopatch)
-            if name is None and ftype.has_dict():
-                # _validate_parsing_request should ensure that eopatch is not None
-                feature_names.extend((ftype, name) for name in eopatch[ftype])  # type: ignore[index]
-            else:
-                feature_names.append((ftype, name))
-        return feature_names
+        renamed_features = self.get_renamed_features(eopatch)
+        return [feature[:2] for feature in renamed_features]  # pattern unpacking messes with typechecking
 
-    def get_renamed_features(
-        self,
-        eopatch: Optional[EOPatch] = None,
-    ) -> List[Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]]:
+    def get_renamed_features(self, eopatch: Optional[EOPatch] = None) -> List[FeatureRenameSpec]:
         """Returns a list of `(feature_type, old_name, new_name)` triples.
 
         For features without a specified renaming the new name is equal to the old one.
@@ -321,16 +295,27 @@ class FeatureParser:
 
         If `eopatch` is not provided the method fails if an all-feature request is in the specification.
         """
-        feature_names: List[Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]] = []
+        parsed_features: List[FeatureRenameSpec] = []
         for feature_spec in self._feature_specs:
-            ftype, old_name, _ = feature_spec
-            self._validate_parsing_request(ftype, old_name, eopatch)
-            if old_name is None and ftype.has_dict():
-                # _validate_parsing_request should ensure that eopatch is not None
-                feature_names.extend((ftype, name, name) for name in eopatch[ftype])  # type: ignore[index]
+            ftype, old_name, new_name = feature_spec
+
+            if ftype is FeatureType.BBOX or ftype is FeatureType.TIMESTAMP:
+                parsed_features.append((ftype, None, None))
+
+            elif old_name is not None and new_name is not None:
+                # checking both is redundant, but typechecker has difficulties otherwise
+                if eopatch is not None and (ftype, old_name) not in eopatch:
+                    raise ValueError(f"Requested feature {(ftype, old_name)} not part of eopatch.")
+                parsed_features.append((ftype, old_name, new_name))
+
+            elif eopatch is not None:
+                parsed_features.extend((ftype, name, name) for name in eopatch[ftype])
             else:
-                feature_names.append(feature_spec)
-        return feature_names
+                raise ValueError(
+                    f"Input of feature parser specifies that for feature type {ftype} all existing features are parsed,"
+                    " but the `eopatch` parameter was not provided."
+                )
+        return parsed_features
 
 
 def parse_feature(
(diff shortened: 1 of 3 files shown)


core/eolearn/core/utils/parsing.py
```diff
@@ -13,15 +13,32 @@
...
+    from typing import Literal  # pylint: disable=ungrouped-imports
 
 if TYPE_CHECKING:
     from ..eodata import EOPatch
 
 
+SingleFeatureSpec = Union[
+    Literal[FeatureType.TIMESTAMP, FeatureType.BBOX], Tuple[FeatureType, str], Tuple[FeatureType, str, str]
+]
+SequenceFeatureSpec = Sequence[Union[SingleFeatureSpec, Tuple[FeatureType, EllipsisType]]]
```
mlubej, line 35: Could `Tuple[FeatureType, EllipsisType]` not be considered as a `SingleFeatureSpec` ?

core/eolearn/core/utils/parsing.py
```diff
@@ -13,15 +13,32 @@
...
+
+DictFeatureSpec = Dict[FeatureType, Union[None, EllipsisType, Iterable[Union[str, Tuple[str, str]]]]]
+
+MultiFeatureSpec = Union[EllipsisType, Tuple[FeatureType, EllipsisType], SequenceFeatureSpec, DictFeatureSpec]
```
mlubej, line 39: I would keep the names, except perhaps for:
- `SingleFeatureSpec` -> `FeatureSpec` ?

I'd say it's obvious it's a single feature thing even when it's used side by side with `MultiFeatureSpec`.

core/eolearn/core/utils/parsing.py
```diff
@@ -13,15 +13,32 @@
...
 if TYPE_CHECKING:
     from ..eodata import EOPatch
 
 
+SingleFeatureSpec = Union[
+    Literal[FeatureType.TIMESTAMP, FeatureType.BBOX], Tuple[FeatureType, str], Tuple[FeatureType, str, str]
+]
+SequenceFeatureSpec = Sequence[Union[SingleFeatureSpec, Tuple[FeatureType, EllipsisType]]]
+
+DictFeatureSpec = Dict[FeatureType, Union[None, EllipsisType, Iterable[Union[str, Tuple[str, str]]]]]
```
mlubej, line 37: I assume the `Dict[FeatureType, Union[None]]` part is for the `TIMESTAMP` and `BBOX`?

If yes, then perhaps would it be more explicit to do:

```python
BaseFeatureSpec = Literal[FeatureType.TIMESTAMP, FeatureType.BBOX]
DictFeatureSpec = Union[Dict[BaseFeatureSpec, None], Dict[FeatureType, Union[EllipsisType, Iterable[Union[str, Tuple[str, str]]]]]]
```

where `BaseFeatureSpec` could be used also above.

core/eolearn/core/utils/parsing.py
```diff
@@ -81,25 +98,35 @@ class FeatureParser:
     - For `get_renamed_features` a list of triples `(feature_type, old_name, new_name)`.
     """
 
-    def __init__(self, features: Union[dict, Sequence], allowed_feature_types: Optional[Iterable[FeatureType]] = None):
+    def __init__(
+        self,
+        features: Union[SingleFeatureSpec, MultiFeatureSpec],
```
mlubej, line 103: I know it might be too late to make such a change now, but the name `features` suggests plural, so perhaps we could drop this convenience of auto recognizing if the argument is a single feature or a list of features. The user would just always provide a list. I believe that would make it simpler, but I'm aware that making such a change for an open-source package might not be as simple.

Anyway, feel free to ignore, just pointing it out.

mlubej, line 103: `Union[SingleFeatureSpec, MultiFeatureSpec]` is used quite a few times throughout the MR. Would it make sense to add a new type for this, such as `AnyFeatureSpec` or something, or is that just complicating?

core/eolearn/core/utils/parsing.py
```diff
@@ -81,25 +98,35 @@ class FeatureParser:
...
         """This method parses and validates input, returning a list of `(ftype, old_name, new_name)` triples.
 
         Due to typing issues the all-features requests are transformed from `(ftype, ...)` to `(ftype, None, None)`.
         This is a correct schema for BBOX and TIMESTAMP while for other features this is corrected when outputting,
         either by processing the request or by substituting ellipses back (case of `get_feature_specifications`).
         """
+
+        if features is FeatureType.BBOX or features is FeatureType.TIMESTAMP:
+            return [(features, None, None)]
+
```
mlubej, line 129: Is this necessary? One could just check if `features` is not a sequence, wrap it in a set if it isn't, and proceed, since such a case is handled below. This way you wouldn't duplicate the logic from L191.

core/eolearn/core/utils/parsing.py
```diff
@@ -139,9 +166,12 @@ def _parse_dict(self, features: dict) -> List[Union[Tuple[FeatureType, str, str]
         return feature_specs
 
     def _parse_sequence(
-        self, features: Sequence
+        self,
+        features: Union[
+            Tuple[FeatureType, str], Tuple[FeatureType, str, str], Tuple[FeatureType, EllipsisType], SequenceFeatureSpec
```
mlubej, line 171: This makes me think if it would make sense to define i.e. 
- `SimpleFeatureSpec = Literal[FeatureType.TIMESTAMP, FeatureType.BBOX]`
- `TupleFeatureSpec = Tuple[FeatureType, str], Tuple[FeatureType, str, str]` 
- `(Single)FeatureSpec = Union[SimpleFeatureSpec, TupleFeatureSpec]`

which could then be used throughout the code

core/eolearn/core/utils/parsing.py
```diff
@@ -267,13 +297,14 @@ def get_features(self, eopatch: Optional[EOPatch] = None) -> List[Tuple[FeatureT
...
-        for feature_type, name, _ in self._feature_specs:
-            self._validate_parsing_request(feature_type, name, eopatch)
-            if name is None and feature_type.has_dict():
-                feature_names.extend(list(zip(repeat(feature_type), eopatch[feature_type])))  # type: ignore
+        feature_names: List[Tuple[FeatureType, Optional[str]]] = []
+        for ftype, name, _ in self._feature_specs:
+            self._validate_parsing_request(ftype, name, eopatch)
+            if name is None and ftype.has_dict():
+                # _validate_parsing_request should ensure that eopatch is not None
+                feature_names.extend((ftype, name) for name in eopatch[ftype])  # type: ignore[index]
```
mlubej, line 305: Can you explain what is this `ignore[index]` ?

core/eolearn/core/utils/parsing.py
```diff
@@ -250,7 +280,7 @@ def _validate_parsing_request(feature_type: FeatureType, name: Optional[str], eo
         if eopatch is not None and name is not None and (feature_type, name) not in eopatch:
             raise ValueError(f"Requested feature {(feature_type, name)} not part of eopatch.")
 
-    def get_feature_specifications(self) -> List[Tuple[FeatureType, object]]:
+    def get_feature_specifications(self) -> List[Tuple[FeatureType, Union[str, EllipsisType]]]:
```
mlubej, line 283: Why not use `Optional[str]` as well, like the list below is defined? Or does it not work that way? I understood the `...` be more used as input.

core/eolearn/core/utils/parsing.py
```diff
@@ -290,20 +321,22 @@ def get_renamed_features(
...
         If `eopatch` is not provided the method fails if an all-feature request is in the specification.
         """
-        feature_names = []
-        for feature_type, old_name, new_name in self._feature_specs:
-            self._validate_parsing_request(feature_type, old_name, eopatch)
-            if old_name is None and feature_type.has_dict():
-                feature_names.extend(
-                    list(zip(repeat(feature_type), eopatch[feature_type], eopatch[feature_type]))  # type: ignore
-                )
+        feature_names: List[Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]] = []
```
mlubej, line 324: This one is also defined quite a few times. Perhaps it could be extracted as `LongFeatureSpec`

The above is only 50 % a joke

core/eolearn/core/utils/parsing.py
```diff
@@ -290,20 +321,22 @@ def get_renamed_features(
...
+        for feature_spec in self._feature_specs:
+            ftype, old_name, _ = feature_spec
+            self._validate_parsing_request(ftype, old_name, eopatch)
+            if old_name is None and ftype.has_dict():
+                # _validate_parsing_request should ensure that eopatch is not None
+                feature_names.extend((ftype, name, name) for name in eopatch[ftype])  # type: ignore[index]
             else:
-                feature_names.append((feature_type, old_name, new_name))
+                feature_names.append(feature_spec)
```
mlubej, line 332: This was a bit too much for my brain, so I only skimmed it

mlubej (commented):
You asked for it.

core/eolearn/core/utils/parsing.py
```diff
@@ -13,15 +13,32 @@
...
-from typing import TYPE_CHECKING, Iterable, List, Optional, Sequence, Tuple, Union, cast
+from typing import TYPE_CHECKING, Dict, Iterable, List, Optional, Sequence, Tuple, Union, cast
 
 from ..constants import FeatureType
+from .types import EllipsisType
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
+else:
+    from typing import Literal  # pylint: disable=ungrouped-imports
```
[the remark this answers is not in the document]
zigaLuksic: I'd prefer to do as little re-importing as possible, as well as avoid users importing Literal from `utils.types`, because this will be a normal import next year (once 3.7 is not supported anymore). But I do agree it's ugly.

core/eolearn/core/utils/parsing.py
zigaLuksic: Most of the time it represents multiple features, since `(FeatureType.DATA, ...)` gets you all the data features. These cases are really hard to classify...

core/eolearn/core/utils/parsing.py
zigaLuksic: I was avoiding that name because of `eo-grow` :see_no_evil: Because in most `eo-learn` uses you need to provide input for the feature parser which is the `Union[single, multi]`. Also another case for why we'd add the `AnyFeatureSpec` (perhaps just named `FeaturesSpec`? Names are stupid).

core/eolearn/core/utils/parsing.py
zigaLuksic: what you wrote wont work for `{FeatureType.BBOX: None, FeatureType.DATA: ...}`

zigaLuksic: this could perhaps be solved with typed dictionaries, but I'm afraid of adding even more types to this monstrosity. I'd prefer to greatly reduce the parser, but open source is a curse in that regard.

core/eolearn/core/utils/parsing.py
zigaLuksic: I tried doing it in `eo-learn 1.0` but it would break a lot of the tasks. Perhaps by introducing gradual changes we can slowly get the parser to a manageable level.

zigaLuksic: it would make sense, a lot of the tasks will use such a notation.

core/eolearn/core/utils/parsing.py
zigaLuksic: I feel like it obscures the flow a bit more, since you now need to go through another 100 lines just to figure out what happened to your `FeatureType.BBOX`. I also don't want to cast anything non-sequency into a sequence, or the users will complain when I take that away from them.

core/eolearn/core/utils/parsing.py
zigaLuksic: Yeah, but imagine having to come up with all the names :see_no_evil:  was also thinking of a similar approach, maybe i should try

core/eolearn/core/utils/parsing.py
zigaLuksic: the `_validate_parsing_request` should ensure that `eopatch` is not `None`. But `mypy` has no way of knowing that. So I have to ignore the case where `eopatch` is `None`.

core/eolearn/core/utils/parsing.py
zigaLuksic: Having `(FeatureType.DATA, None)` stand for all the data features was a workaround that we had to do before we handled the ellipsis type. And internally it's a bit cleaner. But we needed a method that returns them with `...` because of our filesystem module. Hence this method exists.

core/eolearn/core/utils/parsing.py
zigaLuksic: i just moved the unpacking from the `for` loop to line `326`, so that I can use the tuple in the `else`, otherwise the typechecker complains.

core/eolearn/core/utils/parsing.py
mlubej: I guess this is also one of those cases where it ended up hurting us that a dict-like functionality is possible in the first place. I don't think it's such a big issue to just provide a list of features, so perhaps this could as well be ultimately deprecated.

mlubej: Nothing left to do here, perhaps just write down these ideas somewhere so it doesn't get lost.

Resolving.

core/eolearn/core/utils/parsing.py
mlubej: Good point about `eo-grow` - unless it's changed there, since it's less widely used than `eo-learn` at this point?

I like `FeaturesSpec`, I'm only skeptical about potentially misreading it, but typing is anyway meant to limit the users stupidity, not something which they would use themselves.

core/eolearn/core/utils/parsing.py
mlubej: OK, thanks for pointing out that use case. Resolving.

core/eolearn/core/utils/parsing.py
mlubej: I agree it would introduce breaking changes, that was expected, but I still think it might be worth it to do it over the span of a few releases, first adding a deprecation warning and then removing it. Perhaps a good first task?

mlubej: I'll resolve this thread here, let's continue in https://github.com/sentinel-hub/eo-learn/pull/469#discussion_r978494404

core/eolearn/core/utils/parsing.py
mlubej: Yes, it does obscure it...

Actually, I dislike the fact that a single tuple-like feature and the sequence of features are treated the same upon entry. It's like I would prefer to have a Feature class for this, which has a `type` and a `name`, where the `name` can be none, and also all the other properties like `is_timeless` and so on. I think it might make typing in such cases also easier.

Was such a plan ever considered and not followed due to some issues I'm unable to see at this point?

mlubej: I guess one issue here is that it introduces complexity for the users, but perhaps the users could provide the same things and it would just be used in the core functionality as a constructor of the feature class or something.

core/eolearn/core/utils/parsing.py
mlubej: Dick way of saying you don't like the names I proposed!

core/eolearn/core/utils/parsing.py
mlubej: What if you do a hack that `_validate_parsing_request` accepts `Optional[EOPatch]` but then either returns an `EOPatch` or raises an error?

It's hacky, but that way the type changes and you wouldn't need to ignore it.

_this is just something that popped up into my mind, please note I'm not a typing expert_ 🙈

core/eolearn/core/utils/parsing.py
zigaLuksic: `eo-learn` is intended to be 'user friendly' so yeah, we want to support many easy ways for them to specify features.

core/eolearn/core/utils/parsing.py
zigaLuksic: i managed to throw it out in the new commit :D

core/eolearn/core/utils/parsing.py
zigaLuksic: I was thinking of having `FeatureSpec` as it is in eo-grow, but the full type is `FeaturesSpecification` to avoid confusion a bit

core/eolearn/core/utils/parsing.py
mlubej: Feel free to resolve if this is outdated or if you don't plan on making any further changes related to this.

core/eolearn/core/utils/parsing.py
mlubej: Feel free to resolve if this is outdated or if you don't plan on making any further changes related to this.

core/eolearn/core/utils/parsing.py
```diff
@@ -13,15 +13,39 @@
...
 
 if TYPE_CHECKING:
     from ..eodata import EOPatch
 
 
+FeatureSpec = Union[Tuple[Literal[FeatureType.BBOX, FeatureType.TIMESTAMP], None], Tuple[FeatureType, str]]
```
mlubej, line 32: Would `[FeatureType.TIMESTAMP, 'something']` pass? Is that an issue if it does?

core/eolearn/core/utils/parsing.py
```diff
@@ -139,11 +171,14 @@ def _parse_dict(self, features: dict) -> List[Union[Tuple[FeatureType, str, str]
         return feature_specs
 
     def _parse_sequence(
-        self, features: Sequence
-    ) -> List[Union[Tuple[FeatureType, str, str], Tuple[FeatureType, None, None]]]:
-        """Implements parsing and validation in case the input is a sequence."""
+        self,
+        features: Union[
+            Tuple[FeatureType, str], Tuple[FeatureType, str, str], Tuple[FeatureType, EllipsisType], SequenceFeatureSpec
```
mlubej, line 176: Can't this now be `Union[SingleFeatureSpec, SequenceFeatureSpec]`? There are some differences, but I though that bbox/timestamp cases in a tuple with `None` could in principle also be expected, or not?

core/eolearn/core/utils/parsing.py
```diff
@@ -267,19 +281,10 @@ def get_features(self, eopatch: Optional[EOPatch] = None) -> List[Tuple[FeatureT
...
-        feature_names = []
-        for feature_type, name, _ in self._feature_specs:
-            self._validate_parsing_request(feature_type, name, eopatch)
-            if name is None and feature_type.has_dict():
-                feature_names.extend(list(zip(repeat(feature_type), eopatch[feature_type])))  # type: ignore
-            else:
-                feature_names.append((feature_type, name))
-        return feature_names
+        renamed_features = self.get_renamed_features(eopatch)
+        return [feature[:2] for feature in renamed_features]  # pattern unpacking messes with typechecking
```
mlubej, line 285: Hmm, it's weird that this is necessary, how does it understand the `renamed_features` if this unpacking is not done?

core/eolearn/core/utils/parsing.py
zigaLuksic: It would. I don't see much issue with that, otherwise the types get incredibly specific to the point of unreadable :(
One of the core issues with feature parsing is the fact that timestamps and bbox need to be unique. Probably a design flaw that they are treated as features at all.

core/eolearn/core/utils/parsing.py
zigaLuksic: If i do `[(ftype, name) for ftype, name, _ in renamed_features]` it has `ftype: FeatureType` and `name: Optional[str]`, and it loses the link that only `FeatureType.BBOX` (and timestamp) gets a `None` value.

core/eolearn/core/utils/parsing.py
mlubej: Yes, I agree, these exceptions are causing headaches. perhaps some similar approach to datacubes should be considered, where timestamps and bbox are treated as coordinates (t, x, y). 

Resolving.

core/eolearn/core/utils/parsing.py
zigaLuksic: Good catch, there is a potential issue here indeed. Mypy for some reason refines the type to just `SequenceFeatureSpec` and ignores the possibility of Tuples. Will need to figure out a way to clarify that :thinking:

core/eolearn/core/utils/parsing.py
mlubej: Ah, sorry, my bad, I thought `[:2]` took all three elements. I'm more used to seeing `[:-1]`. Ignore this thread.

core/eolearn/core/utils/parsing.py
zigaLuksic: I switched to your suggestion. Mypy continues to wrongly refine the type, but that is a bug which i reported to the mypy repo. No easy way to circumvent things, `Iterable` and `Sequence` are apparently giving them headaches with type refinement.

PR #473, opened 2022-09-28
Avoid inconsistency in paths of FeatureIO

zigaLuksic:
When designing the tests it came up that `FeatureIO` cannot load it's own saved file, due to the inconsistencies with paths.

On saving the object assumed paths returned by `walk_eopatch` which are just a file-system-structured paths of features (e.g. `"eopatch/data/NDVI"`) and lack any extensions.

On loading the object assumed filesystem paths (e.g. `"eopatch/data/NDVI.npy.gz"`).

Turns out it is not that simple to unify the two. What I did is:
- paths are always assumed to be filesystem paths with full extensions
- they are checked for compliance with feature type extension and compression level
- if the compression level is not specified, it is inferred from path (so 0 if no `.gz` ending and 1 otherwise), which is needed for loading, since we have no clue how the feature was saved other than inspecting the path
- for working with 'eopatch paths' a new constructor was added that automatically adds extensions to produce a proper filesystem path

### DIFF
--- a/core/eolearn/core/eodata_io.py
+++ b/core/eolearn/core/eodata_io.py
@@ -16,7 +16,7 @@ import json
 import warnings
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict
-from typing import Any, BinaryIO, Generic, Iterable, Optional, TypeVar, Union
+from typing import Any, BinaryIO, Generic, Iterable, Optional, Type, TypeVar, Union
 
 import fs
 import fs.move
@@ -36,6 +36,7 @@ from .utils.parsing import FeatureParser
 from .utils.vector_io import infer_schema
 
 _T = TypeVar("_T")
+Self = TypeVar("Self", bound="FeatureIO")
 
 
 def save_eopatch(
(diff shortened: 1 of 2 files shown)


core/eolearn/core/eodata_io.py
```diff
@@ -403,12 +430,12 @@ def _jsonify_timestamp(param: object) -> str:
...
-        return FeatureIOBBox(path, filesystem)
+        return FeatureIOBBox
     if ftype in (FeatureType.TIMESTAMP, FeatureType.META_INFO):
-        return FeatureIOJson(path, filesystem)
+        return FeatureIOJson
     if ftype in FeatureTypeSet.VECTOR_TYPES:
-        return FeatureIOGeoDf(path, filesystem)
-    return FeatureIONumpy(path, filesystem)
+        return FeatureIOGeoDf
+    return FeatureIONumpy
```
jgersak, line 441: shouldn't there be another if and default return value some error flag ?

zigaLuksic: This would occur only if we would add a new feature type that requires a new FeatureIO object, and forget to handle it here (but there would be errors as soon as one would try to save it), so I don't think it's worth another 3 lines of code :thinking:

jgersak (approved):

PR #480, opened 2022-09-30
FeatureIO for timestamp

jgersak:
add sub class for handling timestamps
add test

### DIFF
--- a/core/eolearn/core/eodata_io.py
+++ b/core/eolearn/core/eodata_io.py
@@ -38,6 +38,7 @@ from typing import (
     cast,
 )
 
+import dateutil.parser
 import fs
 import fs.move
 import geopandas as gpd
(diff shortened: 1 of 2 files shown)


core/eolearn/core/eodata_io.py
```diff
@@ -453,6 +454,14 @@ def _write_to_file(cls, data: _T, file: Union[BinaryIO, gzip.GzipFile], path: st
         file.write(json_data.encode())
 
 
+class FeatureIOTimeStamp(FeatureIOJson[_T]):
```
zigaLuksic, line 457: This FeatureIO is specialized for `List[dt.datetime]` and should be marked as such. Just change both occurrences of `_T` with the mentioned type. This will correctly adjust the types of `save` and `load` methods.

Also we treat timestamp as a single word, therefore you should use `Timestamp` and not `TimeStamp`.

core/eolearn/tests/test_eodata_io.py
```diff
@@ -314,6 +321,7 @@ def assert_data_equal(data1: Any, data2: Any) -> None:
             ),
         ),
         (FeatureIOJson, {"test": "test1", "test3": {"test": "test1"}}),
+        (FeatureIOTimeStamp, [datetime.datetime(2017, 1, 1, 10, 4, 7), datetime.datetime(2017, 1, 4, 10, 14, 5)]),
```
zigaLuksic, line 324: Do you think it would be worthwhile to add a test for an empty list? Perhaps one for an empty dictionary in FeatureIOJson as well...

PR #483, opened 2022-10-03
Add mypy to ci

zigaLuksic:


.github/workflows/ci_action.yml
```diff
@@ -35,6 +35,33 @@ jobs:
...
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ env.CHECKOUT_BRANCH }}
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
```
jgersak, line 49: Should not be 3.7 oldest that is still supported ?

zigaLuksic: i'd use the same one as for commit hooks

jgersak (approved):

PR #499, opened 2022-11-14
Improve type signatures in sh-io

zigaLuksic:
The file itself is completely untypeable without major refactoring. This fixes some mistakes and adds some additional types to more public methods.


jgersak (approved):

batic:
This forbids passing `time_interval` as `[str, str]`, which is what we do most of the time, meaning this code change will break most(?) a lot of the examples/notebooks/... we have.

PR #502, opened 2022-11-16
Small refactor fo sentinelhub IO tasks

zigaLuksic:
There were quite some things that just straight up didn't make sense. So i changed them.

It might be easier to check changes commit by commit, since a lot of lines were changed in the end.

### DIFF
--- a/io/eolearn/io/sentinelhub_process.py
+++ b/io/eolearn/io/sentinelhub_process.py
@@ -97,7 +97,7 @@ class SentinelHubInputBaseTask(EOTask):
 
         eopatch = eopatch or EOPatch()
 
-        self._check_and_set_eopatch_bbox(bbox, eopatch)
+        eopatch.bbox = self._extract_bbox(bbox, eopatch)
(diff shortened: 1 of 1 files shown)


io/eolearn/io/sentinelhub_process.py
```diff
@@ -301,7 +283,7 @@ def _create_response_objects(self):
 
         return responses
 
-    def _get_timestamp(self, time_interval, bbox):
+    def _get_timestamp(self, time_interval: Optional[RawTimeIntervalType], bbox: BBox) -> List[dt.datetime]:
```
jgersak, line 286: Optional[RawTimeIntervalType], bbox: BBox) -> List[RawTimeIntervalType] ?

jgersak (approved):

io/eolearn/io/sentinelhub_process.py
zigaLuksic: what we get out are timestamps, and not intervals

PR #504, opened 2022-11-18
Feat/use sh testing utils

zigaLuksic:
In many tests we used statistical comparisons, but we have a utility function for that in `sh-py`! And it works pretty well  if one enables pytest assert propagation to reach it (done in `conftest.py`).

I updated all the tests in `eolearn.features` to this format, and the tests become a fair bit shorter and much more readable (when you look at the test case you know which number is which statistic without having to memorize order etc.)

Some values had to be adjusted because the delta went from absolute to relative (current util implementation does not support absolute delta, will need to update)

### DIFF
--- a/features/eolearn/tests/conftest.py
+++ b/features/eolearn/tests/conftest.py
@@ -8,6 +8,8 @@ import pytest
 
 from eolearn.core import EOPatch
 
+pytest.register_assert_rewrite("sentinelhub.testing_utils")  # makes asserts in helper functions work with pytest
+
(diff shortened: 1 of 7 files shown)


jgersak (approved):
LGTM
One question should we first check that `added_feature` is really deleted from eopatch 
or that would be duplication of our test of `del` function?

zigaLuksic:
> LGTM One question should we first check that `added_feature` is really deleted from eopatch or that would be duplication of test for `del` function?

yeah, no reason to test `del` here IMO

PR #506, opened 2022-11-21
Adapt test suite to use sh-py testing util for numpy stats

jgersak:


geometry/eolearn/tests/test_superpixel.py
```diff
@@ -52,10 +45,7 @@
                 max_num_iter=20,
                 sigma=0.8,
             ),
-            0,
-            48,
-            24.6072,
-            25,
+            {"exp_dtype": np.int64, "exp_min": 0, "exp_max": 48, "exp_mean": 24.6072, "exp_median": 25},
```
zigaLuksic, line 48: not sure i'd include the expected dtype unless this is something that we really expect
if some other library changes something and now we by default output np.int32, should we be notified of that? We don't need values larger than int32

geometry/eolearn/tests/test_transformations.py
```diff
@@ -245,16 +254,8 @@ def test_vector_to_raster_result(test_case, test_eopatch):
...
-    assert np.amin(result) == pytest.approx(test_case.img_min, abs=delta), "Minimum values do not match."
-    assert np.amax(result) == pytest.approx(test_case.img_max, abs=delta), "Maximum values do not match."
-    if test_case.img_mean is not None:
-        assert np.mean(result) == pytest.approx(test_case.img_mean, abs=delta), "Mean values do not match."
-    if test_case.img_median is not None:
-        assert np.median(result) == pytest.approx(test_case.img_median, abs=delta), "Median values do not match."
-    assert result.dtype == test_case.img_dtype, "Wrong dtype of result."
-    assert result.shape == test_case.img_shape, "Result is of wrong shape."
+
+    test_numpy_data(result, **test_case.img_exp_statistics, delta=2e-3)  # with delta=1e-3 one test fail
```
zigaLuksic, line 258: i'd prefer changing the test numbers to more closely represent the actual output
using a weird delta is confusing

geometry/eolearn/tests/test_superpixel.py
zigaLuksic: nevermind, saw that it was already in the originals

jgersak: Resolved in chat

PR #514, opened 2022-11-28
Remove option to sample random integer points in triangle

jgersak:
update docstring

### DIFF
--- a/ml_tools/eolearn/tests/test_sampling.py
+++ b/ml_tools/eolearn/tests/test_sampling.py
@@ -22,32 +22,26 @@ from eolearn.ml_tools.sampling import expand_to_grids, random_point_in_triangle
 
 
 @pytest.mark.parametrize(
-    "triangle, generator, values",
+    "triangle, expected_points",
     [
         (
             Polygon([[-10, -12], [5, 10], [15, 4]]),
-            np.random.default_rng(seed=42),
             [Point(7.057238842063997, 5.037835974428286), Point(10.360934995972169, 4.508212300648985)],
         ),
         (
             Polygon([[110, 112], [117, 102], [115, 104]]),
-            np.random.default_rng(seed=42),
             [Point(115.38602941857646, 103.97472742532676), Point(115.19386890346114, 104.02631432574218)],
         ),
         (
             Polygon([[0, 0], [5, 12], [15, 4]]),
-            np.random.default_rng(seed=42),
             [Point(8.259761655074744, 7.46815417512301), Point(11.094879093316592, 5.949786169595648)],
         ),
     ],
 )
-def test_random_point_in_triangle_generator(triangle, generator, values):
+def test_random_point_in_triangle_generator(triangle, expected_points):
+    generator = np.random.default_rng(seed=42)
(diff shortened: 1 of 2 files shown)


zigaLuksic (approved):
This MR could also be used to add a simple test for this function?

ml_tools/eolearn/tests/test_sampling.py
```diff
@@ -14,10 +14,65 @@
...
-from eolearn.ml_tools.sampling import expand_to_grids
+from eolearn.ml_tools.sampling import expand_to_grids, random_point_in_triangle
+
+
+@pytest.mark.parametrize(
+    "triangle, generator, values",
+    [
+        (
+            Polygon([[-10, -12], [5, 10], [15, 4]]),
+            np.random.default_rng(seed=42),
```
zigaLuksic, line 29: i guess you could just put a `seed` here? :thinking:
maybe it doesn't even have to be parametrized according to the seed (since you're using the same in all three cases, gets rid of three lines)

ml_tools/eolearn/tests/test_sampling.py
```diff
@@ -14,10 +14,65 @@
...
+        (
+            Polygon([[0, 0], [5, 12], [15, 4]]),
+            np.random.default_rng(seed=42),
+            [Point(8.259761655074744, 7.46815417512301), Point(11.094879093316592, 5.949786169595648)],
+        ),
+    ],
+)
+def test_random_point_in_triangle_generator(triangle, generator, values):
+    points = [random_point_in_triangle(triangle, generator) for _ in range(2)]
+    assert all(point.x == val.x and point.y == val.y for point, val in zip(points, values))
```
zigaLuksic, line 46: does `point1 == point2` not work out of the box? :thinking: 

Also I'd rename `values` to `expected_points` or something similar

ml_tools/eolearn/tests/test_sampling.py
```diff
@@ -14,10 +14,65 @@
...
+
+@pytest.mark.parametrize(
+    "triangle",
+    [
+        (Polygon([[-10, -12], [5, 10], [15, 4]])),
+        (Polygon([[110, 112], [117, 102], [115, 104]])),
+        (Polygon([[0, 0], [5, 12], [15, 4]])),
+    ],
+)
+def test_random_point_in_triangle_interior(triangle):
```
zigaLuksic, line 61: I think it's best to use `shapely` for this, since it's a high grade library.
You can check [this stackoverflow thread](https://stackoverflow.com/questions/39660851/deciding-if-a-point-is-inside-a-polygon) on ways how to do this.

zigaLuksic (changes requested):

ml_tools/eolearn/tests/test_sampling.py
jgersak: done

ml_tools/eolearn/tests/test_sampling.py
jgersak: done

ml_tools/eolearn/tests/test_sampling.py
jgersak: updated

Issue #530, opened 2022-12-29
[BUG] Issue with the SpatialResizeTask in ImportTiffPipeline when using the resolution approach 


mlubej:
I was importing some tiff and wanted to upsample from a 30m res to a 10m res. The EOP after importing the tiff was 167 x 167, but after the `SpatialResizeTask` it was 167 x 500 (or vice versa, can't remember now). Since the resolution change is the same in x and y, I would expect the outcome to be 500 x 500

I went looking in the code and the issue is how the new size is calculated using the [bbox_to_dimensions](https://github.com/sentinel-hub/sentinelhub-py/blob/develop/sentinelhub/geo_utils.py#L16-L31) utility from `sentinelhub-py`. In the [SpatialResizeTask](https://github.com/sentinel-hub/eo-learn/blob/develop/features/eolearn/features/feature_manipulation.py#L284) the tuple of `[old_resolution, new_resolution]` is provided to the `bbox_to_dimensions` function, which understands this tuple as resolution in horizontal and vertical direction, which is wrong. In this case, a single number (target resolution) should be provided to scale equally in both directions.

So, instead of:

```python
new_size = bbox_to_dimensions(eopatch.bbox, self.parameter_values)
```

it should be 

```python
new_res = self.parameter_values[1]
new_size = bbox_to_dimensions(eopatch.bbox, new_res)
```

After fixing this, the issue I was experiencing goes away.

zigaLuksic:
Since the 'old_resolution' is never used, it feels like it's the docstring that is wrong. AFAIK i rewrote the docstring after adding the ImportTiffPipeline, so it is entirely possible that i borked the description because I had certain assumptions that were incorrect.

mlubej:
> Since the 'old_resolution' is never used, it feels like it's the docstring that is wrong. AFAIK i rewrote the docstring after adding the ImportTiffPipeline, so it is entirely possible that i borked the description because I had certain assumptions that were incorrect.

That seems to make sense. If the `old_res` is not needed, then perhaps it would make sense to use a more complex schema instead of the enum class, where the enum would accept the method and some generic values, and these values would be validated based on the method. Unless there is some other, more intelligent way of doing this.

zigaLuksic:
fixed in https://github.com/sentinel-hub/eo-learn/pull/544

PR #531, opened 2022-12-29
Resolve "[BUG] Issue with the SpatialResizeTask in ImportTiffPipeline when using the resolution approach"

mlubej:
Closes https://github.com/sentinel-hub/eo-learn/issues/530


zigaLuksic:
a different solution will be used

PR #542, opened 2023-01-19
Fix eodata tests

zigaLuksic:
1. Refactored old test of `del`
2. Added some failure messages to test for copying + lazy loading
3. added tests for `get_spatial_dimensions`, `get_features` and `reset_feature_type`

### DIFF
--- a/core/eolearn/tests/test_eodata.py
+++ b/core/eolearn/tests/test_eodata.py
@@ -10,7 +10,7 @@ This source code is licensed under the MIT license found in the LICENSE
 file in the root directory of this source tree.
 """
 import datetime
-from typing import Any, Tuple
+from typing import Any, List, Tuple
 
 import numpy as np
 import pytest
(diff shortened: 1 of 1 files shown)


zigaLuksic:
> Have two suggestions: Add scope of `mini_eopatch` to "module" (test will fail, solution do changes on copied EOPatch) Add missing types

what is the point of setting a scope and then copying it manually?

I'll add the types.

jgersak:
> > Have two suggestions: Add scope of `mini_eopatch` to "module" (test will fail, solution do changes on copied EOPatch) Add missing types
> 
> what is the point of setting a scope and then copying it manually?
> 
> I'll add the types.

ATM when for example `test_reset_feature_type`get called:
 `mini_eopatch` is first created and then copied,
if the scope is set to  "module" the creation step is skipped,
 which save space and time ("free lunch" ... more like a gum in this case)

jgersak (commented):
test_delete add case for:
 + bbox (not implemented yet in eodata (probably need costume warning or ...)) 
 + timestamp
 + meta_info
 + and all features ( assert result == EOPatch() )

test_reset_feature_type and  test_get_spatial_dimension add cases for none existing features

test_get_features add case for empty (EOPatch())

jgersak (approved):

PR #544, opened 2023-01-19
Fixes SpatialResizeTask that switched width/height when using resolution

veseln:
Changes that were made: 

* The signature of the SpatialResizeTask was changed to explicitly take width / height parameters. 
* The bug that switched the width/height when using resolution was fixed 
* Tests were updated to reflect the new signature, tests were added to test for non-symettbic width/height parameters for `ResizeType.SCALE_FACTORS` and `ResizeType.RESOLUTION`

### DIFF
--- a/features/eolearn/features/feature_manipulation.py
+++ b/features/eolearn/features/feature_manipulation.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 import datetime as dt
 import logging
 from functools import partial
-from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union, cast
+from typing import Any, Callable, Dict, Iterable, List, Optional, Union, cast
 
 import numpy as np
 from geopandas import GeoDataFrame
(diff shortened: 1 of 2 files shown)


features/eolearn/features/feature_manipulation.py
```diff
@@ -252,38 +252,45 @@ class SpatialResizeTask(EOTask):
...
-        :param resize_parameters: Instructions on how to perform the resizing process. For example use:
-            `['new_size', [500, 1000]]` to resize the data to size (500, 1000),
-            `['resolution', [10, 20]]` for changing resolution from 10 to 20,
-            `['scale_factors', [3, 3]]` to make the data three times larger.
+        :param resize_type: Type of resizing process to be performed. Options:
+            * `new_size` (combined with width_param, height_param). Resizes data to size (width_param, height_param)
+            * `resolution` (combined with width_param, height_param). Resizes the data to have
+               width_param, height_param resolution over width/height axis. Uses EOPatch bbox to compute.
+            * `scale_factor` (combined with width_param, height_param). Resizes the data by scaling the width
+               and height by a factor set by width_param and height_param.
```
zigaLuksic, line 269: since you have `(combined with width_param, height_param)` maybe you can add that on the first line. Something along the lines of
```
:param resize_type: Determines type of resizing process and how `width_param` and `height_param` are used. Options:
```

zigaLuksic (approved):

features/eolearn/features/feature_manipulation.py
veseln: Thanks. Was struggling with wording this. The suggestion is much better than my version. Will fix.

features/eolearn/features/feature_manipulation.py
```diff
@@ -252,38 +252,46 @@ class SpatialResizeTask(EOTask):
...
+        if self.resize_type == ResizeParam.RESOLUTION:
             if not eopatch.bbox:
                 raise ValueError("Resolution-specified resizing can only be done on EOPatches with a defined BBox.")
-            new_size = bbox_to_dimensions(eopatch.bbox, self.parameter_values)
-            resize_fun_kwargs = {ResizeParam.NEW_SIZE.value: new_size}
+            new_width, new_height = bbox_to_dimensions(eopatch.bbox, (self.width_param, self.height_param))
+            resize_fun_kwargs = {ResizeParam.NEW_SIZE.value: (new_height, new_width)}
         else:
-            resize_fun_kwargs = {self.parameter_kind.value: self.parameter_values}
+            resize_fun_kwargs = {self.resize_type.value: (self.height_param, self.width_param)}
```
mlubej, line 294: Just a comment. I find it confusing that the resolution type is handled here separately, while the case for scaling factor is handled inside of the resize function. 

Additionally, the resize function could be improved, since it's a bunch of optional parameters due to the fact that it offers flexibility.

I would suggest that the logic that was implemented here is moved to the resize function, which would accept these new parameters, and that the function is just use here.

zigaLuksic: resolution cannot be handled inside the functions because it works on numpy arrays, but we need a BBox for the resolution. We could however extract that functionality from the helper function into the task.

mlubej: You're right about the bbox thing, thanks.

Is the function used elsewhere? Perhaps your suggestion would make senes.

PR #561, opened 2023-02-16
Add tests for window_indices_refactoring

mlubej:


mask/eolearn/tests/test_cloud_mask.py
```diff
@@ -10,11 +10,34 @@
...
+        [100, 0, 10, (0, 10)],
+        [100, 1, 10, (0, 10)],
+        [100, 50, 10, (45, 55)],
+        [100, 99, 10, (90, 100)],
+        [100, 100, 10, (90, 100)],
+        [100, 0, 11, (0, 11)],
+        [100, 1, 11, (0, 11)],
+        [100, 50, 11, (45, 56)],
+        [100, 99, 11, (89, 100)],
+        [100, 100, 11, (89, 100)],
```
zigaLuksic, line 34: does this half just repeat with a different window?
I also don't see a case where `window_size` is larger than the number of elements.
Lastly, try adjusting `number_of_elements` to something less 'pretty' in some cases

mask/eolearn/tests/test_cloud_mask.py
```diff
@@ -10,11 +10,34 @@
...
+    ),
+)
+def test_window_indices_function(num_of_elements, middle_idx, window_size, expected_indices):
+    indices = _get_window_indices(num_of_elements, middle_idx, window_size)
+    assert indices == expected_indices
+    assert len(range(*indices)) == window_size
```
zigaLuksic, line 40: since these are meant to be indices, how about you construct a list/numpy array of the appropriate length and then apply these indices :thinking: This way tests are also an indicator of how we expect the result to be used. Literature often suggests that test should also serve as documentation (and we're not there yet)

mask/eolearn/tests/test_cloud_mask.py
mlubej: > does this half just repeat with a different window?

I just wanted to add tests for both odd and even window sizes. Would you prefer I implement it differently? Perhaps there is no need to do it at such extent?

> I also don't see a case where window_size is larger than the number of elements.

Yeah, forgot to re-add it after updating. Updated now along with the last remark.

mask/eolearn/tests/test_cloud_mask.py
mlubej: updated now

mask/eolearn/tests/test_cloud_mask.py
zigaLuksic: i usually mix the cases up quite a bit.
In cases where we want to test a lot of options we could use [hypothesis](https://hypothesis.readthedocs.io/en/latest/) and automate that a bit :thinking: certainly something to look.

Resolving this one, if i find an issue in the test cases i'll open a new thread

mask/eolearn/tests/test_cloud_mask.py
```diff
@@ -10,11 +10,39 @@
...
+        [11, 2, 33, (0, 11)],
+    ),
+)
+def test_window_indices_function(num_of_elements, middle_idx, window_size, expected_indices):
+    indices = _get_window_indices(num_of_elements, middle_idx, window_size)
+    assert indices == expected_indices
+
+    test_list = list(range(num_of_elements))
+    test_slice = test_list[slice(*indices)]
+    assert len(test_slice) == min(num_of_elements, window_size)
```
zigaLuksic, line 45: How about this?

```suggestion
    min_idx, max_idx = _get_window_indices(num_of_elements, middle_idx, window_size)
    assert min_idx, max_idx == expected_indices

    test_list = list(range(num_of_elements))
    assert len(test_list[min_idx:max_idx]) == min(num_of_elements, window_size)
```

zigaLuksic (approved):
it feels like a few cases are redundant, but i'll take too many over too little, so i'm ok with that

check the suggestion and include if it feels right, merge after :)

PR #563, opened 2023-02-16
Slight refactoring of the multitemporal cloud detection functions

mlubej:
Changes:
- added types where relevant
- expanded on some names and made them more intuitive
- added comments here and there
- a bit larger refactoring of the `_update_batches`

### DIFF
--- a/mask/eolearn/mask/cloud_mask.py
+++ b/mask/eolearn/mask/cloud_mask.py
@@ -390,32 +390,31 @@ class CloudMaskTask(EOTask):
         n_times = bands.shape[0]
 
         local_avg, local_var = None, None
-        prev_min_tdx, prev_max_tdx = None, None
+        prev_left, prev_right = None, None
 
-        for tdx in range(n_times):
+        for t_idx in range(n_times):
             # Extract temporal window indices
-            min_tdx, max_tdx = _get_window_indices(n_times, tdx, self.max_proc_frames)
-            rel_tdx = tdx - min_tdx
+            left, right = _get_window_indices(n_times, t_idx, self.max_proc_frames)
 
-            bands_slice = bands[min_tdx:max_tdx]
-            is_data_slice = is_data[min_tdx:max_tdx]
+            bands_slice = bands[left:right]
+            is_data_slice = is_data[left:right]
             masked_bands = np.ma.array(bands_slice, mask=~is_data_slice.repeat(bands_slice.shape[-1], axis=-1))
 
             # Calculate the averages/variances for the local (windowed) streaming data
-            if local_avg is None or (min_tdx, max_tdx) != (prev_min_tdx, prev_max_tdx):
+            if local_avg is None or (left, right) != (prev_left, prev_right):
                 local_avg, local_var = self._update_batches(local_avg, local_var, bands_slice, is_data_slice, sigma)
 
             # Interweave and concatenate
             multi_features = self._extract_multi_features(
-                bands_slice, is_data_slice, local_avg, local_var, rel_tdx, masked_bands, sigma
+                bands_slice, is_data_slice, local_avg, local_var, t_idx - left, masked_bands, sigma
             )
 
-            multi_proba[tdx * img_size : (tdx + 1) * img_size] = self._run_prediction(
+            multi_proba[t_idx * img_size : (t_idx + 1) * img_size] = self._run_prediction(
                 self.multi_classifier, multi_features
             )
 
-            prev_min_tdx = min_tdx
-            prev_max_tdx = max_tdx
+            prev_left = left
+            prev_right = right
 
         return multi_proba[..., None]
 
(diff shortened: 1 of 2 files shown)


mask/eolearn/mask/cloud_mask.py
```diff
@@ -329,32 +312,40 @@ def _dilate_all(self, data):
...
         return data
 
-    def _ssim_stats(self, bands, is_data, win_avg, var, nt_rel, sigma):
+    def _ssim_stats(
+        self,
+        bands: np.ndarray,
+        is_data: np.ndarray,
+        local_avg: np.ndarray,
+        local_var: np.ndarray,
+        rel_tdx: int,
```
zigaLuksic, line 321: what exactly is a `rel_tdx`?

mask/eolearn/mask/cloud_mask.py
```diff
@@ -329,32 +312,40 @@ def _dilate_all(self, data):
 
         return data
 
-    def _ssim_stats(self, bands, is_data, win_avg, var, nt_rel, sigma):
+    def _ssim_stats(
+        self,
+        bands: np.ndarray,
+        is_data: np.ndarray,
+        local_avg: np.ndarray,
+        local_var: np.ndarray,
```
zigaLuksic, line 320: would sure be nice to know what these two actually area. Perhaps not the time (since we're still moving stuff around), but at some point we'll need to add docstrings

mask/eolearn/mask/cloud_mask.py
```diff
@@ -389,96 +380,103 @@ def _do_single_temporal_cloud_detection(self, bands):
...
-        loc_mu = None
-        loc_var = None
-
-        prev_nt_min = None
-        prev_nt_max = None
+        local_avg, local_var = None, None
+        prev_min_tdx, prev_max_tdx = None, None
 
-        for t_i in range(n_times):
+        for tdx in range(n_times):
```
zigaLuksic, line 395: what is a tdx? a time index? the loop body is long, and this seems like an important part of it, a longer name seems appropriate

mask/eolearn/mask/cloud_mask.py
```diff
@@ -389,96 +380,105 @@ def _do_single_temporal_cloud_detection(self, bands):
...
 
-        else:
-            win_avg_bands = self._map_sequence(bands_t[-1][None, ...], partial(self._win_avg, sigma=sigma))
-            win_avg_is_data = self._map_sequence(is_data_t[-1][None, ...], partial(self._win_avg, sigma=sigma))
+        avg_data = self._map_sequence(data, local_avg_func)
+        avg_data_mask = self._map_sequence(data_mask, local_avg_func)
+        avg_data_mask[avg_data_mask == 0.0] = 1.0
 
-            win_avg_is_data[win_avg_is_data == 0.0] = 1.0
+        var_data = self._map_sequence(data, local_var_func)
```
mlubej, line 440: This part confuses me. I don't understand why you would need to divide with `avg_data_mask`. IIRC, this is a value between 0 and 1 for each time and for each band. In case the value is 0, it's manually set to 1 (because of the division error).

I could see this as a sort of weighting, but I'm not sure why you would increase the values of `avg_data` by dividing with something smaller than 1.

Additionally, why isn't the same applied for the `var_data` ?

My intuition tells me that this should be dropped, but it's risky at this point...

zigaLuksic (commented):
gonna trust you on a lot of this :)

mask/eolearn/mask/cloud_mask.py
zigaLuksic: we'll probably need to do some heavier refactoring before this becomes a bit clearer.

mask/eolearn/mask/cloud_mask.py
mlubej: `idx` is standard for an index, so I opted for `tdx` as sort of a time index. I can make it longer, but it's also used as an appendix in a lot of variables.

mask/eolearn/mask/cloud_mask.py
mlubej: since `tdx` is a time index (see below), `rel_tdx` seemed like a better choice to represent the `tdx` relative to the `min_tdx`. At least compared to `rel_t_i`   ¯\\\_(ツ)\_/¯

zigaLuksic: i never would've guessed it was relative to `min_tdx` :see_no_evil:
Lets leave it as `rel_<whatever>` and maybe we can fix it later.

mask/eolearn/mask/cloud_mask.py
mlubej: - renamed to `t_idx`
- used `left` and `right` for window min/max indices, i think it makes more sense. if not, we can discuss better names in a live session

mask/eolearn/mask/cloud_mask.py
mlubej: For now I actually removed it from this place and added it to the single function (where it's used) as a `local_t_idx`, which probably makes more sense, since everything `local_` is windows specific.

Anyway, we can discuss better names in in a live session, as mentioned below.

zigaLuksic (approved):

PR #564, opened 2023-02-17
Fix types in CloudMaskTask

zigaLuksic:
Highlights a lot of problematic areas in the code.

### DIFF
--- a/mask/eolearn/mask/cloud_mask.py
+++ b/mask/eolearn/mask/cloud_mask.py
@@ -13,7 +13,7 @@ file in the root directory of this source tree.
 import logging
 import os
 from functools import partial
-from typing import Optional, Tuple, Union
+from typing import Optional, Tuple, Union, cast
 
 import cv2
 import numpy as np
@@ -23,7 +23,7 @@ from typing_extensions import Protocol
 
 from sentinelhub import BBox, bbox_to_resolution
 
-from eolearn.core import EOTask, FeatureType, execute_with_mp_lock
+from eolearn.core import EOPatch, EOTask, FeatureType, execute_with_mp_lock
 
 from .utils import map_over_axis, resize_images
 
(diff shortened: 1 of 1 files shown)


mask/eolearn/mask/cloud_mask.py
```diff
@@ -385,12 +384,14 @@ def _do_single_temporal_cloud_detection(self, bands):
...
         """Performs a cloud detection process on multiple scenes at once"""
         multi_proba = np.empty(np.prod(bands.shape[:-1]))
-        img_size = np.prod(bands.shape[1:-1])
+        img_size = int(np.prod(bands.shape[1:-1]))
 
         n_times = bands.shape[0]
 
-        local_avg, local_var = None, None
-        prev_left, prev_right = None, None
+        local_avg: Optional[np.ndarray] = None
```
mlubej, line 391: What are the pros/cons of changing the type to include the Optional nature here, instead of in the function definition?

mask/eolearn/mask/cloud_mask.py
```diff
@@ -438,9 +446,11 @@ def _update_batches(
 
         var_data = self._map_sequence(data, local_var_func)
 
-        if local_avg is None:
+        if local_avg is None or local_var is None:
             local_avg = avg_data / avg_data_mask
+            local_avg = cast(np.ndarray, local_avg)
```
mlubej, line 451: Is this needed to drop the optional part?

mask/eolearn/mask/cloud_mask.py
zigaLuksic: could you elaborate?

mask/eolearn/mask/cloud_mask.py
zigaLuksic: yeah, because we are reusing names. And `local_avg` is `Optional[np.ndarray]` (because the condition is `or`) and `mypy` doesn't want to specialize that. We can avoid that by using different names. I didnt feel like thinking about that too hard, because the whole procedure is trash.

mask/eolearn/mask/cloud_mask.py
mlubej: nevermind, I'm blind.

mlubej (approved):

PR #566, opened 2023-02-17
Copying always copies the BBox

zigaLuksic:
Since the BBox is the spatial definition of the EOPatch it should always be copied.

### DIFF
--- a/core/eolearn/core/eodata.py
+++ b/core/eolearn/core/eodata.py
@@ -502,7 +502,7 @@ class EOPatch:
         :param features: Features to be copied into a new `EOPatch`. By default, all features will be copied. Note that
             `BBOX` is always copied.
         :param deep: If `True` it will make a deep copy of all data inside the `EOPatch`. Otherwise, only a shallow copy
-            of `EOPatch` will be made. Note that `BBOX` and `TIMESTAMP` will be deepcopied even with a shallow copy.
+            of `EOPatch` will be made. Note that `BBOX` and `TIMESTAMP` will be copied even with a shallow copy.
         :return: An EOPatch copy.
         """
         if deep:
--- a/core/eolearn/core/eodata.py
+++ b/core/eolearn/core/eodata.py
@@ -461,7 +461,7 @@ class EOPatch:
         if not features:  # For some reason deepcopy and copy pass {} by default
             features = ...
 
-        new_eopatch = EOPatch()
+        new_eopatch = EOPatch(bbox=copy.copy(self.bbox))
         for feature_type, feature_name in parse_features(features, eopatch=self):
             if feature_type.has_dict():
                 new_eopatch[feature_type][feature_name] = self[feature_type].__getitem__(feature_name, load=False)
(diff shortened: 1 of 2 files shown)


core/eolearn/core/eodata.py
```diff
@@ -461,7 +461,7 @@ def __copy__(self, features: FeaturesSpecification = ...) -> EOPatch:
         if not features:  # For some reason deepcopy and copy pass {} by default
             features = ...
 
-        new_eopatch = EOPatch()
+        new_eopatch = EOPatch(bbox=copy.copy(self.bbox))
```
mlubej, line 464: potential idea if we are going to be doing this a lot, it would perhaps make sense that we add a `.copy(deep: bool = False)` method to the `BBox` class, perhaps also the `Geometry`.

If this is more of a one and done thing, it's fine as is.

core/eolearn/core/eodata.py
```diff
@@ -499,7 +499,8 @@ def __deepcopy__(self, memo: Optional[dict] = None, features: FeaturesSpecificat
     def copy(self, features: FeaturesSpecification = ..., deep: bool = False) -> EOPatch:
         """Get a copy of the current `EOPatch`.
 
-        :param features: Features to be copied into a new `EOPatch`. By default, all features will be copied.
+        :param features: Features to be copied into a new `EOPatch`. By default, all features will be copied. Note that
+            `BBOX` is always copied.
```
mlubej, line 503: When we reach the point where stop referring to the BBox as one of the features, this should also be removed.  Perhaps it's safer to remove the additional comment due to this?

zigaLuksic: It should be noted somewhere. I can move it to the general body of the docstring

core/eolearn/core/eodata.py
zigaLuksic: i what custom behaviour would you want there that is not provided by `copy.copy`?
I think a better question is if `EOPatch` actually needs custom implementations of `__copy__` and `__deepcopy__` or if the default behaviour is good enough. The `FeatureIO` objects complicate stuff a bit i believe.

core/eolearn/core/eodata.py
zigaLuksic: I tried moving it into the body so that it was "Get a copy of the current `EOPatch`. The `bbox` is always copied.", but i'm not sure it's much better

core/eolearn/core/eodata.py
mlubej: You can ignore my initial comment, but can you explain how do the custom implementations of `__copy__` and `__deepcopy__` differ from the default ones?

core/eolearn/core/eodata.py
mlubej: I mean't it more in the direction if it even makes sense in the first place to add this comment, since this would hold by default as a consequence of the assumption that we cannot have an EOPatch without a bbox. Or am I again looking too far ahead?

mlubej: Nevermind, I see that this comment is already in line with the future changes. Resolving.

core/eolearn/core/eodata.py
zigaLuksic: The main difference is that we ensure that `FeatureIO` objects are not loaded, but also not deep-copied (because they have a filesystem attribute :/ copying uses pickling behind the scenes), so we do a workaround for that

mlubej (approved):

PR #567, opened 2023-02-17
Update copy test

jgersak:
combine test fore copy and deepcopy
add fixture to create smaller EOPatches from fixture Patch 
use deep_eq for EOPatch comperison


core/eolearn/tests/test_core_tasks.py
```diff
@@ -71,38 +73,34 @@ def patch_fixture():
...
+    "features, expected_patch",
+    [
+        ([(FeatureType.MASK_TIMELESS, "mask"), FeatureType.BBOX], ["mask_timeless", "bbox"]),
+        ([FeatureType.TIMESTAMP, (FeatureType.SCALAR, "values"), FeatureType.BBOX], ["scalar", "timestamp", "bbox"]),
+    ],
+    indirect=["expected_patch"],
+)
+def test_partial_copy(features: List[FeatureSpec], patch: EOPatch, expected_patch: EOPatch) -> None:
+    partial_copy = DeepCopyTask(features=features).execute(patch)
+    assert deep_eq(partial_copy, expected_patch)
```
zigaLuksic, line 103: this is nearly 10 lines of extra code and also less readable :thinking:

jgersak: Planing to use this fixture more other tests, if there would be like 10 test cases it would be less code mark.parametrize is boilerplate code heavy in this case still prefer having two DRY test cases instead of "long" test.

core/eolearn/tests/test_core_tasks.py
```diff
@@ -71,38 +73,34 @@ def patch_fixture():
...
-    patch_copy = CopyTask().execute(patch)
-
-    assert patch == patch_copy, "Copied patch is different"
-
-    patch_copy.data["new"] = np.arange(1).reshape(1, 1, 1, 1)
-    assert "new" not in patch.data, "Dictionary of features was not copied"
+@pytest.mark.parametrize("task", [DeepCopyTask, CopyTask])
+def test_copy(task: CopyTask, patch: EOPatch) -> None:
+    patch_copy = task().execute(patch)
+    assert deep_eq(patch_copy, patch)
```
zigaLuksic, line 79: also why do you compare them with `deep_eq` instead of `==`?

jgersak: `patch.data["bands"]` is np.array usually when compere two np.arrays `numpy.testing.assert_array_equal` is used not `==` applied same logic here might be wrong :thinking:

core/eolearn/tests/test_core_tasks.py
zigaLuksic: speaking of DRY, you are repeating yourself in every test case :)
it also seems like there could be a problem, the way you constructed your `expected_patch_fixture` the whole `data_timeless` gets copied. Is there only a single feature in there? If so, select a different feature type, one with multiple features.

core/eolearn/tests/test_core_tasks.py
zigaLuksic: EOPatches are not numpy arrays. You should check how they implement equality

core/eolearn/tests/test_core_tasks.py
```diff
@@ -71,38 +72,33 @@ def patch_fixture():
...
+@pytest.mark.parametrize(
+    "features, expected_patch",
+    [
+        ([(FeatureType.MASK_TIMELESS, "mask"), FeatureType.BBOX], ["mask_timeless", "bbox"]),
+        ([FeatureType.TIMESTAMP, (FeatureType.SCALAR, "values")], ["scalar", "timestamp"]),
+    ],
+    indirect=["expected_patch"],
+)
+def test_partial_copy(features: List[FeatureSpec], patch: EOPatch, expected_patch: EOPatch) -> None:
+    assert DeepCopyTask(features=features).execute(patch) == expected_patch
```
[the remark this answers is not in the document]
mlubej: Ah, I see I was too hasty, the features need to be taken from the `patch` fixture. Then I guess we just need to change the line where `patch_expected` is defined into:

```python
patch_expected = EOPatch(**{feature_type: patch[feature] for feature_type, feature in features_dict.items()})
```

core/eolearn/tests/test_core_tasks.py
```diff
@@ -71,38 +72,39 @@ def patch_fixture():
     return patch
```
mlubej, line 72: BTW, I noticed there is already an existing fixture for core tests [here](https://github.com/sentinel-hub/eo-learn/blob/feat/rename-timestamp/core/eolearn/tests/conftest.py#L20). Any idea why a dedicated fixture is used here?

jgersak: `TestEOPatch`is kinda big and loading it 40 times should be that quick but i might be wrong going to "time" test it after other refactoring if this is ok ?

core/eolearn/tests/test_core_tasks.py
```diff
@@ -71,38 +72,33 @@ def patch_fixture():
...
-def test_deepcopy(patch):
-    patch_deepcopy = DeepCopyTask().execute(patch)
-
-    assert patch == patch_deepcopy, "Deep copied patch is different"
+    assert (patch_copy != patch) if task == DeepCopyTask else (patch_copy == patch)
 
-    patch_deepcopy.data["new"] = np.arange(1).reshape(1, 1, 1, 1)
-    assert "new" not in patch.data, "Dictionary of features was not copied"
+    patch_copy.data["new"] = np.arange(1).reshape(1, 1, 1, 1)
+    assert "new" not in patch.data
```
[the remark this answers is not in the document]
jgersak: when u do copy addresses of all variables are assigned to their copies patch_copy.data["new"] is new variable create after copy is called so it shoud not be copyed unless the copy were called one level up (for example if you would copy list of EOPatch-es)

core/eolearn/tests/test_core_tasks.py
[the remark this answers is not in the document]
jgersak: Code was updated is this still relevant ?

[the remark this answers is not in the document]
mlubej: @zigaLuksic how do you feel about the suggestion in this thread? In order to compare all the features a new EOPatch can be created and the eopatch comparison is used, instead of looping over features and doing basically the same thing as the eopatch comparison does behind the scenes. The latter is currently implemented, but I would suggest the former since it looks cleaner, but I'm not aware of any potential performance drops or anything in that direction.

core/eolearn/tests/test_core_tasks.py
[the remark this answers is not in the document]
zigaLuksic: in order for 'eopatches' to mimic behaviour of `dict`, the copying ensures that you get a new 'dict' with the pointers to old data. So if you update the old data in the first dict, it should be mimicked in the second one. But if you add/remove data from either of the two dicts, the other should not change.

core/eolearn/tests/test_core_tasks.py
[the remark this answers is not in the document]
zigaLuksic: after talking with @mlubej we decided that this is a better approach (as it does not use an ad-hoc implementation).

However it can be simplified. No reason to iterate and compare features. The features can be extracted with `get_features()` and compared as subsets. Then all existing features are iterated over and checked that they match (as is done now).

core/eolearn/tests/test_core_tasks.py
zigaLuksic: In the long run it would be great if we get rid of the test eopatch as it is a lot of data on the repo.
Here we don't much care about what data is in, only that it is populated. Although it could be improved a bit (some floats, some random etc.)

What i would like to point out is that this EOPatch only has a single feature per type, which should be adjusted a bit. This can be done manually for the sake of this MR, but in the long run we would want a utility for generating 'random-ish' patches.

core/eolearn/tests/test_core_tasks.py
```diff
@@ -71,38 +72,39 @@ def patch_fixture():
...
-    assert not np.array_equal(patch.data["bands"], patch_deepcopy.data["bands"]), "Data should be copied"
+@pytest.mark.parametrize(
+    "features",
+    [
+        [(FeatureType.MASK_TIMELESS, "mask"), (FeatureType.BBOX, None)],
+        [(FeatureType.TIMESTAMP, None), (FeatureType.SCALAR, "values")],
+    ],
+)
+def test_partial_copy(features: List[FeatureSpec], patch: EOPatch) -> None:
+    patch_copy = DeepCopyTask(features=features)(patch)
```
zigaLuksic, line 95: parametrize with the task type (so that CopyTask is also tested) as discussed previous week

core/eolearn/tests/test_core_tasks.py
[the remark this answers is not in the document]
mlubej: thank you both, makes sense

core/eolearn/tests/test_core_tasks.py
jgersak: Adjusted a bit.

core/eolearn/tests/test_core_tasks.py
jgersak: Updated.

zigaLuksic: another note, the type of `CopyTask` is not `CopyTask` but `Type`. You can use `Type[CopyTask]` to indicate it has to be a `CopyTask` or it's subclass.

jgersak: done

core/eolearn/tests/test_core_tasks.py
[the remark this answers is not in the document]
jgersak: done

PR #569, opened 2023-02-20
Test add rename remove

jgersak:


core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,69 @@ def test_io_task_pickling(filesystem, task_class):
...
+)
+def test_add_bbox_timestemps(feature: FeatureSpec, feature_data: Union[np.ndarray], patch: EOPatch) -> None:
+    assert patch[feature] != feature_data
+    patch = AddFeatureTask(feature)(patch, feature_data)
+    assert patch[feature] == feature_data
+
+
+def test_rename_feature(patch: EOPatch) -> None:
+    feature_name = "bands"
+    new_feature_name = "data"
```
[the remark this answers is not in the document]
mlubej: Additionally, I think I would prefer to see whole features here, not just names. It looks better when you call these features below, instead of hardcoding `patch.data` and such.

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,69 @@ def test_io_task_pickling(filesystem, task_class):
...
+
+    with pytest.raises(KeyError):
+        patch.data[new_feature_name]
+    patch_copy = copy.deepcopy(patch)
+
+    patch = RenameFeatureTask((FeatureType.DATA, feature_name, new_feature_name))(patch)
+    assert np.array_equal(patch.data[new_feature_name], patch_copy.data[feature_name])
+    assert feature_name not in patch[FeatureType.DATA]
```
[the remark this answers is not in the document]
jgersak: Imo it is not needed, because feature_name is from the fixture and we probably should not test pytest fixtuers init in test

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,69 @@ def test_io_task_pickling(filesystem, task_class):
...
+    with pytest.raises(KeyError):
+        patch[feature]
+
+    del patch_copy[feature]
+    assert patch == patch_copy
+
+
+@pytest.mark.parametrize("feature", [(FeatureType.BBOX, None), (FeatureType.TIMESTAMP, None)])
+def test_remove_bbox_timestamp(feature: FeatureSpec, patch: EOPatch) -> None:
+    # this test should fail for bbox and timestamps after rework
```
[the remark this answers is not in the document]
jgersak: Rework will show, i hope both.

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,68 @@ def test_io_task_pickling(filesystem, task_class):
...
+
+    patch = RenameFeatureTask((f_type, f_name, f_new_name))(patch)
+    assert np.array_equal(patch[(f_type, f_new_name)], patch_copy[(f_type, f_name)])
+    assert f_name not in patch[f_type]
+
+
+def test_remove_feature(patch: EOPatch) -> None:
+    feature = (FeatureType.DATA, "bands")
+    patch_copy = copy.deepcopy(patch)
+    assert len(patch[feature]) != 0
```
mlubej, line 193: what are you really trying to test here exactly? It's not so clear from the assertion itself.

jgersak: usual we can test with assert object if something  exist (is not None) u can not do this kind of test for np.array this test is not the best so i m open for code improvement :)

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,69 @@ def test_io_task_pickling(filesystem, task_class):
...
-    assert feature_name not in patch.mask, "Feature was not removed"
+@pytest.mark.parametrize(
+    "feature, feature_data",
+    [
+        ((FeatureType.BBOX, None), BBox((24.54, 56.45, 95.4, 13.43), CRS(3857))),
+        ((FeatureType.TIMESTAMP, None), [datetime(2022, 1, 1, 10, 4, 7), datetime(2022, 1, 4, 10, 14, 5)]),
+    ],
+)
+def test_add_bbox_timestemps(feature: FeatureSpec, feature_data: Union[np.ndarray], patch: EOPatch) -> None:
+    assert patch[feature] != feature_data
```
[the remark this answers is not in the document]
jgersak: Done and added.

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,69 @@ def test_io_task_pickling(filesystem, task_class):
     assert isinstance(unpickled_task, task_class)
 
 
-def test_add_rename_remove_feature(patch):
+def test_add_feature(patch: EOPatch) -> None:
     cloud_mask = np.arange(10).reshape(5, 2, 1, 1)
     feature_name = "CLOUD MASK"
```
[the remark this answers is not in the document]
jgersak: Done

core/eolearn/tests/test_core_tasks.py
mlubej: if the idea is only to test if the feature exists, you could just check with `assert feature in patch`, right?

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,69 @@ def test_io_task_pickling(filesystem, task_class):
     assert isinstance(unpickled_task, task_class)
 
 
-def test_add_rename_remove_feature(patch):
+def test_add_feature(patch: EOPatch) -> None:
     cloud_mask = np.arange(10).reshape(5, 2, 1, 1)
```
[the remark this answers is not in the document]
jgersak: Void if you want i can update test, atm eolearn does not check if added feat if no timelees have same dim as timestamps

core/eolearn/tests/test_core_tasks.py
jgersak: Yes, tnx done.

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,68 @@ def test_io_task_pickling(filesystem, task_class):
...
+    "feature, feature_data",
+    [
+        ((FeatureType.BBOX, None), BBox((24.54, 56.45, 95.4, 13.43), CRS(3857))),
+        ((FeatureType.TIMESTAMP, None), [datetime(2022, 1, 1, 10, 4, 7), datetime(2022, 1, 4, 10, 14, 5)]),
+    ],
+)
+def test_add_bbox_timestamps(feature: FeatureSpec, feature_data: np.ndarray) -> None:
+    # this test should fail for bbox and timestamps after rework
+    patch = EOPatch()
+    assert not patch[feature]
```
mlubej, line 173: I didn't know that you could test feature existence for bbox and timestams like this. TBH it feels wrong and scary, since I would expect this to crash due to the features not being present. But I see that in case of bbox and timestamps it really doesn't crash because they are [handled differently](https://github.com/sentinel-hub/eo-learn/blob/fix/readthedocs-links-not-working/core/eolearn/core/eodata.py#L250-L251):

```
...
bbox: Optional[BBox] = attr.ib(default=None)
timestamp: List[dt.datetime] = attr.ib(factory=list)
...
```

I would still much rather prefer to see `assert not feature in patch` used here.

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,68 @@ def test_io_task_pickling(filesystem, task_class):
...
-    patch = RemoveFeatureTask((FeatureType.MASK, new_feature_name))(patch)
-    assert feature_name not in patch.mask, "Feature was not removed"
+@pytest.mark.parametrize(
+    "feature, feature_data",
+    [
+        ((FeatureType.BBOX, None), BBox((24.54, 56.45, 95.4, 13.43), CRS(3857))),
+        ((FeatureType.TIMESTAMP, None), [datetime(2022, 1, 1, 10, 4, 7), datetime(2022, 1, 4, 10, 14, 5)]),
+    ],
+)
+def test_add_bbox_timestamps(feature: FeatureSpec, feature_data: np.ndarray) -> None:
```
mlubej, line 170: I don't see any reason why the `add` functionality should be split for bbox/timestamps and for other features. Couldn't you add a new parametrization entry for the cloud mask from L153 here? The logic is the same:
- create empty patch
- check that the feature doesn't exist
- add new feature with feature data
- check if feature of eopatch is the same as feature data
The same assertion (`assert patch[feature] == feature_data`) should work, no need to use `np` specific checks.

mlubej: same for testing the remove functionality

core/eolearn/tests/test_core_tasks.py
jgersak: done

core/eolearn/tests/test_core_tasks.py
jgersak: updated

core/eolearn/tests/test_core_tasks.py
```diff
@@ -11,16 +11,16 @@
...
+from datetime import datetime
 from typing import Dict, Iterable, Tuple, Union
 
 import numpy as np
 import pytest
 from fs.osfs import OSFS
 from fs.tempfs import TempFS
 from fs_s3fs import S3FS
-from numpy.testing import assert_equal
+from numpy.testing import assert_array_equal, assert_equal
```
mlubej, line 23: What is the difference between these two? I see both used in the same context below

jgersak: `assert_array_equal` expected inputs are np.array, `assert_equal` can take (scalars, lists, tuples, dictionaries or numpy arrays), `assert_array_equal` looks faster due to less checks constraints and numpy operations

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,49 @@ def test_io_task_pickling(filesystem, task_class):
...
+    "feature, feature_data",
+    [
+        ((FeatureType.MASK, "CLOUD MASK"), np.arange(10).reshape(5, 2, 1, 1)),
+        ((FeatureType.BBOX, None), BBox((24.54, 56.45, 95.4, 13.43), CRS(3857))),
+        ((FeatureType.TIMESTAMP, None), [datetime(2022, 1, 1, 10, 4, 7), datetime(2022, 1, 4, 10, 14, 5)]),
+    ],
+)
+def test_add_feature(feature: FeatureSpec, feature_data: np.ndarray) -> None:
+    # this test should fail for bbox and timestamps after rework
+    patch = EOPatch()
```
zigaLuksic, line 162: assume that `EOPatch` objects cannot be instantiated without a `bbox` (as that will be deprecated soon)

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,49 @@ def test_io_task_pickling(filesystem, task_class):
...
 
-def test_add_rename_remove_feature(patch):
-    cloud_mask = np.arange(10).reshape(5, 2, 1, 1)
-    feature_name = "CLOUD MASK"
-    new_feature_name = "CLM"
+@pytest.mark.parametrize(
+    "feature, feature_data",
+    [
+        ((FeatureType.MASK, "CLOUD MASK"), np.arange(10).reshape(5, 2, 1, 1)),
+        ((FeatureType.BBOX, None), BBox((24.54, 56.45, 95.4, 13.43), CRS(3857))),
```
zigaLuksic, line 156: replace with something else, perhaps `FeatureType.META_INFO`

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,49 @@ def test_io_task_pickling(filesystem, task_class):
...
+def test_rename_feature(patch: EOPatch) -> None:
+    f_type, f_name, f_new_name = FeatureType.DATA, "bands", "new_bands"
+    assert (f_type, f_new_name) not in patch
+    patch_copy = copy.deepcopy(patch)
 
-    patch = RemoveFeatureTask((FeatureType.MASK, new_feature_name))(patch)
-    assert feature_name not in patch.mask, "Feature was not removed"
+    patch = RenameFeatureTask((f_type, f_name, f_new_name))(patch)
+    assert_array_equal(patch[(f_type, f_new_name)], patch_copy[(f_type, f_name)])
+    assert (f_type, f_name) not in patch
```
zigaLuksic, line 179: add description, so that it is clear that the problem is that the input patch was changed

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +149,49 @@ def test_io_task_pickling(filesystem, task_class):
...
+
+
+@pytest.mark.parametrize(
+    "feature", [(FeatureType.DATA, "bands"), (FeatureType.BBOX, None), (FeatureType.TIMESTAMP, None)]
```
zigaLuksic, line 183: is there a reason why you're adding a case for `BBOX` if we know it won't be allowed to happen? Wouldn't it be better to add a test where this fails, and mark it for skipping for now?

Also open an issue and link it to the existing one, so that we keep track of ensuring that this task cannot remove `BBOX`

jgersak: Done as suggested.

core/eolearn/tests/test_core_tasks.py
jgersak: done

core/eolearn/tests/test_core_tasks.py
jgersak: done

core/eolearn/tests/test_core_tasks.py
jgersak: done

zigaLuksic (approved):

core/eolearn/tests/test_core_tasks.py
```diff
@@ -148,28 +151,54 @@ def test_io_task_pickling(filesystem, task_class):
...
 
-    patch = AddFeatureTask((FeatureType.MASK, feature_name))(patch, cloud_mask)
-    assert np.array_equal(patch.mask[feature_name], cloud_mask), "Feature was not added"
 
-    patch = RenameFeatureTask((FeatureType.MASK, feature_name, new_feature_name))(patch)
-    assert np.array_equal(patch.mask[new_feature_name], cloud_mask), "Feature was not renamed"
-    assert feature_name not in patch[FeatureType.MASK], "Old feature still exists"
+@pytest.mark.parametrize("feature", [(FeatureType.DATA, "bands"), (FeatureType.TIMESTAMP, None)])
+def test_remove_feature(feature: FeatureSpec, patch: EOPatch) -> None:
+    # this test should fail for bbox and timestamps after rework
```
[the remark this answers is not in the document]
jgersak: done

Issue #497, opened 2022-11-09
readthedocs links not working


mlubej:
Some links in readme.md files aren't working, i.e. in:
- https://github.com/sentinel-hub/eo-learn/tree/master/core
- https://github.com/sentinel-hub/eo-learn/tree/master/coregistration

They lead to

<img width="822" alt="image" src="https://user-images.githubusercontent.com/37869250/200830325-cac818d5-eb85-4069-9d64-c78c5b71f8ff.png">

zigaLuksic:
oh damn, those haven't been looked at in ages
i'm not sure we even should have READMEs there :thinking:

PR #570, opened 2023-02-20
Fix/readthedocs links not working

mlubej:
Changes
- update the links to be the same as in case of https://eo-learn.readthedocs.io/en/latest/eotasks.html#eotasks
- remove link in the `visualization` readme since it's not listed in the link above
- in some cases there was an issue with the link rendering due to cases of  "[`", I added a new line to the readme processor
- formatting of readme files (whitespaces, newlines, ..)

@zigaLuksic the links are not displayed on PyPI, however, I think they fit nicely under the subpackages, since the user can simply be pointed to the specific place in the docs. If you think that this shouldn't be the case and the docs are standalone as they are (with no external links pointing to any specific places), then I'll remove the links as well.

Closes https://github.com/sentinel-hub/eo-learn/issues/497


core/README.md
```diff
@@ -2,4 +2,4 @@
 
 The main subpackage which implements basic building blocks (EOPatch, EOTask and EOWorkflow) and commonly used functionalities.
 
-For more information on the package content, visit [readthedocs](https://eo-learn.readthedocs.io/en/latest/eolearn.core.html).
+For more information on the package content, visit [readthedocs](https://eo-learn.readthedocs.io/en/latest/eotasks.html#core).
```
zigaLuksic, line 5: should we have the links to the tasks? Or do we add a link to the root of the docs for the module, e.g. https://eo-learn.readthedocs.io/en/latest/reference/eolearn.core.html

docs/source/conf.py
```diff
@@ -275,6 +275,7 @@ def process_readme():
         readme = file.read()
 
     readme = readme.replace("# eo-learn", "# Introduction").replace("docs/source/", "")
+    readme = readme.replace("[`", "[").replace("`]", "]")
```
zigaLuksic, line 278: why do we do this?

mlubej: There are issues in some of the docs, namely here: https://eo-learn.readthedocs.io/en/latest/#more-examples

image

It is because backticks are used in the url name template. This removes the backticks. It works in markdown, but not in the linked docs. We could just remove the backticks, but I noticed the same rule is applied also in `eo-grow`.

zigaLuksic: ah, i see. consider this resolved

core/README.md
mlubej: We could even do both. If you're reading on the docs you can conveniently jump to the codebase, and if you're on git you can conveniently jump to the relevant docs/tasks.

I can add this if you agree with the above.

zigaLuksic: So it would be something along the lines of
```
For more information about the module visit [readthedocs](blabla.html), or see the [module task index](blaeotasks.html#core)
```

That sounds nice

mlubej: done

visualization/README.md
```diff
@@ -1,5 +1,3 @@
 # eolearn.visualization subpackage
 
 A collection of data visualization utilities.
-
-For more information on the package content, visit [readthedocs](https://eo-learn.readthedocs.io/en/latest/eolearn.visualization.html).
```
zigaLuksic, line 5: forgot this one ;)

mlubej: Added only the link for the module content, since no tasks link exists.

zigaLuksic (approved):

PR #573, opened 2023-02-21
Rename EOPatch attribute timestamp -> timestamps

mlubej:
Changes:
- rename the `timestamp` attribute to `timestamps`
- add `timestamp` property which triggers an `EODeprecationWarning`
- replace usage of `timestamp` to `timestamps` across the code, tests, notebooks and docs
- rename value of `FeatureType.TIMESTAMP` from `timestamp` to `timestamps`
- had to update some functionalities so that it works after the change (added a conversion + deprecation)
  - loading the eopatch (timestamps saved under `timestamp.json`)
  - for the vector feature the `TIMESTAMP` column name was kept, but a global variable was created and it's usage implemented throughout the code


mlubej:
@zigaLuksic turns out that running tests doesn't trigger the `EODeprecation` warning, so I wasn't able to check if it pops up this way, but I think I covered the switch of all the old usages of `.timestamp`

mlubej:
> @zigaLuksic turns out that running tests doesn't trigger the `EODeprecation` warning, so I wasn't able to check if it pops up this way, but I think I covered the switch of all the old usages of `.timestamp`

Nevermind, it seems that these are present in the logs here

core/eolearn/core/eodata.py
```diff
@@ -185,11 +187,23 @@ def _parse_feature_value(self, value: object, feature_name: str) -> gpd.GeoDataF
...
-                raise ValueError(
-                    f"{self.feature_type} feature has to contain a column 'TIMESTAMP' with timestamps but "
-                    f"feature {feature_name} doesn't not have it."
-                )
+            if self.feature_type is FeatureType.VECTOR:
+                if "TIMESTAMP" in value:
+                    warn(
+                        (
+                            f"{self.feature_type} feature has to contain a column 'TIMESTAMPS' instead of 'TIMESTAMP'"
+                            " (deprecated)"
```
zigaLuksic, line 195:
```suggestion
                            f"{self.feature_type} feature has to contain a column 'TIMESTAMPS' instead of 'TIMESTAMP'."
                            " In the future the old name will no longer be supported."
```

core/eolearn/core/eodata_io.py
```diff
@@ -219,6 +220,17 @@ def walk_main_folder(filesystem: FS, folder_path: str) -> Iterator[Tuple[Feature
         else:
             ftype_str, fname = raw_path, ...
 
+        if ftype_str == "timestamp":
+            warnings.warn(
+                (
+                    f"EOPatch at {filesystem.getsyspath(folder_path)} contains the deprecated `timestamp` feature. "
+                    "Save the EOPatch again to update it."
```
zigaLuksic, line 227:
```suggestion
                    f"EOPatch at {filesystem.getsyspath(folder_path)} contains the deprecated `timestamp` feature."
                    " The old name will no longer be valid in the future. You can re-save the `EOPatch` to update it."
```

Forgot it also affects files. This could be a bit annoying for a while :)

io/eolearn/io/extra/meteoblue.py
```diff
@@ -92,8 +92,10 @@ def _get_modified_eopatch(eopatch: Optional[EOPatch], bbox: Optional[BBox]) -> E
 
     def _prepare_time_intervals(self, eopatch, time_interval):
         """Prepare a list of time intervals for which data will be collected from meteoblue services"""
-        if not eopatch.timestamp and not time_interval:
-            raise ValueError("Time interval should either be defined with eopatch.timestamp of time_interval parameter")
+        if not eopatch.timestamps and not time_interval:
+            raise ValueError(
+                "Time interval should either be defined with eopatch.timestamps of time_interval parameter"
```
zigaLuksic, line 97:
```suggestion
                "Time interval should either be defined with `eopatch.timestamps` or `time_interval` parameter"
```

mlubej:
Pointed out by @batic, in case of VECTOR dataframes, the columns name should probably still be `TIMESTAMP` since the column represents row values...

After a thought I agree with this and I would suggest to revert the change for the case of vector features.

zigaLuksic:
> Pointed out by @batic, in case of VECTOR dataframes, the columns name should probably still be `TIMESTAMP` since the column represents row values...
> 
> After a thought I agree with this and I would suggest to revert the change for the case of vector features.

I know we are now far more careful with flexibility, but what if either option is ok? Would that be an acceptable evil?

mlubej:
> I know we are now far more careful with flexibility, but what if either option is ok? Would that be an acceptable evil?

It could be acceptable, but not sure what is the benefit, since the default for the dataframe is already intuitive (`TIMESTAMP`) and this would only be extending the flexibility, not increasing the backwards compatibility. I would vote for reverting back to `TIMESTAMP`. If this wasn't an issue so far, then it's not so problematic in the first place.

zigaLuksic:
> > I know we are now far more careful with flexibility, but what if either option is ok? Would that be an acceptable evil?
> 
> It could be acceptable, but not sure what is the benefit, since the default for the dataframe is already intuitive (`TIMESTAMP`) and this would only be extending the flexibility, not increasing the backwards compatibility. I would vote for reverting back to `TIMESTAMP`. If this wasn't an issue so far, then it's not so problematic in the first place.

My train of thought was that using `TIMESTAMPS` would be more in-line with the eopatch naming, but it's not really a 1:1 interface, so i guess keeping only `TIMESTAMP` is fine.

zigaLuksic:
Another thought. Should we also rename `FeatureType.TIMESTAMP` to `FeatureType.TIMESTAMPS`? Not sure if worth since we're trying to remove it in the future anyway. But might be the right thing to do :/

mlubej:
> Another thought. Should we also rename `FeatureType.TIMESTAMP` to `FeatureType.TIMESTAMPS`? Not sure if worth since we're trying to remove it in the future anyway. But might be the right thing to do :/

Yes, this is also something I pointed out in the original description (it's a bit hidden, sorry). I wanted to check with you first, I also think that it makes sense to do it, even though a change is expected.

core/eolearn/core/eodata.py
```diff
@@ -185,10 +187,10 @@ def _parse_feature_value(self, value: object, feature_name: str) -> gpd.GeoDataF
...
 
         if isinstance(value, gpd.GeoDataFrame):
-            if self.feature_type is FeatureType.VECTOR and FeatureType.TIMESTAMP.value.upper() not in value:
+            if self.feature_type is FeatureType.VECTOR and TIMESTAMP_COLUMN not in value:
                 raise ValueError(
-                    f"{self.feature_type} feature has to contain a column 'TIMESTAMP' with timestamps but "
-                    f"feature {feature_name} doesn't not have it."
+                    f"{self.feature_type} feature has to contain a column '{TIMESTAMP_COLUMN}' with timestamps but "
+                    f"feature {feature_name} does not not have it."
                 )
```
zigaLuksic, line 194: this will break any old EOPatches right?

mlubej: no, the `TIMESTAMP_COLUMN` value remained the same (`"TIMESTAMP"`) since it's more in line with the dataframe format.

zigaLuksic: right! thanks :see_no_evil:

zigaLuksic (approved):

mlubej:
I feel scared merging this, you pull the trigger :D

PR #577, opened 2023-02-22
Rename FeatureType.TIMESTAMP -> FeatureType.TIMESTAMPS

mlubej:
Changes:
- added a custom metaclass for a deprecated enum member (thanks @zigaLuksic)
- implemented the change throughout the code, docs and notebooks

Related to https://github.com/sentinel-hub/eo-learn/pull/573


core/eolearn/core/constants.py
```diff
@@ -10,15 +10,46 @@
...
+    deprecation_msg = None
+    if name in ("TIMESTAMP", "timestamp"):
+        deprecation_msg = (
+            "`FeatureType.TIMESTAMP` has been renamed to `FeatureType.TIMESTAMPS` and will be removed in a future"
+            " version. Please switch to using `FeatureType.TIMESTAMPS`."
+        )
+        name = "TIMESTAMPS" if name == "TIMESTAMP" else "timestamps"
+
+    if deprecation_msg:
+        warnings.warn(deprecation_msg, category=EODeprecationWarning, stacklevel=3)
```
zigaLuksic, line 35: as discussed in person, do we skip the deprecation warning and just autocorrect? Since it will be removed soon anyway

mlubej: Yes, we can skip the deprecation and just do the inclusion.

mlubej: warning removed, but I left the functionality there  for later (had to add a mypy ignore statement because the code was never reached)

mlubej:
Added also a quick test as per suggestion by @zigaLuksic

zigaLuksic (approved):

PR #581, opened 2023-02-23
Update test merge

jgersak:


core/eolearn/tests/test_core_tasks.py
```diff
@@ -55,8 +55,13 @@
 def patch_fixture():
     patch = EOPatch()
     patch.data["bands"] = np.arange(2 * 3 * 3 * 2).reshape(2, 3, 3, 2)
+    patch.data["CLP"] = np.arange(2 * 3 * 3 * 1).reshape(2, 3, 3, 1)
+    patch.mask["CLM"] = np.arange(2 * 3 * 3 * 1).reshape(2, 3, 3, 1)
```
zigaLuksic, line 59: now all of them are doubled :thinking:  and they all contain the same kind of data, so if CLM and CLP are mixed up, the test wont discover it. Use `np.full` and things like that, vary shapes a bit.

core/eolearn/tests/test_core_tasks.py
```diff
@@ -99,11 +104,15 @@ def test_deepcopy(patch):
 
 
 def test_partial_copy(patch):
-    partial_copy = DeepCopyTask(features=[(FeatureType.MASK_TIMELESS, "mask"), FeatureType.BBOX]).execute(patch)
+    partial_copy = DeepCopyTask(
+        features=[(FeatureType.MASK_TIMELESS, "mask"), (FeatureType.MASK_TIMELESS, "LULC"), FeatureType.BBOX]
```
zigaLuksic, line 108: here you copy both MASK_TIMELESS features and in the next case you copy both SCALAR features...

zigaLuksic: why are there changes if it's not part of this MR?

jgersak: old aka here curnt `test_partial_copy` in test copy FeatureType insted of FeatureSpec Fixtuer EOPatch were updated for `test_merge_features` i did minimal change posible on test copy that test are Passing 

if merge is going to use its owen Fixtuer EOPatch the change is going to be reverted on this branch

core/eolearn/tests/test_core_tasks.py
jgersak: Updated

zigaLuksic: much better :D
I would however have for instance 3 `mask_timeless` features
and maybe a single `mask` feature? :thinking:  because now there are always multiple features for each feature type

jgersak: Updated need to add instance in data so marge on other axis then `-1` can be tested.

zigaLuksic (approved):

PR #587, opened 2023-02-27
Update test_kwargs

jgersak:
replace test_kwargs with test_map_kwargs_passing and test_zip_kwargs_passing
update names test_zip_features according to current changes in test_map_features

### DIFF
--- a/core/eolearn/tests/test_core_tasks.py
+++ b/core/eolearn/tests/test_core_tasks.py
@@ -428,7 +428,7 @@ def test_map_kwargs_passing(input_feature: FeatureSpec, kwargs: Dict[str, Any], 
     mapped_patch = MapFeatureTask(input_feature, (FeatureType.META_INFO, "kwargs"), kwargs_map, **kwargs)(patch)
 
     expected_output = kwargs_map(mapped_patch[input_feature], **kwargs)
-    assert_array_equal(mapped_patch[(FeatureType.META_INFO, "kwargs")], expected_output)
+    assert mapped_patch[(FeatureType.META_INFO, "kwargs")] == expected_output
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


core/eolearn/tests/test_core_tasks.py
```diff
@@ -376,6 +401,53 @@ def test_map_features(test_eopatch):
...
+    ],
+)
+def test_map_kwargs_passing(input_feature: FeatureSpec, kwargs: Dict[str, Any], patch: EOPatch) -> None:
+    def kwargs_map(data, *, some=3, **kwargs) -> tuple:
+        return some, kwargs
+
+    mapped_patch = MapFeatureTask(input_feature, (FeatureType.META_INFO, "kwargs"), kwargs_map, **kwargs)(patch)
+
+    expected_output = kwargs_map(mapped_patch[input_feature], **kwargs)
+    assert_array_equal(mapped_patch[(FeatureType.META_INFO, "kwargs")], expected_output)
```
zigaLuksic, line 448: these two should be equal, not array equal.

it also makes no sense to keep this and the above test. This was suggested as a simplification of `test_map_kwargs`. 
If you do not like this test, then I believe you can merge `test_map` and `test_map_kwargs` since the tests are identical except for the `kwargs` parameter

jgersak: > these two should be equal, not array equal.

Done.

> it also makes no sense to keep this and the above test. This was suggested as a simplification of test_map_kwargs.
If you do not like this test, then I believe you can merge test_map and test_map_kwargs since the tests are identical except for the kwargs parameter

Agree, i think that "new" test kinda do not test if parameters were appealed to map function just if they were passed.  
Kinda done, parametrization look a bit "ugly".

zigaLuksic: the joint tests are ok enough, the parametrization is large but i guess it has to be.

PR #588, opened 2023-02-27
Refactor eodata_io (part 1)

zigaLuksic:
The refactorings are (or should be) entirely cosmetic.
I mostly extracted parts of code into helper functions, so that it is easier to see what happens when.

More refactoring will follow, but I want to keep it bite sized.

### DIFF
--- a/core/eolearn/core/eodata_io.py
+++ b/core/eolearn/core/eodata_io.py
@@ -81,52 +81,27 @@ def save_eopatch(
     """A utility function used by `EOPatch.save` method."""
     patch_exists = filesystem.exists(patch_location)
 
-    # if not patch_exists:
-    #     filesystem.makedirs(patch_location, recreate=True)
-
     eopatch_features = list(walk_eopatch(eopatch, patch_location, features))
     fs_features = list(walk_filesystem(filesystem, patch_location)) if patch_exists else []
 
     _check_collisions(overwrite_permission, eopatch_features, fs_features)
 
-    features_to_save: List[Tuple[Type[FeatureIO], Any, str]] = [
-        (FeatureIOBBox, eopatch.bbox, fs.path.combine(patch_location, BBOX_FILENAME))
-    ]
+    # Data must be collected before any tinkering with files due to lazy-loading
+    features_to_save = _prepare_features_to_save(eopatch, patch_location, eopatch_features)
 
-    if eopatch.bbox is None:  # remove after BBox is never None
-        features_to_save = []
-
-    for ftype, fname, feature_path in eopatch_features:
-        if ftype == FeatureType.BBOX:
-            continue
-        # the paths here do not have file extensions, but FeatureIO.save takes care of that
-        feature_io = _get_feature_io_constructor(ftype)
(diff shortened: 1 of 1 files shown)


core/eolearn/core/eodata_io.py
```diff
@@ -95,6 +120,17 @@ def save_eopatch(
     else:
         _check_letter_case_collisions(eopatch_features, [])
 
+
+def _remove_old_eopatch(filesystem: FS, patch_location: str) -> None:
+    filesystem.removetree(patch_location)
```
mlubej, line 125: what happens if someone does `EOPatch().save('/', overwrite_permission=OverwritePermission.OVERWRITE_PATCH)` 😬

core/eolearn/core/eodata_io.py
```diff
@@ -95,6 +120,17 @@ def save_eopatch(
...
+    if patch_location != "/":  # avoid redundant filesystem.makedirs if the location is '/'
+        filesystem.makedirs(patch_location, recreate=True)
```
mlubej, line 127: Can you explain this in more detail? I'm not sure what exactly is the reasoning here (I know there's no real change from before, just trying to follow the code).

core/eolearn/core/eodata_io.py
zigaLuksic: their problem :) any other idea on how to do patch overwriting? I guess we could try and identify all possible files (since we know the expected patch structure) and remove those... But that is a lot of extra work.

core/eolearn/core/eodata_io.py
zigaLuksic: if someone does `EOPatch().save("/", filesystem)` the root is never removed so it doesnt need to be recreated (or so Matej A claims).

I do feel like this is redundant (especially since `OVERWRITE_PATCH` is not a common choice), but didn't want to alter code in a refactor MR. But since you pointed it out, we might as well remove the conditional if we all agree :)

core/eolearn/core/eodata_io.py
mlubej: One way that comes to mind is to do a listdir and filter it with a regex pattern, which could somehow be auto-constructed from feature types. I agree it's some extra work, but doesn't seem impossible. If you think it's worth spending a couple of hours to increase the safety here, I can tackle this.

zigaLuksic: you can open up an issue about it, and we'll see if we ever get around to doing it

core/eolearn/core/eodata_io.py
mlubej: Ah I see yes in [the docs](/Users/mlubej/work/projects/sh-project/eo-learn/example_data/TestEOPatch')

So it's just optimizing to not call it, since the code wouldn't actually remove it, but the question is what do we really save anyway. I also think we can remove the conditional.

zigaLuksic: it's saving a single IO call in a very very niche case. I agree that it's not worth the code complication. Removed

mlubej (approved):

PR #589, opened 2023-02-27
A utility for generating test EOPatches with dummy data

mlubej:
Changes:
- adds a `PatchGenerator` class which has the functionality that was discussed internally
- adds a `PatchGeneratorConfig` dataclass which contains a more advanced setup of patch generation settings

Possible to:
- provide a custom bbox and/or timestamps
- provide a set of features to be included in the patch
- randomly generate random features
- generate dummy data for all features
- limit which feature types should be generated
- other things which are put under the `PatchGeneratorConfig`

### DIFF
--- a/core/eolearn/core/eodata.py
+++ b/core/eolearn/core/eodata.py
@@ -17,6 +17,7 @@ import copy
 import datetime as dt
 import logging
 from abc import ABCMeta, abstractmethod
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Set, Tuple, Type, TypeVar, Union, cast, overload
 from warnings import warn
 
@@ -24,7 +25,9 @@ import attr
 import dateutil.parser
 import geopandas as gpd
 import numpy as np
+import pandas as pd
 from fs.base import FS
+from numpy.random import Generator
 from typing_extensions import Literal
 
 from sentinelhub import CRS, BBox
@@ -36,7 +39,7 @@ from .exceptions import EODeprecationWarning
 from .types import EllipsisType, FeatureSpec, FeaturesSpecification
 from .utils.common import deep_eq, is_discrete_type
 from .utils.fs import get_filesystem
-from .utils.parsing import parse_features
+from .utils.parsing import FeatureParser, parse_features
 
 _T = TypeVar("_T")
 _Self = TypeVar("_Self")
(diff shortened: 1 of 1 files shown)


mlubej:
Even though it seems that this should be a relatively straight-forward task, it's brings a lot of decisions to be made for how to go about it. The current code seems to be in line with what was discussed internally. We can see later if some parts are over-engineered or still missing.

Once the utility seems to be in a mature state (if not already) we can start looking at how to use it in the tests.

core/eolearn/core/eodata.py
```diff
@@ -731,3 +734,92 @@ def plot(
...
+
+    def __post_init__(self):
+        self.bbox = BBox((0, 0, *self.raster_shape), crs=CRS.UTM_33N)
+        self.timestamps = pd.date_range(*self.timestamps_range, periods=self.timestamps_periods)
+        if not self.allowed_feature_types:
+            self.allowed_feature_types = [ftype for ftype in FeatureType if ftype.is_raster()]
+
+
+class PatchGenerator:
+    """A class for generating EOPatches with dummy data."""
```
mlubej, line 762: Might try to convert this back to a function now... theres a lot of definitions/docstrings ballast which might reduce the code footprint.

zigaLuksic: Agreed. One should (in general) use classes in 3 cases:
1. It has a mutable state that the methods interact with
2. Inheritance simplifies things (happens less often than OOP people believe)
3. (python specific) you want to organize things under a common namespace

Try switching to functions and I'll do a more thorough review then.

core/eolearn/core/eodata.py
```diff
@@ -731,3 +734,92 @@ def plot(
             config=config,
             **kwargs,
         )
+
+
+@dataclass
+class PatchGeneratorConfig:
```
zigaLuksic, line 740: I would move this to a new module `eolearn/core/utils/testing` or `eolearn/core/utils/test`.

mlubej: moved.

core/eolearn/core/eodata.py
mlubej: converted, much nicer now

core/eolearn/core/utils/testing.py
```diff
@@ -0,0 +1,83 @@
...
+
+@dataclass
+class PatchGeneratorConfig:
+    """Dataclass containing a more complex setup of the PatchGenerator class."""
+
+    num_timestamps: int = 5
+    timestamps_range: Tuple[dt.datetime, dt.datetime] = (dt.datetime(2019, 1, 1), dt.datetime(2019, 12, 31))
+    timestamps: List[dt.datetime] = field(init=False, repr=False)
+
+    num_random_features: int = 5
```
zigaLuksic, line 38: this is no longer relevant right?

zigaLuksic (approved):

core/eolearn/core/utils/testing.py
mlubej: right, removed. good catch!

mlubej: damn, I thought I pushed... Seems like I didn't. @zigaLuksic do I edit it straight on `develop`? Or you can remove it one of the incoming MRs...

zigaLuksic: develop is/should be protected.
you can do a tiny MR for that and merge it yourself :)

PR #591, opened 2023-02-28
Update extract and explode bands test

jgersak:
+ fixture change from test_eopatch -> patch
+ add types
+ split test that test failing

### DIFF
--- a/core/eolearn/tests/test_core_tasks.py
+++ b/core/eolearn/tests/test_core_tasks.py
@@ -387,15 +387,15 @@ def test_map_features_fails(patch: EOPatch) -> None:
 @pytest.mark.parametrize(
     "feature,  task_input",
     [
-        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "MOVED_BANDS"): [2, 4, 6]}),
-        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "MOVED_BANDS"): [2]}),
-        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "MOVED_BANDS"): (2,)}),
-        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "MOVED_BANDS"): 2}),
+        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "EXPLODED_BANDS"): [2, 4, 6]}),
+        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "EXPLODED_BANDS"): [2]}),
+        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "EXPLODED_BANDS"): (2,)}),
+        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "EXPLODED_BANDS"): 2}),
         (
             (FeatureType.DATA, "bands"),
             {(FeatureType.DATA, "B01"): [0], (FeatureType.DATA, "B02"): [1], (FeatureType.DATA, "B02 & B03"): [1, 2]},
         ),
-        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "MOVED_BANDS"): []}),
+        ((FeatureType.DATA, "bands"), {(FeatureType.DATA, "EXPLODED_BANDS"): []}),
     ],
 )
 def test_explode_bands(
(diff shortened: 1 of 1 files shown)


core/eolearn/tests/test_core_tasks.py
```diff
@@ -379,47 +379,48 @@ def test_map_features(test_eopatch):
...
 )
 def test_explode_bands(
-    test_eopatch: EOPatch,
-    feature: FeatureType,
+    patch: EOPatch,
+    feature: Tuple[FeatureType, str],
     task_input: Dict[Tuple[FeatureType, str], Union[int, Iterable[int]]],
-):
+) -> None:
     move_bands = ExplodeBandsTask(feature, task_input)
```
zigaLuksic, line 398: the 'move' naming is weird and doesn't fit this test (or the 'extract bands' test)

core/eolearn/tests/test_core_tasks.py
```diff
@@ -379,47 +379,48 @@ def test_map_features(test_eopatch):
...
+    bands = [2, 4, 6]
+    move_bands = ExtractBandsTask((FeatureType.DATA, "bands"), (FeatureType.DATA, "MOVED_BANDS"), bands)
+    patch = move_bands(patch)
+    assert np.array_equal(patch.data["MOVED_BANDS"], patch.data["bands"][..., bands])
 
     old_value = patch.data["MOVED_BANDS"][0, 0, 0, 0]
     patch.data["MOVED_BANDS"][0, 0, 0, 0] += 1.0
-    assert patch.data["REFERENCE_SCENES"][0, 0, 0, bands[0]] == old_value
+    assert patch.data["bands"][0, 0, 0, bands[0]] == old_value
     assert old_value + 1.0 == approx(patch.data["MOVED_BANDS"][0, 0, 0, 0])
```
zigaLuksic, line 417: this is a weird and unclear test. Figure out what it does and replace it with something better

core/eolearn/tests/test_core_tasks.py
```diff
@@ -379,47 +379,48 @@ def test_map_features(test_eopatch):
...
 
-    bands = [2, 4, 16]
-    move_bands = ExtractBandsTask((FeatureType.DATA, "REFERENCE_SCENES"), (FeatureType.DATA, "MOVED_BANDS"), bands)
+
+def test_extract_bands_fails(patch: EOPatch) -> None:
```
zigaLuksic, line 420: do we have a custom error for this? or why is this tested?

jgersak: define custom we have `ValueError` if not all bands are smaller then number of featuers but its not custom like `EOErrorLenght` 

from blame Matej added it some time ago, and I did not find good reason to remove or not it 
(aka i m not against removing it)

zigaLuksic: is the error raised by us, or is it just an error raised by `numpy` due to indexing?

jgersak: Error is raised by us.

zigaLuksic: ok, then the test stays

core/eolearn/tests/test_core_tasks.py
jgersak: Updated, name not needed

core/eolearn/tests/test_core_tasks.py
jgersak: It checks that `patch.data["bands"]` and `patch.data["new"]` are two different object replaced with id test

core/eolearn/tests/test_core_tasks.py
```diff
@@ -406,8 +406,9 @@ def test_explode_bands(
 def test_extract_bands(patch: EOPatch) -> None:
     bands = [2, 4, 6]
     patch = ExtractBandsTask((FeatureType.DATA, "bands"), (FeatureType.DATA, "MOVED_BANDS"), bands)(patch)
+
     assert_array_equal(patch.data["MOVED_BANDS"], patch.data["bands"][..., bands])
-    assert id(patch.data["MOVED_BANDS"]) != id(patch.data["bands"])
+    assert all([(id(patch.data["MOVED_BANDS"][band]) != id(patch.data["bands"][band]) for band in bands)])
```
zigaLuksic, line 411: I see you're very enthusiastic about `id` tests, but you're underestimating numpy.
Numpy can also return so called `views`. Look at the following snippet from the console:

```python
In [1]: import numpy as np

In [2]: x = np.ones((10, 10))

In [3]: y = x[3:5, 3:5]

In [4]: z = x[3:5, 3:5]

In [5]: id(y) == id(z)
Out[5]: False

In [6]: z[0, 0] = 0

In [7]: y
Out[7]: 
array([[0., 1.],
       [1., 1.]])
```

core/eolearn/tests/test_core_tasks.py
zigaLuksic: you still use the `move` naming, e.g. `"MOVED_BANDS"`

jgersak: Renamed

PR #592, opened 2023-02-28
Update test eop merge

jgersak:
Add test for merging eopatches

### DIFF
--- a/core/eolearn/tests/test_core_tasks.py
+++ b/core/eolearn/tests/test_core_tasks.py
@@ -481,31 +481,6 @@ def test_create_eopatch():
     assert np.array_equal(patch.data["bands"], data)
 
 
-def test_kwargs():
-    patch = EOPatch()
-    shape = (3, 5, 5, 2)
-
-    data1 = np.random.randint(0, 5, size=shape)
-    data2 = np.random.randint(0, 5, size=shape)
-
-    patch[(FeatureType.DATA, "D1")] = data1
-    patch[(FeatureType.DATA, "D2")] = data2
-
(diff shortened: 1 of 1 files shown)


zigaLuksic:
I wouldn't create too advanced examples here, since this is mostly tested in `test_eodata_merge.py`. A simple test or two that it is correctly linked to `EOPatch` should suffice

jgersak:
> I wouldn't create too advanced examples here, since this is mostly tested in test_eodata_merge.py. A simple test or two that it is correctly linked to EOPatch should suffice

Test case is significantly simplified, i wanted to write L529 less convoluted but kinda give up after a while ....

core/eolearn/tests/test_core_tasks.py
```diff
@@ -481,9 +481,34 @@ def test_create_eopatch():
...
+        bbox=DUMMY_BBOX,
+        timestamps=dummy_timestamps,
+    )
 
-    task.execute(test_eopatch, test_eopatch, test_eopatch)
+    merged_patch = MergeEOPatchesTask(time_dependent_op="max", timeless_op="concatenate")(patch1, patch2)
+
+    expected_patch = EOPatch(
+        data={"bands": np.ones((2, 4, 4, 3), dtype=np.float32)},
+        mask_timeless={"LULC": np.arange(0, 32).reshape((16, 2), order="F").reshape(4, 4, 2)},
```
zigaLuksic, line 504: it is a bit difficult to understand, but the only other option I see is `np.concatenate([np.arange(0,16).reshape(...), ...])` which also feels a bit weird. So let's just leave it.

zigaLuksic (approved):

core/eolearn/tests/test_core_tasks.py
jgersak: this solution would also demands that concatenation is done out side EOPatch init (i have tried it and like it even less)

PR #596, opened 2023-03-02
Update tests

jgersak:
Add missing types to tests
Update test_create_eopatch

### DIFF
--- a/core/eolearn/tests/test_core_tasks.py
+++ b/core/eolearn/tests/test_core_tasks.py
@@ -474,26 +474,15 @@ def test_extract_bands_fails(patch: EOPatch) -> None:
 
 
 @pytest.mark.parametrize(
-    "features, expected",
+    "features",
     [
-        ({}, 0),
-        ({"data": {"bands": np.arange(0, 32).reshape(1, 4, 4, 2)}}, 1),
-        (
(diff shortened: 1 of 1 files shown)


core/eolearn/tests/test_core_tasks.py
```diff
@@ -473,12 +473,27 @@ def test_extract_bands_fails(patch: EOPatch) -> None:
...
+def test_create_eopatch(features: Dict[str, Any], expected: int) -> None:
+    created_patch = CreateEOPatchTask()(**features)
+    assert len(created_patch.get_features()) == expected
+
+    for feature in created_patch.get_features():
+        type, name = feature
+        if isinstance(created_patch[feature], np.ndarray):
+            assert_array_equal(created_patch[feature], features[type.value][name])
+        else:
+            assert features[type.value] == created_patch[feature]
```
zigaLuksic, line 496: Try to avoid using `type` as a variable name.

Also, here i believe you could do just
```python
assert CreateEOPatchTask()(**features) == EOPatch(**features)
```

I think it's a fair assumption that the interface to `CreateEOPatchTask` and the `EOPatch` constructor are the same.

jgersak: Agree test was testing more `EOPatch()` then `CreateEOPatchTask()`, updated.

zigaLuksic (approved):

PR #598, opened 2023-03-03
Prevent bbox deletion

zigaLuksic:
Also deprecate `reset_feature_type` because `del` does it better.

### DIFF
--- a/core/eolearn/core/eodata.py
+++ b/core/eolearn/core/eodata.py
@@ -28,6 +28,7 @@ from fs.base import FS
 from typing_extensions import Literal
 
 from sentinelhub import CRS, BBox
+from sentinelhub.exceptions import deprecated_function
 
 from .constants import TIMESTAMP_COLUMN, FeatureType, OverwritePermission
 from .eodata_io import FeatureIO, load_eopatch, save_eopatch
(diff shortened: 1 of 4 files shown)


core/eolearn/core/eodata.py
```diff
@@ -533,7 +544,7 @@ def reset_feature_type(self, feature_type: FeatureType) -> None:
         if feature_type.has_dict():
             self[feature_type] = {}
         elif feature_type is FeatureType.BBOX:
-            self[feature_type] = None
+            raise ValueError("The BBox of an EOPatch should never be undefined.")
```
jgersak, line 547: I do not know how to improve error massage just se a smal issue wit our current typing for `bbox: Optional[BBox]`

core/eolearn/tests/test_eodata.py
```diff
@@ -178,6 +178,23 @@ def test_delete_existing_feature(feature: FeatureSpec, mini_eopatch: EOPatch) ->
...
+@pytest.mark.parametrize("feature_type", [FeatureType.DATA, FeatureType.TIMESTAMPS])
+def test_delete_existing_feature_type(feature_type: FeatureType, mini_eopatch: EOPatch) -> None:
+    old = mini_eopatch.copy(deep=True)
+
+    del mini_eopatch[feature_type]
+    assert feature_type not in mini_eopatch
+
+    for ftype, fname in old.get_features():
+        if ftype != feature_type:
+            assert_array_equal(old[ftype, fname], mini_eopatch[ftype, fname])
```
jgersak, line 190: i think we should not use `assert_array_equal` on BBOX

core/eolearn/tests/test_eodata.py
```diff
@@ -178,6 +178,23 @@ def test_delete_existing_feature(feature: FeatureSpec, mini_eopatch: EOPatch) ->
...
+
+
+def test_cannot_delete_bbox(mini_eopatch: EOPatch) -> None:
+    with pytest.raises(ValueError):
+        del mini_eopatch[FeatureType.BBOX]
```
jgersak, line 195: It works correctly for FeatureSpec (FeatureType.BBOX, None) a test case could be added

core/eolearn/core/eodata.py
zigaLuksic: It is a bit 'in the future'. I can change it to a warning if you think it would be better.

jgersak: if it is going to be a bit i guess it can stay, just want to point out possible "miss erroring" ...

core/eolearn/tests/test_eodata.py
zigaLuksic: it cant be :) because BBOX cannot be deleted

core/eolearn/core/eodata.py
zigaLuksic: i will keep it as it is, because we already force BBOX to be present whenever possible. Let the users see a glimpse of the future. In the next release there will be a deprecation warning about BBox being necessary anyway.

core/eolearn/tests/test_eodata.py
zigaLuksic: parametrized

jgersak (approved):

PR #599, opened 2023-03-03
Test patch generator

jgersak:


core/eolearn/core/utils/testing.py
```diff
@@ -77,6 +77,9 @@ def _generate_feature_data(
...
     rng: np.random.Generator, ftype: FeatureType, timestamps: List[dt.datetime], config: PatchGeneratorConfig
-) -> Tuple[int, ...]:
-    time, height, width, depth = len(timestamps), *config.raster_shape, rng.integers(*config.depth_range)
-    return (time, height, width, depth) if ftype.is_temporal() else (height, width, depth)
+) -> List[int]:
+    shape = [len(timestamps)] if ftype.is_temporal() else []  # time
+    f_dim: int = ftype.ndim() or 0
+    shape.extend(config.raster_shape if f_dim >= 3 else [])  # height, width
+    shape.append(rng.integers(*config.depth_range))  # depth
+    return shape
```
zigaLuksic, line 85: I cannot understand this code at all
Describe the problem & solution to me in human words.

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,60 @@
...
+
+from eolearn.core.constants import FeatureType
+from eolearn.core.types import FeatureSpec, FeaturesSpecification
+from eolearn.core.utils.testing import patch_generator
+
+
+def test_patch_generator_default() -> None:
+    patch = patch_generator()
+    assert hasattr(patch, "bbox")
+    assert hasattr(patch, "timestamps")
```
zigaLuksic, line 15: this can never be false, an EOPatch ALWAYS has those attributes by construction
what is the purpose of this test?

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_patch_generator_not_default() -> None:
```
zigaLuksic, line 18:
```suggestion
def test_patch_generator_set_bbox_timestamps() -> None:
```

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,60 @@
...
+        (FeatureType.LABEL, "label"),
+        (FeatureType.DATA_TIMELESS, "data_timeless"),
+        (FeatureType.MASK_TIMELESS, "mask_timeless"),
+        (FeatureType.SCALAR_TIMELESS, "scalar_timeless"),
+        (FeatureType.LABEL_TIMELESS, "label_timeless"),
+    ],
+)
+def test_patch_generator_dim(feature: FeatureSpec) -> None:
+    patch = patch_generator(feature)
+    assert patch[feature].ndim == feature[0].ndim()
```
zigaLuksic, line 42: this should be taken care of by the `EOPatch` (or rather by the `FeatureTypeDict` objects)
Did you find an example where the `ndim` of an feature value was wrong?

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_patch_generator_fails() -> None:
+    with pytest.raises(ValueError):
+        # fails because it is not `raster` and only raster features are supported
+        patch_generator((FeatureType.META_INFO, "meta_info"))
```
zigaLuksic, line 48: parametrize and include `FeatureType.VECTOR` and `FeatureType.VECTOR_TIMELESS`

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.mark.parametrize("seed", [0, 1, 42, 100])
+@pytest.mark.parametrize(
+    "featurs",
+    [{}, {FeatureType.DATA: ["bands, CLP"]}, {FeatureType.DATA: ["bands, CLP"], FeatureType.MASK_TIMELESS: "LULC"}],
+)
+def test_patch_generator_seed(seed: int, featurs: FeaturesSpecification) -> None:
+    data_feature = (FeatureType.DATA, "neki")
```
zigaLuksic, line 57: why is this used instead of the (misspelled) `featurs`?

zigaLuksic (commented):
There is no test that would check that all the requested features were indeed generated

core/eolearn/core/utils/testing.py
mlubej: The issue here is that the old code assumes that the feature shape will always be of image type, like discussed internally in https://git.sinergise.com/eo/code/eo-learn/-/issues/194

We had this piece of code before we (@zigaLuksic and me) together changed it to what is seen on the left.

Basically this constructs the feature shape based on the feature type. Not sure why L82 is there though. @jgersak if you were afraid that this would fail in case of meta info or vector types, this shouldn't be the case, because only raster type features are allowed, see L55

jgersak:
```py
    time, height, width, depth = len(timestamps), *config.raster_shape, rng.integers(*config.depth_range)
    shape_dict = {4: (time, height, width, depth), 3: (height, width, depth), 2: (time, depth), 1: (depth,)}
    return shape_dict[ftype.ndim()]
```
this was first solution and matic thought that it is too much of "black magic" so i do adaptation of his:

current solution should first look if ftype is_temporal and int len timestamps  or empty list dependent on that 
second test is if ftype.ndim() > 2 all our features that are that dim have spatial component so raster_shape can be append 
all features have free component on last place (L84)

jgersak: L82  is to make mypy happy xD

mlubej: > this was first solution and matic thought that it is too much of "black magic" so i do adaptation of his:


It looks more elegant, but I just wasn't sure if it's safe to assume that a single dimension number always corresponds to a unique shape representation. currently this seems to be the case, but ... 

perhaps I'm wrong and this is better, I thought that the changes here were obvious enough

zigaLuksic: How about we focus on code that is human readable, like the original was?
The first thing that i noticed is that yall were saying "not eveything is an image"
good, try something like this then.

```python
    time, depth = len(timestamps), rng.integers(*config.depth_range)
    if <is_image_ftype>:
        return (time, height, width, depth) if ftype.is_temporal() else (height, width, depth)
    return <whatever belongs here>
```

If things are not images we also have no clue what the dimension is supposed to be anyway (or at leas the current config doesn't tell us) so such a separation is much more sensible.

core/eolearn/tests/test_utils/test_testing.py
jgersak: yes if you throw FeatureType.SCALAR into patch_generator before u get 4 dim back this was the whole point of this "branch/issue/tests" ... but ATM agrre with your last point that for 2 dim data we do not know the size that we want

imo this test shuld stay mybe not in this form but for feturs that we know how to generate data one test and one that throw ValueError for all that we do not know how to generate ...

zigaLuksic: really? please provide me with the code, because we clearly have a bug.

The 'regular' version fails as it should. So it's very weird that it goes through.
```python
In [1]: from eolearn.core import FeatureType, EOPatch

In [2]: import numpy as np

In [3]: eop = EOPatch()

In [4]: eop[FeatureType.SCALAR, "something"] = np.zeros((10, 10, 10, 10))
# THIS FAILS, AS IT SHOULD
```

mlubej: > yes if you throw FeatureType.SCALAR into patch_generator before u get 4 dim back

This was because we assumed in the code that "raster" means "image-type", which we then discovered it's not true, so we got height and width also for SCALAR* and LABEL*.  If the feature shape constructor works as is supposed to (now that we fixed it) this should be a problem anymore and there shouldn't be any bugs related to this. The only issue is our understanding of "is_raster" or perhaps it's definition.

zigaLuksic: my point here is that the `assert` can never be triggered. L42 is unreachable, because if the data has the wrong `ndim` then L41 will throw an error.

core/eolearn/core/utils/testing.py
jgersak: reverted,  update supported_feature_types in patch_generator

core/eolearn/tests/test_utils/test_testing.py
jgersak: removed

core/eolearn/tests/test_utils/test_testing.py
jgersak: removed, add cases for fails

core/eolearn/tests/test_utils/test_testing.py
jgersak: done

core/eolearn/tests/test_utils/test_testing.py
jgersak: updated

zigaLuksic:
i know this is outside the initial scope of the MR, but could we change `patch_generator` to `generate_eopatch`?

core/eolearn/core/utils/testing.py
```diff
@@ -79,4 +79,7 @@ def _get_feature_shape(
     rng: np.random.Generator, ftype: FeatureType, timestamps: List[dt.datetime], config: PatchGeneratorConfig
 ) -> Tuple[int, ...]:
     time, height, width, depth = len(timestamps), *config.raster_shape, rng.integers(*config.depth_range)
-    return (time, height, width, depth) if ftype.is_temporal() else (height, width, depth)
+    f_dim: int = ftype.ndim() or 0
+    if f_dim >= 3:
```
mlubej, line 83: so, this is the ugly part that we talked about and perhaps it would really make sense to change it as we discussed to:

```suggestion

    if ftype.is_spatial() and not ftype.is_vector():
```

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,43 @@
...
+    "feature",
+    [
+        (FeatureType.META_INFO, "meta_info"),
+        (FeatureType.VECTOR, "vector"),
+        (FeatureType.VECTOR_TIMELESS, "vector_timeless"),
+    ],
+)
+def test_generate_eopatch_fails(feature: FeatureSpec) -> None:
+    with pytest.raises(ValueError):
+        # fails because it is not `raster` and only raster features are supported
```
zigaLuksic, line 31: remove

zigaLuksic (commented):
there is still no test that would show that the features ever get generated, let alone that they match the specifications of the config.

Also if custom timestamps are given, they must match the length of them.

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@pytest.mark.parametrize("seed", [0, 1, 42, 100])
+@pytest.mark.parametrize(
+    "features",
+    [{}, {FeatureType.DATA: ["bands, CLP"]}, {FeatureType.DATA: ["bands, CLP"], FeatureType.MASK_TIMELESS: "LULC"}],
+)
+def test_generate_eopatch_seed(seed: int, features: FeaturesSpecification) -> None:
+    patch1 = generate_eopatch(features)
+    patch2 = generate_eopatch(features)
```
mlubej, line 42: `seed` is not used

mlubej:
> Also if custom timestamps are given, they must match the length of them.

This is probably already covered by `test_generate_eopatch_set_bbox_timestamps` where the timestamps arrays are directly compared

core/eolearn/tests/test_utils/test_testing.py
jgersak: done

core/eolearn/tests/test_utils/test_testing.py
jgersak: done

jgersak:
> there is still no test that would show that the features ever get generated, let alone that they match the specifications of the config. Also if custom timestamps are given, they must match the length of them.

 dummy test added if you want i can create TEST_CASE testing but its going to take some time :)

zigaLuksic:
> > Also if custom timestamps are given, they must match the length of them.
> 
> This is probably already covered by `test_generate_eopatch_set_bbox_timestamps` where the timestamps arrays are directly compared

the timestamps are compared, but i need comparison of timestamps and the temporal dimension of feature data

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,58 @@
...
+
+def test_generate_eopatch_set_bbox_timestamps() -> None:
+    bbox = BBox((0, 0, 10, 10), crs=CRS("EPSG:32633"))
+    timestamps = [dt.datetime(2019, 1, 1)]
+    patch = generate_eopatch((FeatureType.DATA, "bands"), bbox=bbox, timestamps=timestamps)
+
+    assert patch.bbox == bbox
+    assert patch.timestamps == timestamps
+
+    assert patch[(FeatureType.DATA, "bands")].shape[0] == len(timestamps)
```
mlubej, line 22: might check the spatial dimensions as well, since the info is all there. 

Added slight improvements:

```suggestion
def test_generate_eopatch_set_bbox_timestamps() -> None:
    feature = (FeatureType.DATA, "bands")
    bbox = BBox((0, 0, 10, 10), crs=CRS("EPSG:32633"))
    timestamps = [dt.datetime(2019, 1, 1)]
    
    default_config = PatchGeneratorConfig()
    patch = generate_eopatch((FeatureType.DATA, "bands"), bbox=bbox, timestamps=timestamps, config=default_config)

    assert patch.bbox == bbox
    assert patch.timestamps == timestamps

    time, *raster_shape, _ = patch[feature].shape
    assert time == len(timestamps)
    assert raster_shape == default_config.raster_shape
```

zigaLuksic: I suggest the above suggestion is done in a separate test. Let's not test too many things at once.

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,58 @@
...
+
+@pytest.mark.parametrize("seed", [0, 1, 42, 100])
+@pytest.mark.parametrize(
+    "features",
+    [{}, {FeatureType.DATA: ["bands, CLP"]}, {FeatureType.DATA: ["bands, CLP"], FeatureType.MASK_TIMELESS: "LULC"}],
+)
+def test_generate_eopatch_seed(seed: int, features: FeaturesSpecification) -> None:
+    patch1 = generate_eopatch(features, seed=seed)
+    patch2 = generate_eopatch(features, seed=seed)
+    assert patch1 == patch2
```
zigaLuksic, line 46: why is there still no test that all the features are present?? :joy:

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def test_generate_eopatch_config() -> None:
+    test_config = {
+        "num_timestamps": 3,
+        "timestamps_range": (dt.datetime(2022, 1, 1), dt.datetime(2022, 12, 31)),
+        "max_integer_value": 1,
+        "raster_shape": (1, 1),
+        "depth_range": (1, 2),
```
zigaLuksic, line 55: make these two more interesting, so that we don't fall into an edge-case where everything defaults to 1 or something

jgersak: > make these two more interesting
updated

core/eolearn/tests/test_utils/test_testing.py
jgersak: done

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,140 @@
...
+        GenerateTestCase(
+            seed=3,
+            data_features=(FeatureType.MASK, "CLM"),
+            expected_statistics={
+                (FeatureType.MASK, "CLM"): {
+                    "exp_shape": (5, 98, 151, 2),
+                    "exp_min": 0,
+                    "exp_max": 255,
+                    "exp_mean": 127.38907960535208,
+                    "rel_delta": 0.0001,
```
zigaLuksic, line 85: since `rel_delta` doesn't really get changed, maybe hardcode it in the test (less clutter in test cases)

I would also keep only a few values after the decimal point, also so that there is less text to look at. Otherwise it's fine

core/eolearn/tests/test_utils/test_testing.py
```diff
@@ -0,0 +1,140 @@
...
+        {"num_timestamps": 3, "max_integer_value": 1, "raster_shape": (1, 1), "depth_range": (1, 2)},
+        {"num_timestamps": 7, "max_integer_value": 333, "raster_shape": (3, 27), "depth_range": (5, 15)},
+    ],
+)
+def test_generate_eopatch_config(config: Dict[str, Any]) -> None:
+    mask_feature = (FeatureType.MASK, "mask")
+
+    patch = generate_eopatch(mask_feature, config=PatchGeneratorConfig(**config))
+
+    time, raster_x, raster_y, depth = patch[mask_feature].shape
```
zigaLuksic, line 136: AFAIK it's `time, height, width, depth`, so you got x and y mixed up (doesn't affect the test, just the names are switched)

jgersak: switch the names

core/eolearn/tests/test_utils/test_testing.py
jgersak: done

PR #603, opened 2023-03-03
Add T-Digest task

meengel:
As part of the Global Earth Monitor Project, a solution for data normalisation of datasets larger than RAM had to be found. The [T-Digest algorithm by Ted Dunning](https://arxiv.org/abs/1902.04023) has been identified as a suitable tool to do so. The following EOTask integrates that algorithm and computes the T-Digest representation of a chosen feature of an EOPatch. The resulting T-Digest feature may be used to efficiently compute quantiles of the underlying dataset.

Please note that a reducing method to merge the partial T-Digest representations is necessary to use a representation of the whole dataset, i.e. all EOPatches of an area of interest. That merging method has already been implemented and will be included at some point - ongoing discussion. In conjunction with the EOTask presented here, that enables quantile based normalisation or statistical analysis of datasets larger than RAM in EO.


batic:
Hi @meengel 

Thanks for the PR. There are a few things that I'd ask you to do before we review the t-digest:

 * bump your branch to latest develop branch (to ease the PR, to get the latest versions of pre-commit hooks, etc.)
 * run the `pip install -r requirements-dev.txt` and then perform the following tests to see that everything is as expected:
 * `pre-commit run --all-files` (this will run all the linters and code checkers - black, isort, flake8); see what they report and fix 
 * also run `mypy ml_tools` 

PS: You can see the errors from the CI following the "Details" links in the checks box above.

meengel:
@batic @zigaLuksic Please have a look - I made everything you wanted but flake still throws an complexity error.
However, reducing the complexity of the task is not meaningful in my opinion.

zigaLuksic:
> @batic @zigaLuksic Please have a look - I made everything you wanted but flake still throws an complexity error. However, reducing the complexity of the task is not meaningful in my opinion.

Hi @meengel, I can guarantee you that the complexity of the code can be reduced and it will be worthwhile. However changing research code to production code is not a trivial effort, so I'll guide you through it (and don't be discouraged by a lot of comments, that is the nature of the process).

As a rough outline of what we're going to do:
1. Removing redundant parts of code
2. Splitting larger methods into smaller ones
3. Adding types (eo-learn will soon prevent any code from not being annotated)
4. Adding tests (incredibly important for maintainers)

I'll go ahead and make a MR that focuses on step 1, and we'll proceed from there.

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,199 @@
...
+                    " features of the same length!"
+                )
+        else:
+            self.in_feature = [in_feature]
+            if isinstance(out_feature, list):
+                raise RuntimeError(
+                    "TDigestTask: you cannot get a list of T-Digest representations from a single feature!"
+                )
+            else:
+                self.out_feature = [out_feature]
```
zigaLuksic, line 74: If i understand correctly, here you have either 1 feature mapping to 1 feature, or N features mapping to N features?
In any case, i suggest you use our *feature parsers* which also support a large range of inputs. The easiest way for you is to take a look at [the init of the MapFeatureTask](https://github.com/sentinel-hub/eo-learn/blob/master/core/eolearn/core/core_tasks.py#L404) which has a similar approach when it comes to features.

This way you also don't need to think about 'do I have a single feature, do i have many'. If a single feature is passed as input, it gets transformed into a list with a single element, and you should be able to work with it as if it were a regular list of features.

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,199 @@
...
+                            eopatch[out_feature][i, j, k].batch_update(eopatch[in_feature][:, i, j, k].flatten())
+                    elif len(shape) == 3:  # add feature is timeless check
+                        for i, j, k in product(range(shape[0]), range(shape[1]), range(shape[2])):
+                            eopatch[out_feature][i, j, k] = td.TDigest()
+                            eopatch[out_feature][i, j, k].batch_update(eopatch[in_feature][i, j, k].flatten())
+                    else:
+                        raise RuntimeError(
+                            f"TDigestTask: {self.mode} mode only defined for DATA, DATA_TIMELESS, MASK or MASK_TIMELESS"
+                            " feature types!"
+                        )
```
zigaLuksic, line 114: to avoid the need for such runtime errors, i suggest you specify the `allowed_feature_types` in the feature parser in the `__init__`. See [here](https://github.com/sentinel-hub/eo-learn/blob/master/core/eolearn/core/utils/parsing.py#L306) for the type of the field.

In your case it should be something like `parse_features(input_features, allowed_feature_types=[FeatureType.DATA, FeatureType.DATA_TIMELESS, ...])`

If different modes require different feature types, that is something that you can also do in the init (so determine the value of a `allowed_types` variable based on `mode` and then use it in the parser).

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,199 @@
...
+                if pixelwise:
+                    raise NotImplementedError(f"TDigestTask: pixelwise for {self.mode} mode not implemented yet!")
+
+                else:
+                    eopatch[out_feature] = td.TDigest()
+                    eopatch[out_feature].batch_update(eopatch[in_feature].flatten())
+
+        # errorneous modes
+        else:
+            raise RuntimeError(f"TDigestTask: mode {self.mode} not implemented!")
```
zigaLuksic, line 196: anything that is not implemented should be removed from the code. Since some `mode` types are not supported, they should also be removed from the docs. It is simple to add them in a future MR if needed, no need for such placeholders.

For the `total + pixelwise` issue, the error should be raised in the `init` and should probably be a `ValueError`. Also mention in the docs that the combination is not possible.

ml_tools/eolearn/ml_tools/TDigestTask.py
zigaLuksic: the users still get an error message, it is just not as specific. I would suggest that suitability of inputs is described in the docs e.g. (i'm making stuff up) `- The 'total' mode computes the total T-Digest representation of the whole feature accumulating all timestamps, bands and pixels. Cannot be computed on timeless data.`

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,178 @@
...
+        elif pixelwise and self.mode == "total":
+            raise ValueError("Total mode does not support pixelwise=True.")
+        else:
+            self.pixelwise = pixelwise
+
+        # set mode
+        if not mode:
+            self.mode = "standard"
+        else:
+            self.mode = mode
```
zigaLuksic, line 65: This can likewise be solved by having `mode: str = "standard"` and `self.mode=mode`.

If you want to help out the users even more, you can import `Literal` from `typing_extensions` and have `mode: Literal["standard", "timewise", "monthly", "total"] = "standard"`. This way a type-checker can warn them about typos

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,178 @@
...
+
+        elif mode == "timewise" or mode == "monthly":
+            allowed_in_types = [FeatureType.DATA, FeatureType.MASK]
+            if self.pixelwise:
+                allowed_out_types = [FeatureType.DATA]
+            else:
+                allowed_out_types = [FeatureType.SCALAR]
+
+        elif mode == "total":
+            allowed_in_types = None
```
zigaLuksic, line 88: does this mean that no input types are allowed? or that all are?
If all are allowed use the `...` value

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,178 @@
...
+            allowed_in_types = [
+                FeatureType.DATA,
+                FeatureType.DATA_TIMELESS,
+                FeatureType.MASK,
+                FeatureType.MASK_TIMELESS,
+            ]
+            if self.pixelwise:
+                allowed_out_types = [FeatureType.DATA_TIMELESS]
+            else:
+                allowed_out_types = [FeatureType.SCALAR_TIMELESS]
```
zigaLuksic, line 78:
```suggestion
            allowed_out_types = [FeatureType.DATA_TIMELESS] if pixelwise else [FeatureType.SCALAR_TIMELESS]
```

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,178 @@
...
+
+        elif mode == "timewise" or mode == "monthly":
+            allowed_in_types = [FeatureType.DATA, FeatureType.MASK]
+            if self.pixelwise:
+                allowed_out_types = [FeatureType.DATA]
+            else:
+                allowed_out_types = [FeatureType.SCALAR]
```
zigaLuksic, line 85:
```suggestion
            allowed_out_types = [FeatureType.DATA] if pixelwise else [FeatureType.SCALAR]
```

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # total mode
+        elif self.mode == "total":
+            for in_feature, out_feature in zip(self.in_feature, self.out_feature):
+                eopatch[out_feature] = td.TDigest()
+                eopatch[out_feature].batch_update(eopatch[in_feature].flatten())
+
+        # errorneous modes
+        else:
+            raise RuntimeError(f"TDigestTask: mode {self.mode} not implemented!")
```
zigaLuksic, line 175: this can no longer happen, since you prevent this case in the `__init__`

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # return
```
zigaLuksic, line 177: you can now remove all comments, since the code is obvious enough about it

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,178 @@
...
+                allowed_out_types = [FeatureType.DATA]
+            else:
+                allowed_out_types = [FeatureType.SCALAR]
+
+        elif mode == "total":
+            allowed_in_types = None
+            allowed_out_types = [FeatureType.SCALAR_TIMELESS]
+
+        else:
+            raise ValueError(f"The mode {mode} is not allowed.")
```
zigaLuksic, line 92: this whole "figuring out the allowed types" is quite a few lines, which break the readers concentration.

It would be best if you extract them in a helper method (a static method even), something like `_get_allowed_ftypes(mode: str, pixelwise: bool) -> Tuple[List[FeatureType], List[FeatureType]] `.

The types depend on how the `None` you used is interpreted, but we'll get to that after it has been extracted

zigaLuksic: Another great news for this part specifically. We just enabled a function option for `allowed_feature_types` on `develop`.
If you update your branch you'll be able to do `allowed_feature_types=my_func` which should accept a feature type and return a bool (whether or not that feature type is allowed).

You can then extract all this logic into a function and use it in the feature parser. Since you'll need to pass in `mode` you can use `partial` to make your life easier.

ml_tools/eolearn/ml_tools/TDigestTask.py
meengel: So you mean to support python 3.7? because from 3.8 ongoing, you could import it from typing itself, right?

zigaLuksic: we'll probably be removing support for 3.7 in the next major release, until then we need the whole codebase to be 3.7 compliant. But you are indeed correct, in 3.8 and higher you can import it directly from `typing`.

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,168 @@
...
+        allowed_out_types = [FeatureType.DATA] if pixelwise else [FeatureType.SCALAR]
+
+    elif mode == "total":
+        allowed_in_types = ...
+        allowed_out_types = [FeatureType.SCALAR_TIMELESS]
+
+    if ifin:
+        return True if allowed_in_types is Ellipsis else featuretype in allowed_in_types
+    else:
+        return True if allowed_out_types is Ellipsis else featuretype in allowed_out_types
```
zigaLuksic, line 168:
```suggestion
ModeTypes = Literal["standard", "timewise", "monthly", "total"]


def _is_input_ftype(feature_type: FeatureType, mode: ModeTypes) -> bool:
    if mode == "standard":
        return feature_type.is_image()
    if mode in ("timewise", "monthly"):
        return feature_type in [FeatureType.DATA, FeatureType.MASK]
    return True


def _is_output_ftype(feature_type: FeatureType, mode: ModeTypes, pixelwise: bool) -> bool:
    if mode == "standard":
        return feature_type == (FeatureType.DATA_TIMELESS if pixelwise else FeatureType.SCALAR_TIMELESS)

    if mode in ("timewise", "monthly"):
        return feature_type == (FeatureType.DATA if pixelwise else FeatureType.SCALAR)

    return feature_type == FeatureType.SCALAR_TIMELESS
```

This can be made simpler by separating the input and output type function. I also introduced a custom type, which you can also use in the `__init__` (the docs will inline the actual `Literal` type, so this is just for our convenience)

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,168 @@
...
+                if self.pixelwise:
+                    eopatch[out_feature] = np.empty(shape, dtype=object)
+                    for time_, i, j, k in product(range(shape[0]), range(shape[1]), range(shape[2]), range(shape[3])):
+                        eopatch[out_feature][time_, i, j, k] = td.TDigest()
+                        eopatch[out_feature][time_, i, j, k].batch_update(eopatch[in_feature][time_, i, j, k].flatten())
+
+                else:
+                    eopatch[out_feature] = np.empty(shape[[0, -1]], dtype=object)
+                    for k in range(shape[-1]):
+                        eopatch[out_feature][time_, k] = td.TDigest()
```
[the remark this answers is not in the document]
meengel: sry, somehow erased it from the for-loop

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,168 @@
...
+            out_feature, allowed_feature_types=partial(_get_allowed_ftypes, ifin=False, mode=mode, pixelwise=pixelwise)
+        )
+
+        if len(self.in_feature) != len(self.out_feature):
+            raise ValueError(
+                f"The number of input ({len(self.in_feature)}) and output features ({len(self.out_feature)}) must"
+                " match."
+            )
+
+    def execute(self, eopatch: EOPatch) -> EOPatch:
```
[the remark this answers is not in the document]
meengel: tried to solve the second by a generator - did not find a very meaningful solution for the pixelwise.

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,188 @@
...
+                in_feature=self.in_feature, out_feature=self.out_feature, eopatch=eopatch, pixelwise=self.pixelwise
+            )
+
+        elif self.mode == "monthly":
+            _process_monthly(
+                in_feature=self.in_feature, out_feature=self.out_feature, eopatch=eopatch, pixelwise=self.pixelwise
+            )
+
+        elif self.mode == "total":
+            _process_total(in_feature=self.in_feature, out_feature=self.out_feature, eopatch=eopatch)
```
zigaLuksic, line 95: since all of these methods use the `_looper` (except for the final one, but we can just ignore the shape there), let's extract the `for loop` outside, so that we'll have something like:
```python
        for in_feature_, out_feature_, shape in _looper(in_feature=in_feature, out_feature=out_feature, eopatch=eopatch):
            if self.mode == "standard":
                _process_standard(...)
            elif self.mode == "timewise":
                ...
```

Then you might also notice that the `if`s are kinda repetitive. And you can also make these functions return the `np.ndarray` instead of assigning the value in the function. So maybe we can do something like:
```python
processing_function =  {
    "timewise": _process_timewise,
    ...
}[self.mode]

for in_feature_, out_feature_, shape in _looper(in_feature=in_feature, out_feature=out_feature, eopatch=eopatch):
    eopatch[out_feature] = processing_function(in_feature, eopatch)
```

You'll need a special case for `monthly` but that can probably be solved with the use of `kwargs`.

ml_tools/eolearn/ml_tools/TDigestTask.py
```diff
@@ -0,0 +1,188 @@
...
+
+def _process_standard(
+    in_feature: FeaturesSpecification, out_feature: FeaturesSpecification, eopatch: EOPatch, pixelwise: bool
+) -> None:
+    for in_feature_, out_feature_, shape in _looper(in_feature=in_feature, out_feature=out_feature, eopatch=eopatch):
+        if pixelwise:
+            eopatch[out_feature_] = np.empty(shape[-3:], dtype=object)
+            for i, j, k in product(range(shape[-3]), range(shape[-2]), range(shape[-1])):
+                eopatch[out_feature_][i, j, k] = td.TDigest()
+                eopatch[out_feature_][i, j, k].batch_update(eopatch[in_feature_][..., i, j, k].flatten())
```
zigaLuksic, line 136: this comes up often. Perhaps consider something along the lines of
```python
def get_tdigest(values: np.ndarray) -> td.TDigest:
    result = td.TDigest()
    result.batch_update(values.flatten())
    return result
```
This way you have just:
```python
out_array[i, j, k] = get_tdigest(in_array[..., i, j k])
```

ml_tools/eolearn/ml_tools/TDigestTask.py
meengel: Please have a look at the [commit](https://github.com/sentinel-hub/eo-learn/pull/603/commits/21251363dc8c3f7254d788f99cd9d9c4c0981148) if it's okay like this.

zigaLuksic: looks like it's going in the right direction for sure :D
I'll run our CI, it `pylint` will likely complain about certain names.

Besides that I also noticed that you always do `.flatten()` before passing arrays into `get_tdigest`, but you already flatten inside, so you don't need to do that.

While i'm sure we could reduce the code a bit further, I feel like it's in a good enough place for now.
After satistfying the CI and fixing the `flatten` thing, I would focus on setting up test cases.

I'd start with a small easy test case or two, just so we know the task runs and does its thing.

meengel: Sry, completely missed the flatten-thing. Done now. Please confirm that you are taking care of the pylint and the first tests.

zigaLuksic: ah, since you changed the name of the file, you also need to adjust the `__init__` file.

meengel: ah, ups, dumb mistake. Done.

meengel: Regarding the tests: what should they show? Which cases should be captured? How much of effort do you expect that to be?

zigaLuksic: at the very least it would be nice to have one or two cases of things going as they should (just to check that the end-to-end process works). We sometimes just run it on some inputs and then assert that the statistics match the expected ones (see [here](https://github.com/sentinel-hub/eo-learn/blob/master/features/eolearn/tests/test_haralick.py) for instance).

You can also verify just that the object you receive is populated with correct types and that the shapes are ok (like [here](https://github.com/sentinel-hub/eo-learn/blob/master/features/eolearn/tests/test_feature_manipulation.py#L261)).

From my viewpoint we can also add the tests in a later MR if a deadline has to be met.

meengel: I let a student quickly test the current version and it seems to work in the context of the eo-learn examples, we provide.
Hence, I would suggest, we continue without tests at a first stage as we have to finish the IGARSS project together which will be bit of work as well.
Later, these tests should definitely be added.

It would be great if you could take care of that and inform me what's been wrong so I can improve next time!

PR #604, opened 2023-03-06
Fix issues for EOPatch loading when "load all" specifications were given

zigaLuksic:


core/eolearn/tests/test_eodata_io.py
```diff
@@ -121,26 +123,47 @@ def test_overwriting_non_empty_folder(eopatch, fs_loader):
...
+        assert eopatch == loaded_eopatch
+
+
+@mock_s3
+@pytest.mark.parametrize("fs_loader", FS_LOADERS)
+@pytest.mark.parametrize(
+    "save_features, load_features",
+    [
+        (..., ...),
+        (..., ...),
```
mlubej, line 149: Is this accidentally duplicated? With the `with` statement below I would imagine the state is reset for each entry in the parametrization list.

If the state is not reset, I would expect this to fail, since the default overwrite_permission is ADD_ONLY.

zigaLuksic: it's an oopsie :see_no_evil:

core/eolearn/tests/test_eodata_io.py
```diff
@@ -121,26 +123,46 @@ def test_overwriting_non_empty_folder(eopatch, fs_loader):
 
 @mock_s3
 @pytest.mark.parametrize("fs_loader", FS_LOADERS)
-def test_save_load(eopatch, fs_loader):
+@pytest.mark.parametrize(
+    "save_kwargs, load_kwargs",
+    [
+        ({"path": "/random_path"}, {"path": "/random_path"}),
+        (
+            {"path": "/", "overwrite_permission": OverwritePermission.OVERWRITE_FEATURES},
```
mlubej, line 131: Before the MR the tests actually tested saving the eopatch in the same location and with feature-overwriting permissions. AFAIK, the changes proposed in this MR now always start with an empty FS, so the overwrite_permission parameter shouldn't in fact have any effect here.

zigaLuksic: I have no clue what the old test was actually doing. Judging by the name it was just testing the interaction of saving and loading. There are other tests for adding features and overwriting stuff. But they also save and load. So all in all, the test suite is garbage.

mlubej: Then in this case I would simply limit to testing once with a random path and once with `lazy_loading = False`. Otherwise I feel it might be confusing if we come back to this some time later and expect this to be overwriting something, when in fact it's not. Especially since there are other tests testing this as well, I guess.

zigaLuksic: there are actually so many tests that save+load (like the `save_load_partial`) that we might as well remove this one altogether if you agree

zigaLuksic: done

mlubej (approved):
I checked that eo-grow tests which were failing before due to these issues now also pass

PR #605, opened 2023-03-06
Feature comparison test util

jgersak:


core/eolearn/core/utils/testing.py
```diff
@@ -80,3 +83,15 @@ def _get_feature_shape(
 ) -> Tuple[int, ...]:
     time, height, width, depth = len(timestamps), *config.raster_shape, rng.integers(*config.depth_range)
     return (time, height, width, depth) if ftype.is_temporal() else (height, width, depth)
+
+
+def assert_feature_match(tested_feature: Any, expected_feature: Any) -> None:
+    if type(tested_feature) is not type(expected_feature):
+        assert type(tested_feature) is type(expected_feature)
```
zigaLuksic, line 90: this is really weird... why have the assert in a conditional?
also directly comparing types is problematic for classes and inheritance

core/eolearn/core/utils/testing.py
```diff
@@ -80,3 +83,15 @@ def _get_feature_shape(
...
+
+    if isinstance(tested_feature, np.ndarray):
+        assert_array_equal(tested_feature, expected_feature)
+    elif isinstance(tested_feature, gpd.GeoDataFrame):
+        assert_geodataframe_equal(tested_feature, expected_feature)
```
zigaLuksic, line 95: in the snippet of code that you removed, you can see how we can compare `GeoDataFrames` so that equivalent definitions of CRS are ok. That would probably be better (IO sometimes changes the CRS def to an equivalent one)r.

core/eolearn/core/utils/testing.py
```diff
@@ -80,3 +83,15 @@ def _get_feature_shape(
 ) -> Tuple[int, ...]:
     time, height, width, depth = len(timestamps), *config.raster_shape, rng.integers(*config.depth_range)
     return (time, height, width, depth) if ftype.is_temporal() else (height, width, depth)
+
+
+def assert_feature_match(tested_feature: Any, expected_feature: Any) -> None:
```
zigaLuksic, line 88: change to `assert_feature_data_equal` or `assert_equal_features` or similar

zigaLuksic: and add docstrings or pylint will eat you

core/eolearn/core/utils/testing.py
jgersak: > why have the assert in a conditional? 
you cant simply do `assert False`

probably the best solution is to remove this two lines, because if type are not the same the test fill fail any how ?

zigaLuksic: you actually can do `assert False` :)
you can also have just `assert type(tested_feature) is type(expected_feature)`. But that is again vulnerable to subclasses, so something like `assert isinstance(a, type(b)) or isinstance(b, type(a))` would be needed.

If you remove them, then i'd call `isinstance` on both elements, `assert_arrays_equal` is not very picky about the type of input (but we want to be)

jgersak: > you actually can do assert False :)
you can but flake8 dose not relay like it

> If you remove them, then i'd call isinstance on both elements, assert_arrays_equal is not very picky about the type of input (but we want to be)
kinda like this option :)

core/eolearn/core/utils/testing.py
jgersak: done

core/eolearn/core/utils/testing.py
jgersak: done

core/eolearn/core/utils/testing.py
```diff
@@ -80,3 +83,16 @@ def _get_feature_shape(
 ) -> Tuple[int, ...]:
     time, height, width, depth = len(timestamps), *config.raster_shape, rng.integers(*config.depth_range)
     return (time, height, width, depth) if ftype.is_temporal() else (height, width, depth)
+
+
+def assert_feature_data_equal(tested_feature: Any, expected_feature: Any) -> None:
+    """A test util function for checking if two features are equal."""
```
zigaLuksic, line 89:
```suggestion
    """Asserts that the data of two features is equal. Cases are specialized for common data found in EOPatches."""
```

zigaLuksic (approved):

Issue #578, opened 2023-02-22
[BUG] Failing tests on MacOS related to lock-related EOExecutor tests


mlubej:
**Describe the bug**

Tests mentioned above fail on MacOS. Not sure if they already failed before or not. Probably not many people run this anyway.

**To Reproduce**

1. Git gud job
2. Git gud salary
3. Git gut macbook
4. Run eo-learn tests 

**Expected behavior**

Passing tests

**Environment**

- Python 3.10
- `pyenv` venv environment
- `sh-py` and `eo-learn` installed in edit mode (`develop` versions)

**Stack trace or screenshots**

**Desktop (please complete the following information):**

 - OS: MacOS Ventura 13.2.1

**Additional context**

This might be related to the known issue of MacOS and Windows, where the multiprocessing spawn method has to be set to `fork` via:

```python
import multiprocessing
multiprocessing.set_start_method('fork')
```

This method is the default on Linux, but not on Mac and probably not on Windows as well. This is what I usually have to do to make parallelization work in notebooks and such. 

The mentioned tests also work if I set the method in the failing tests file to `fork`, but I'm not sure if that's the proper way to go about it.

batic:
For some context, see also [bp-33725](https://bugs.python.org/issue33725).

PR #610, opened 2023-03-08
Make sure that the examples use an explicit bbox

zigaLuksic:
Did yall know we had a non-working core example notebook for a while? cause I didn't. But the EONode definitions were ill formed :joy:

### DIFF
--- a/examples/core/CoreOverview.ipynb
+++ b/examples/core/CoreOverview.ipynb
@@ -56,7 +56,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 1,
    "metadata": {
     "scrolled": false
    },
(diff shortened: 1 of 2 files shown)


mlubej (commented):
I agree with the changes, but just pointing out the following:
- on macOS this crashes if you don't set the mp method to `fork` (see #578)
- I get a crash in the final call of `executor.make_report()` with the following trace:

image

This should be caught in https://github.com/sentinel-hub/eo-learn/blob/fix-example-to-use-bbox/visualization/eolearn/visualization/eoexecutor.py#L183 but for some reason this is an OSError in my case. 

First time seeing this, not sure what to do.

zigaLuksic:
> First time seeing this, not sure what to do.

Switch to Linux :shrug:
I snooped around a bit, and both OSError and TypeError get thrown from the `inspect.getsource`. Should I add OSError to `except` in this MR or a separate one?

mlubej:
> Switch to Linux 🤷

image


> Should I add OSError to except in this MR or a separate one?

this one is fine

mlubej (approved):

PR #611, opened 2023-03-08
Move tests away from nonlocated eopatches and remove None as input in IO tasks

zigaLuksic:


core/eolearn/tests/test_eodata.py
```diff
@@ -88,17 +91,19 @@ def test_vector_feature_types() -> None:
...
         eop.bbox = invalid_entry
 
 
 @pytest.mark.parametrize(
     "valid_entry", [["2018-01-01", "15.2.1992"], (datetime.datetime(2017, 1, 1, 10, 4, 7), datetime.date(2017, 1, 11))]
 )
 def test_timestamp_valid_feature_type(valid_entry: Any) -> None:
-    eop = EOPatch()
+    eop = EOPatch(bbox=DUMMY_BBOX, timestamps=valid_entry)
     eop.timestamps = valid_entry
```
mlubej, line 107: can be removed 

```suggestion
```

io/eolearn/tests/test_raster_io.py
```diff
@@ -365,8 +365,7 @@ def test_time_dependent_feature_with_timestamps(test_eopatch):
 @pytest.mark.parametrize("no_data_value, data_type", [(np.nan, float), (0, int), (None, float), (1, np.byte)])
 def test_export_import_sequence(no_data_value, data_type):
     """Tests import and export tiff tasks on generated array with different values of no_data_value."""
-    eopatch = EOPatch()
-    eopatch.bbox = BBox((0, 0, 1, 1), crs=CRS.WGS84)
+    eopatch = EOPatch(bbox=BBox((0, 0, 1, 1), crs=CRS.WGS84))
```
mlubej, line 368: this repeats quite a lot. would it make sense to have a fixture for some empty eopatch instead?

core/eolearn/tests/test_eodata.py
```diff
@@ -88,17 +91,19 @@ def test_vector_feature_types() -> None:
     "invalid_entry", [0, list(range(4)), tuple(range(5)), {}, set(), [1, 2, 4, 3, 4326, 3], "BBox"]
 )
 def test_bbox_feature_type(invalid_entry: Any) -> None:
```
mlubej, line 93: I find this usage of `invalid_entry` slightly annoying and would suggest to rename to `invalid_bbox`/`invalid_timestamps` across the tests, if you agree.

core/eolearn/tests/test_eodata.py
zigaLuksic: why should it be removed? It checks that no error is raised if it's assigned to the attribute (different mechanism kicks AFAIK)

core/eolearn/tests/test_eodata.py
zigaLuksic: sure, i have no attachments to this particular naming

zigaLuksic: done

core/eolearn/tests/test_eodata.py
mlubej: I assumed the attribute assignment is tested elsewhere, here you just moved the timestamps assignment into the EOPatch init. I guess it doesn't hurt, but if this is already tested elsewhere, perhaps the flow of the test becomes weird.

zigaLuksic: love your faith in our test suite :)
but no, this is the only place. However now that i think of it, since the `__init__` is autogenerated by attrs, it could be removed from the initialization :thinking: But that is kinda white-box testing.

io/eolearn/tests/test_raster_io.py
mlubej: ok, let's leave it like it is for now then

mlubej (approved):

PR #612, opened 2023-03-08
Add deprecation warning to EOPatches without a BBox

zigaLuksic:
Sorry, sorry, i know, i did a lot of changes that are not 100% related:
1. Added deprecation warning to init and to setter, added test for it
2. Removed pytest.ini because it was messing with warnings depending on which folder you called `pytest` from
3. Adjusted tests so that the warnings from point 1. is not triggered
4. Adjusted a test for zip task, because it raised a warning about dividing by zero
5. cleaned up requirements :see_no_evil:


mlubej (approved):
I trust you with the requirement changes, other things seem straight-forward.

PR #615, opened 2023-03-09
Improve feature parsing capabilities and deprecate `FeatureTypeSet`

zigaLuksic:


core/eolearn/core/constants.py
```diff
@@ -170,7 +205,23 @@ def file_format(self) -> MimeType:
         return MimeType.JSON
 
 
-class FeatureTypeSet:
+class DeprecatedCollectionClass(type):
+    """A custom EnumMeta class for catching the deprecated Enum members of the FeatureType Enum class."""
```
mlubej, line 209:
```suggestion
    """A custom meta class for raising a warning when collections of the deprecated FeatureTypeSet class are used."""
```

core/eolearn/core/eodata.py
```diff
@@ -699,7 +699,7 @@ def consolidate_timestamps(self, timestamps: List[dt.datetime]) -> Set[dt.dateti
         good_timestamp_idxs = [idx for idx, _ in enumerate(self.timestamps) if idx not in remove_from_patch_idxs]
         good_timestamps = [date for idx, date in enumerate(self.timestamps) if idx not in remove_from_patch_idxs]
 
-        relevant_features = set(FeatureTypeSet.TEMPORAL_TYPES) - {FeatureType.TIMESTAMPS}
+        relevant_features = filter(lambda ftype: ftype.is_temporal() and ftype != FeatureType.TIMESTAMPS, FeatureType)
```
mlubej, line 702: might be more elegant:

```suggestion
        relevant_features = filter(lambda ftype: ftype.is_temporal() and not ftype.is_meta(), FeatureType)
```

core/eolearn/core/core_tasks.py
```diff
@@ -277,7 +277,7 @@ def __init__(
         self.shape: Union[None, Tuple[int, int, int], Tuple[int, int, int, int]]
 
         try:
-            self.shape_feature = self.parse_feature(shape)
+            self.shape_feature = self.parse_feature(shape)  # type: ignore[arg-type]
```
mlubej, line 280: Can you provide more detail here what exactly `mypy` is fussing about?

geometry/eolearn/geometry/transformations.py
```diff
@@ -131,11 +131,9 @@ def _parse_main_params(
     ) -> Tuple[Union[GeoDataFrame, Tuple[FeatureType, str]], Tuple[FeatureType, str]]:
         """Parsing first 2 parameters - what vector data will be used and in which raster feature it will be saved"""
         if not _is_geopandas_object(vector_input):
-            vector_input = self.parse_feature(vector_input, allowed_feature_types=FeatureTypeSet.VECTOR_TYPES)
+            vector_input = self.parse_feature(vector_input, allowed_feature_types=lambda fty: fty.is_vector())
 
-        parsed_raster_feature = self.parse_feature(
-            raster_feature, allowed_feature_types=FeatureTypeSet.RASTER_TYPES_3D.union(FeatureTypeSet.RASTER_TYPES_4D)
-        )
+        parsed_raster_feature = self.parse_feature(raster_feature, allowed_feature_types=lambda fty: fty.is_image())
```
mlubej, line 136: beautiful

features/eolearn/features/feature_manipulation.py
```diff
@@ -48,9 +49,8 @@ def __init__(
         :param filter_func: A callable that takes a numpy evaluates to bool.
         :param filter_features: A collection of features which will be filtered into a new EOPatch
         """
-        self.feature = self.parse_feature(
-            feature, allowed_feature_types=FeatureTypeSet.TEMPORAL_TYPES.difference([FeatureType.VECTOR])
-        )
+
+        self.feature = self.parse_feature(feature, allowed_feature_types=lambda fty: fty.is_temporal())
```
mlubej, line 53: Should be:

```suggestion
        self.feature = self.parse_feature(feature, allowed_feature_types=lambda fty: fty.is_temporal() and not fty.is_vector())
```

core/eolearn/core/eotask.py
```diff
@@ -67,11 +77,30 @@ def __call__(self, *eopatches, **kwargs):
...
+        return parse_feature(feature, eopatch, allowed_feature_types)
+
+    @staticmethod
+    def parse_features(
+        feature: FeaturesSpecification,
+        eopatch: Optional[EOPatch] = None,
+        allowed_feature_types: Union[EllipsisType, Iterable[FeatureType], Callable[[FeatureType], bool]] = ...,
+    ) -> List[FeatureSpec]:
+        """See `eolearn.core.utils.parse_features`."""
+        return parse_features(feature, eopatch, allowed_feature_types)
```
mlubej, line 96:
```suggestion
    def parse_features(
        features: FeaturesSpecification,
        eopatch: Optional[EOPatch] = None,
        allowed_feature_types: Union[EllipsisType, Iterable[FeatureType], Callable[[FeatureType], bool]] = ...,
    ) -> List[FeatureSpec]:
        """See `eolearn.core.utils.parse_features`."""
        return parse_features(features, eopatch, allowed_feature_types)
```

features/eolearn/features/feature_manipulation.py
zigaLuksic: that was my first thought as well, but if you look at L86 there is a method for filtering vector features (???)

core/eolearn/core/core_tasks.py
zigaLuksic: `shape: Union[Tuple[int, ...], FeatureSpec]`, so it's not a suitable input for the feature parser. Which explains why the whole thing is in a `try except` statement.

features/eolearn/features/feature_manipulation.py
mlubej: It seems that the feature used for filtering shouldn't be a vector, but vectors are allowed to be present under `filter_features`, where the filter is then only applied. under `_get_filtered_indices` only temporal array-type data is expected, so perhaps I'd even use `fty.is_temporal() and fty.is_array()` in this case, but the outcome is the same.

mlubej: I see that even the timestamps feature is allowed, but it's quite ugly since the variable used in ``SimpleFilterTask` is quite generic and I think often used `_get_filtered_indices` is `img`, which is then improper naming, but ultimately my final suggestion above using `is_array()` is not correct, so ignore it.

core/eolearn/core/core_tasks.py
mlubej: horrible, but resolving for now.

features/eolearn/features/feature_manipulation.py
zigaLuksic: ah i see, i completely missed that there is a feature and features :see_no_evil: 
we also need to keep in mind that this task can do filtering on timestamps, which might break once we remove it from `FeatureType`

Reverted to previous allowed types and included the suggested test

mlubej (approved):

PR #621, opened 2023-03-13
Apply patch generator

jgersak:


core/eolearn/tests/test_core_tasks.py
```diff
@@ -44,30 +44,25 @@
...
+    patch = generate_eopatch(
+        {
+            FeatureType.DATA: ["CLP"],
+            FeatureType.MASK: ["CLM"],
+            FeatureType.MASK_TIMELESS: ["mask", "LULC", "RANDOM_UINT8"],
+            FeatureType.SCALAR: ["values", "CLOUD_COVERAGE"],
+        }
+    )
+    bands_config = PatchGeneratorConfig(depth_range=(8, 9))
+    patch.data["bands"] = generate_eopatch((FeatureType.DATA, "bands"), config=bands_config).data["bands"]
```
zigaLuksic, line 63: i do not like this :thinking:

core/eolearn/tests/test_core_tasks.py
```diff
@@ -44,30 +44,25 @@
...
+    patch.data["CLP_S2C"] = np.zeros((5, 98, 151, patch.data["CLP"].shape[3]), dtype=np.int64)
```
zigaLuksic, line 64: any reason for using zeros with `np.int64`?

core/eolearn/tests/test_eodata.py
```diff
@@ -18,19 +18,20 @@
...
-    eop.mask["ones"] = np.ones((2, 6, 6, 1), dtype=int)
-    eop.mask["twos"] = np.ones((2, 3, 3, 2), dtype=int) * 2
-    eop.mask_timeless["threes"] = np.ones((3, 3, 1), dtype=np.uint8) * 3
+    eop = generate_eopatch(
+        {
+            FeatureType.DATA: ["bands", "zeros"],
+            FeatureType.MASK: ["ones", "twos"],
+            FeatureType.MASK_TIMELESS: ["threes"],
+        }
+    )
```
zigaLuksic, line 34: the names kinda dont make sense anymore :see_no_evil: but i think we can leave it as is, unless you have some good ideas.

core/eolearn/tests/test_eodata_io.py
```diff
@@ -44,16 +44,16 @@
 
 @pytest.fixture(name="eopatch")
 def eopatch_fixture():
-    eopatch = EOPatch(bbox=DUMMY_BBOX)
-    mask = np.zeros((3, 3, 2), dtype=np.int16)
-    data = np.zeros((2, 3, 3, 2), dtype=np.int16)
-    eopatch.mask_timeless["mask"] = mask
-    eopatch.data["data"] = data
-    eopatch.timestamps = [datetime.datetime(2017, 1, 1, 10, 4, 7), datetime.datetime(2017, 1, 4, 10, 14, 5)]
```
zigaLuksic, line 52: how many timestamps are generated by default? :thinking:

ml_tools/eolearn/tests/test_sampling.py
```diff
@@ -134,11 +133,8 @@ def test_get_mask_of_samples(small_image: np.ndarray, n_samples: Dict[int, int])
 
 @pytest.fixture(name="eopatch")
 def eopatch_fixture(small_image: np.ndarray) -> EOPatch:
-    t, h, w, d = 10, *small_image.shape, 5
-    eopatch = EOPatch(bbox=BBox((0, 0, 1, 1), CRS(3857)))
-    eopatch.data["bands"] = np.arange(t * h * w * d).reshape(t, h, w, d)
-    eopatch.mask_timeless["raster"] = small_image.reshape(small_image.shape + (1,))
```
zigaLuksic, line 140: this is lost

core/eolearn/tests/test_core_tasks.py
jgersak: do you have better idea how to make current generator output feature with X bands ... (i spent some time here also do not like it but did not found beater solution :( )

core/eolearn/tests/test_core_tasks.py
jgersak: not really just need this dimensions ...

core/eolearn/tests/test_eodata.py
jgersak: have the same problem maybe standard: CLP, mask, CLM, LULC ...

core/eolearn/tests/test_eodata_io.py
jgersak: 5 why most of the test dose not really care about that ...

core/eolearn/tests/test_core_tasks.py
zigaLuksic: is there any reason why it has to be that specific amount of bands?
In that case i'd generate something by hand, so that it is clear what precisely it is

core/eolearn/tests/test_core_tasks.py
zigaLuksic: this setup is a bit weird for me. Why does it need to be of that precise shape? why does it need to match CLP?

core/eolearn/tests/test_eodata.py
zigaLuksic: can be "A", "B", "C" as well i guess.

core/eolearn/tests/test_eodata_io.py
zigaLuksic: do two or three test runs of this module, then change the config to two timestamps and try it again. If there is no significant difference in speed (so the difference is under let's say 15%) then keep it as is, otherwise we can think if it would be better to switch back to 2.

eodata IO is not a very speedy submodule for testing, but most issues come from the mockS3. Still, don't want to make it slower.

core/eolearn/tests/test_core_tasks.py
jgersak: it is used for merging on 1 dim ... so dim 2-4 needs to be the same ...

jgersak: bands were moved in speared fixture, can do same thing for `test_merge_features` but is going to be ugly

core/eolearn/tests/test_eodata_io.py
jgersak: no significant difference

ml_tools/eolearn/tests/test_sampling.py
jgersak: test dose not assert what output should be, but just asserts if mask and resolute are aligned

core/eolearn/tests/test_eodata.py
jgersak: renamed

core/eolearn/tests/test_core_tasks.py
```diff
@@ -44,34 +44,32 @@
...
+            FeatureType.SCALAR: ["values", "CLOUD_COVERAGE"],
+        }
+    )
+    patch.data["CLP_S2C"] = np.zeros((5, 98, 151, patch.data["CLP"].shape[3]), dtype=np.int64)
+
     patch.meta_info["something"] = np.random.rand(10, 1)
     return patch
 
 
+@pytest.fixture(name="patch_bands")
```
zigaLuksic, line 68: the name is a bit odd. `eopatch_to_explode` perhaps?

core/eolearn/tests/test_core_tasks.py
zigaLuksic: cant you just do `patch.data["CLP_S2C"] = np.zeros_like(patch.data["CLP"].shape[3]))`

core/eolearn/tests/test_eodata.py
```diff
@@ -320,9 +321,9 @@ def test_equals() -> None:
 @pytest.mark.parametrize(
     "feature, expected_dim",
     [
-        [(FeatureType.DATA, "zeros"), (3, 3)],
-        [(FeatureType.MASK, "ones"), (6, 6)],
-        [(FeatureType.MASK_TIMELESS, "threes"), (3, 3)],
+        [(FeatureType.DATA, "A"), (98, 151)],
+        [(FeatureType.MASK, "C"), (98, 151)],
+        [(FeatureType.MASK_TIMELESS, "E"), (98, 151)],
```
zigaLuksic, line 326: all the test cases have the same spatial dimension, not the best

core/eolearn/tests/test_core_tasks.py
```diff
@@ -44,34 +44,32 @@
...
+        {
+            FeatureType.DATA: ["bands", "CLP"],
+            FeatureType.MASK: ["CLM"],
+            FeatureType.MASK_TIMELESS: ["mask", "LULC", "RANDOM_UINT8"],
+            FeatureType.SCALAR: ["values", "CLOUD_COVERAGE"],
+        }
+    )
+    patch.data["CLP_S2C"] = np.zeros_like(patch.data["CLP"])
+
     patch.meta_info["something"] = np.random.rand(10, 1)
```
zigaLuksic, line 64:
```suggestion
    patch.meta_info["something"] = "beep boop"
```
Just to make sure the patches don't change during runs

core/eolearn/tests/test_core_tasks.py
```diff
@@ -426,19 +424,23 @@ def test_explode_bands(
         assert_array_equal(patch[new_feature], patch[feature][..., bands])
 
 
-def test_extract_bands(patch: EOPatch) -> None:
+def test_extract_bands(eopatch_to_explode: EOPatch) -> None:
```
zigaLuksic, line 427: ah, I see it is used here as well
:thinking: 

The name `eopatch_with_many_bands` is kinda stupid, so maybe just keep it as it is

core/eolearn/tests/test_eodata.py
```diff
@@ -317,18 +318,27 @@ def test_equals() -> None:
     assert eop1 != eop2
 
 
+@pytest.fixture(scope="function", name="eopatch_spatial_dim")
+def eopatch_spatial_dim_fixtuer() -> EOPatch:
```
zigaLuksic, line 322:
```suggestion
def eopatch_spatial_dim_fixture() -> EOPatch:
```

core/eolearn/tests/test_utils/test_parsing.py
```diff
@@ -181,6 +166,15 @@ def test_allowed_feature_types_iterable(test_input: FeaturesSpecification, allow
         FeatureParser(features=test_input, allowed_feature_types=allowed_types)
 
 
+@pytest.fixture(name="eopatch", scope="module")
+def eopatch_fixture():
+    patch = generate_eopatch(
+        {FeatureType.DATA: ["data", "CLP"], FeatureType.MASK: ["data", "IS_VALID"], FeatureType.MASK_TIMELESS: ["LULC"]}
+    )
+    patch.meta_info = {"something": "else"}
```
zigaLuksic, line 174: i see this comes up a few times :thinking: 
If you want you can play with the `generate_eopatch` to also create dummy  `meta_info`. But for that I'd keep it very simple, like `patch[random_meta_feature] = _random_string(rng)` and
```python
import string

def _random_string(rng):
    random_letters = rng.choice(string.ascii_lettes, 20)
    return "".join(random_letters)
```

If you think that is not worth it, we can keep it as it is

zigaLuksic (approved):

PR #625, opened 2023-03-31
Use evalscript generation and timestamps availability utility from sh py

mlubej:
Updated code for using the evalscript generation and timestamps availability utility from sh-py.

- Needs https://github.com/sentinel-hub/sentinelhub-py/pull/441 to be merged first
- Tests were cleaned up since the functionality is now tested at the source


io/eolearn/io/sentinelhub_process.py
```diff
@@ -645,40 +579,21 @@ def __init__(
         data_collection: DataCollection = DataCollection.DEM,
         **kwargs: Any,
     ):
+        dem_band = data_collection.bands[0].name
+        renamed_feature: Tuple[FeatureType, str, str]
```
zigaLuksic, line 583: this whole feature juggling is a bit odd, can you tell me what is the reason behind it?

mlubej: The whole `SentinelHubDemTask` is based off of `SentinelHubEvalscriptTask` which accepts features as `FeaturesSpecification`, but internally converts them all to renamed features, where the original name is supposed to be used in the extraction from the band, while the new name serves as the feature name in the eopatch

In this DEM task I found it annoying that the name was hardcoded before, and at the same time I had to take into account the option that the user provides a custom name for the requested feature. I found it easiest to achieve this by using a renamed feature from scratch and just populate the values accordingly based on all possible cases.

zigaLuksic: i checked the code outside of the diff and it's not as messy as I feared. It seems like you can get rid of the value error by using  `allowed_feature_types` of the parser.

mlubej: thanks for checking, updated

zigaLuksic (approved):

PR #633, opened 2023-04-21
Avoid pylint warning

zigaLuksic:
We've had some issues with pylint + schedule in the recent days, this is a semi-exploratory MR


zigaLuksic:
Discovered that the issue was in the editable installation flag (we knew that it can cause issues, but it never did on runners up until now).

will open a different MR to master

PR #634, opened 2023-04-21
Fix pylint CI

zigaLuksic:


.github/workflows/ci_action.yml
```diff
@@ -52,10 +52,8 @@ jobs:
 
       - name: Install packages
         run: |
-          sudo apt-get update
-          sudo apt-get install -y build-essential libgdal-dev graphviz proj-bin gcc libproj-dev libspatialindex-dev
```
mlubej, line 56: how come this is not needed anymore?

zigaLuksic: it never was for pylint, we probably forgot to remove it. Likely to have happened when mypy and pylint were separated from regular tests into their own job.

mlubej (approved):

PR #644, opened 2023-05-10
Refactor interpolation

zigaLuksic:
While doing refactoring i put in some additional regression tests just to make sure things are identical to how they were.
Behaviour should not have been altered at all (but by the gods, i was tempted!).

I suggest that the review is done commit by commit, I tried to group them together in smaller chunks that are more sensible.

I plan on doing some more changes in the near future, but don't want to make a MR that is too large. I believe we can structure it even better, but i'll settle for removing 70 lines of code :D
The tests show no unexpected slowdown (in fact it is a small speedup, but 10% can be just statistical noise)


features/eolearn/tests/test_interpolation.py
```diff
@@ -230,41 +230,47 @@ def execute(self, eopatch):
...
+                unknown_value=5,
+                bounds_error=False,
+                copy_features=[
+                    (FeatureType.MASK, "IS_VALID"),
+                    (FeatureType.DATA, "NDVI", "NDVI_OLD"),
+                    (FeatureType.MASK_TIMELESS, "LULC"),
+                ],
+            ),
+            result_len=69,
+            expected_statistics={},
```
devisperessutti, line 250: this probably has to do with changes to tests, but where are the values to compare with now?

devisperessutti (commented):
very neat, same functionality with much cleaner and clearer code :clap:

not a huge fan of removing docstrings, although of helper methods, but I understand the intent. Minor clarification about the tests.

features/eolearn/tests/test_interpolation.py
zigaLuksic: these test cases are only used to check that the `copy_features` parameter works as intended. Even the old `test_copied_fields` (where these cases are used) never checked the statistics at all.

zigaLuksic:
> not a huge fan of removing docstrings, although of helper methods, but I understand the intent. Minor clarification about the tests.

I mostly removed the descriptions of parameters, and even those mostly in cases where they were trivial, e.g. `:param mask: Mask array`.

devisperessutti (approved):

PR #653, opened 2023-05-16
Adjust EOPatch machinery

zigaLuksic:
With `bbox` and `timestamp` slowly leaving the kingdom of `FeatureType` i wanted to see if it makes more sense to have them as properties.

Well then it clashed with `attrs`. And when removed, it turned out that we've already been eager-loading these two features (due to `__repr__` loading them. So we made that more explicit. This resulted in massive simplification of general get and set methods.

Sadly it is a bit larger than i wanted it to be, because it took time to see if it's actually worth it at all. I'm sure this would also let us build upon in the future.

- removes `attrs` style definition
- moves special handling of bbox and timestamps to properties (meta_info has to be separate for now, until it becomes a regular feature w.r.t IO)
- simplify python magic
- some general refactoring

Also keep in mind that 'yknow what, i'm not sure we should do this' is completely fine as a response, this started out as an experiment and I kinda like the results.


core/eolearn/core/eodata.py
```diff
@@ -254,100 +231,104 @@ class EOPatch:
...
-        cannot be parsed correctly it raises an error.
-
-        :raises: TypeError, ValueError
-        """
-
-        if feature_type is FeatureType.BBOX and (value is None or isinstance(value, BBox)):
-            if value is None:
-                warn(MISSING_BBOX_WARNING, category=EODeprecationWarning, stacklevel=2)
-            return value
+    def _parse_to_datetime(value: Union[dt.date, str]) -> dt.datetime:
```
batic, line 294: Perhaps the `parse_time` from sh-py could/should be used here?

core/eolearn/core/eodata.py
```diff
@@ -435,6 +408,7 @@ def __contains__(self, key: object) -> bool:
             "`(feature_type, feature_name)` pairs."
         )
 
+    @deprecated_function(EODeprecationWarning, "Use the `merge` method instead.")
```
batic, line 411: When was this deprecated? Perhaps it could be thrown out.

core/eolearn/core/eodata.py
```diff
@@ -254,100 +231,104 @@ class EOPatch:
...
-    vector_timeless: _FeatureDictGeoDf = attr.ib(factory=_FeatureDictGeoDf.empty_factory(FeatureType.VECTOR_TIMELESS))
-    meta_info: _FeatureDictJson = attr.ib(factory=_FeatureDictJson.empty_factory(FeatureType.META_INFO))
-    bbox: Optional[BBox] = attr.ib(default=None)
-    timestamps: List[dt.datetime] = attr.ib(factory=list)
-
-    def __attrs_post_init__(self) -> None:
-        if self.bbox is None:
-            warn(MISSING_BBOX_WARNING, category=EODeprecationWarning, stacklevel=2)
+    # establish types of property value holders
+    _timestamps: List[dt.datetime]
```
batic, line 235: Does this mean we can't create EOpatches without timestamps anymore?

core/eolearn/core/eodata.py
```diff
@@ -254,100 +231,104 @@ class EOPatch:
...
+    _bbox: Optional[BBox]
```
batic, line 236: Why difference with `_timestamps`?

core/eolearn/core/eodata.py
```diff
@@ -254,100 +231,104 @@ class EOPatch:
...
+    _meta_info: Union[FeatureIOJson, _FeatureDictJson]
```
batic, line 237: Same as for `bbox`. I think all three should be optional, until we remove the deprecation warning and make  the `bbox` and `timestamps` non-optional.

batic (commented):
I think the approach is better, but some things went above my head, so I'd appreciate somebody else having another look.

core/eolearn/core/eodata.py
zigaLuksic: That is because the 'missing-value' for `timestamps` is currently `[]` and not `None`. So if the user does not provide a value for `timestamps` it gets set to `[]`.

Switching the default to `None` is planned, but also codebreaking.

core/eolearn/core/eodata.py
zigaLuksic: Similarly to timestamps the 'missing' default here is `{}`

core/eolearn/core/eodata.py
zigaLuksic: Splendid idea, not sure how i missed it :see_no_evil:

core/eolearn/core/eodata.py
zigaLuksic: yesterday :D
I've never seen `patch1 + patch2` used in actual code and I'd kinda prefer not to

core/eolearn/core/eodata.py
batic: 👍

mlubej (approved):
Went through the changes with @zigaLuksic and they make sense, I especially like that the interface becomes more intuitive for the users.

There are some things which were harder to grasp for me, especially when it comes to featureIO objects, pickling, and any consequences related to this, but I believe here we already set a line to not force the PR out of scope.

PR #654, opened 2023-05-17
Extract EOPatch.merge as function

zigaLuksic:


core/eolearn/core/eodata_merge.py
```diff
@@ -31,7 +31,7 @@
 OperationInputType = Union[Literal[None, "concatenate", "min", "max", "mean", "median"], Callable]
 
 
-def merge_eopatches(
+def merge_eopatch_content(
```
batic, line 34: Should this become a "private" method in `eodata.py`?

zigaLuksic: i guess it could, we can just silence pylint when it complains about importing

zigaLuksic: done

PR #658, opened 2023-05-18
Move merge_eopatches to eodata_merge

zigaLuksic:
After sleeping on it it feels more correct to handle the circular import inside the deprecated method, rather than having weird wrappers. Also goes along with your instinct that the wrapper could be removed.

### DIFF
--- a/core/eolearn/core/__init__.py
+++ b/core/eolearn/core/__init__.py
@@ -21,7 +21,8 @@ from .core_tasks import (
     SaveTask,
     ZipFeatureTask,
 )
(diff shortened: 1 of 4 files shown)


batic (approved):
Ah. 🤤 

Better, despite the not so innocent looking cyclic-import pylint ignore block.

zigaLuksic:
> Ah. drooling_face
> 
> Better, despite the not so innocent looking cyclic-import pylint ignore block.

Agreed, but since all the tests pass i'm not really worried that much. Especially since we do the same with visualization (a few lines lower)

PR #661, opened 2023-05-19
Enable `ruff`

zigaLuksic:
In a future PR we can also get rid of flake8 now.
Also we might want to migrate `isort` over to `ruff`, even the isort creators [advise so](https://news.ycombinator.com/item?id=34062974)

I suggest you proceed commit by commit.
Some flags are more radical, so i opened new issues for those.

P.S. VSCode has a ruff extension written by the author of ruff

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,11 @@ repos:
             --ignore-init-module-imports,
           ]
 
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: "v0.0.269"
+    hooks:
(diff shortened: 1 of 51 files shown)


zigaLuksic:
I went throught he list of all codes that ruff has, and tried to enable any that made sense. In some cases there are too many mistakes to fix and not much benefit (linting docs for example). In other cases the things they check are not that important to us (checking Django specific code etc.).
But I tried all of them, enabled what made sense, delayed any that are too major for this MR (and opened an issue for them). We can periodically go and check if some new ones also make sense.

PR #662, opened 2023-05-19
Split github actions per trigger and make use of reusable workflows

mlubej:
Changes:
- convert the old ci_action YAML file to a reusable action
- create new caller workflows for the `schedule` trigger and the `push/pull_requests` triggers

This should fix the issue of using the workflow file from the default branch and running the tests on the `develop` branch, sometimes resulting in a failed build.

After merging, this MR needs to be merged to `develop` as well, since the caller workflow here references the workflow on develop, which should be reusable, but currently isn't yet.


mlubej:
Seems there are some issues with tests, perhaps these issues were resolved/updated in `develop`, or perhaps were fixed with updated versions of the `develop` workflow. I guess that means that perhaps it's better to make these changes on top of `develop` and just wait it out.

@zigaLuksic let me know if I should prepare this PR on `develop`

zigaLuksic:
> Seems there are some issues with tests, perhaps these issues were resolved/updated in `develop`, or perhaps were fixed with updated versions of the `develop` workflow. I guess that means that perhaps it's better to make these changes on top of `develop` and just wait it out.
> 
> @zigaLuksic let me know if I should prepare this PR on `develop`

Yes, these changes were fixed on develop.
We can either:
1. open a mirror MR with these changes on develop, and make sure the tests go through there. We then close that one and merge this one in faith that it will perform just as  well
2. Just merge to develop since the release of the new version is near and the schedule build remains broken for a few days.

Also, since you're the last person changing the schedule line, you'll be getting emails about the schedule builds. I can adjust the time afterwards a bit, so that i'm registered as the person to email to if you'd prefer to be spam free.

mlubej:
> Yes, these changes were fixed on develop. We can either:
> 
> 1. open a mirror MR with these changes on develop, and make sure the tests go through there. We then close that one and merge this one in faith that it will perform just as  well
> 2. Just merge to develop since the release of the new version is near and the schedule build remains broken for a few days.
> 

I think option 2 should be completely fine, I'll prepare a mirror MR and we can just wait for the release.

> Also, since you're the last person changing the schedule line, you'll be getting emails about the schedule builds. I can adjust the time afterwards a bit, so that i'm registered as the person to email to if you'd prefer to be spam free.

I'll add a commend to the schedule line and you can remove it afterwards, so you're the last one that made the change

zigaLuksic:
> I'll add a commend to the schedule line and you can remove it afterwards, so you're the last one that made the change

I'll just move the schedule forward by a minute :wink: that works for sure, not 100% that a comment would register properly

mlubej:
Closing in favor of #663

PR #664, opened 2023-05-22
Add nbqa-ruff and remove flake8

zigaLuksic:
I decided that we should remove `flake8` because `ruff` covers all the plugins that we use and can therefore be used as a drop-in replacement (together with black, according to ruff FAQ).

I don't want to keep two configurations for the same thing. But i am willing to reconsider the last commit.


pyproject.toml
```diff
@@ -18,7 +18,17 @@ line_length = 120
...
+ignore = [
+    "C408",   # complains about `dict()` calls, we use them to avoid too many " in the code
+    "SIM117", # wants to always combine `with` statements, gets ugly for us
+    "SIM108", # tries to aggresively inline `if`, not always readable
+    "A003",   # complains when ATTRIBUTES shadow builtins, we have objects that implement `filter` and such
+    "COM812", # trailing comma missing, fights with black
+    "PD011",  # suggests `.to_numpy` instead of `.values`, also does this for non-pandas objects...
+    # potentially fixable
+    "B904", # want `raise ... from None` instead of just `raise ...`
+    "B028", # always demans a stacklevel argument when warning
```
batic, line 30: demans -> demands

Issue #637, opened 2023-05-02
[FEAT] TDigestTask handle nans


alunacob:
**What is the problem? Please describe.**

Current implementation fails when the data values in the feature where TDigest is computed, are all nan for a timestamp

**Here's the solution**

Manage properly NaNs 

**Alternatives**

Skip them from the TDigest computation

**Additional context**

The issue was produced while using meteobluerasterTask with meteo variables from 1970, which did not yield any data for some periods of time. The TDigestTask would fail for the whole time range, even if from 1984, data was available.

mlubej:
Was merged in #667, closing

PR #667, opened 2023-05-23
TDigestTask enabling nanfiltering and insertion of own processing functions.

meengel:
As written in the [enhancement request](https://github.com/sentinel-hub/eo-learn/issues/637), there is a necessity for enabling the filtering of nan-values within the features used for the tdigest representation. The Pull-Request enables the Task to do so and, additionally, enables the users to provide own processing functions such as monthly per year accumulations, for example.

@zigaLuksic @batic Would be really happy if you could include it to eo-learn! Later, one should definitely take care of the underlying tdigest-implementation but that's something for another Pull-Request.

### DIFF
--- a/ml_tools/eolearn/ml_tools/tdigest.py
+++ b/ml_tools/eolearn/ml_tools/tdigest.py
@@ -1,6 +1,5 @@
 """
 The module provides an EOTask for the computation of a T-Digest representation of an EOPatch.
-Requires installation of `eolearn.ml_tools[TDIGEST]`.
 
 Copyright (c) 2017- Sinergise and contributors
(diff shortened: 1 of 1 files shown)


ml_tools/eolearn/ml_tools/tdigest.py
```diff
@@ -1,6 +1,5 @@
 """
 The module provides an EOTask for the computation of a T-Digest representation of an EOPatch.
-Requires installation of `eolearn.ml_tools[TDIGEST]`.
```
zigaLuksic, line 3: This can be reverted

zigaLuksic:
Due to time constraints with releasing the new version I will merge this and fix the issues myself this time around.

PR #668, opened 2023-05-23
Improve permissions and deprecate OVERWRITE_PATCH

zigaLuksic:
sad that we need to copy-paste that meta-class (or complicate it to oblivion), but sometimes life is that way

### DIFF
--- a/core/eolearn/core/constants.py
+++ b/core/eolearn/core/constants.py
@@ -303,6 +303,21 @@ class OverwritePermission(Enum):
     - `OVERWRITE_PATCH` - Overwrite entire content of saved EOPatch and replace it with the new content.
     """
 
-    ADD_ONLY = 0
-    OVERWRITE_FEATURES = 1
-    OVERWRITE_PATCH = 2
+    ADD_ONLY = "ADD_ONLY"
+    OVERWRITE_FEATURES = "OVERWRITE_FEATURES"
+    OVERWRITE_PATCH = "OVERWRITE_PATCH"
+
+    @classmethod
+    def _missing_(cls, value: object) -> "OverwritePermission":
+        permissions_mapping = {0: "ADD_ONLY", 1: "OVERWRITE_FEATURES", 2: "OVERWRITE_PATCH"}
+        if isinstance(value, int) and value in permissions_mapping:
+            deprecation_msg = (
+                f"Please use strings to instantiate overwrite permissions, e.g., instead of {value} use"
+                f" {permissions_mapping[value]!r}"
+            )
+            warnings.warn(deprecation_msg, category=EODeprecationWarning, stacklevel=3)
+
+            return cls(permissions_mapping[value])
+        if isinstance(value, str) and value.upper() in cls._value2member_map_:
+            return cls(value.upper())
+        return super()._missing_(value)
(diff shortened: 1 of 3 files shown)


core/eolearn/core/constants.py
```diff
@@ -292,7 +293,30 @@ class FeatureTypeSet(metaclass=DeprecatedCollectionClass):
...
+    """A custom EnumMeta class for catching the deprecated Enum members of the FeatureType Enum class."""
+
+    def __getattribute__(cls, name: str) -> Any:  # noqa[N805]
+        return super().__getattribute__(_warn_and_adjust_permissions(name))
+
+    def __getitem__(cls, name: str) -> Any:  # noqa[N805]
+        return super().__getitem__(_warn_and_adjust_permissions(name))
+
+    def __call__(cls, value: str, *args: Any, **kwargs: Any) -> Any:  # noqa[N805]
+        return super().__call__(_warn_and_adjust_permissions(value), *args, **kwargs)
```
mlubej, line 316: perhaps it would make sense to change the `name` parameter to `permission` or something. `name` is still used from the FeatureType enum

core/eolearn/core/eodata_io.py
```diff
@@ -106,8 +106,10 @@ def save_eopatch(
     # Data must be collected before any tinkering with files due to lazy-loading
     data_for_saving = list(_yield_features_to_save(eopatch, eopatch_features, patch_location))
 
-    if overwrite_permission is OverwritePermission.OVERWRITE_PATCH and patch_exists:
-        _remove_old_eopatch(filesystem, patch_location)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=EODeprecationWarning)
```
mlubej, line 110: is this ignored because the warning will be shown already by using the permission in the first place? was the warning triggering twice?

core/eolearn/tests/test_eodata_io.py
```diff
@@ -269,8 +288,12 @@ def test_cleanup_different_compression(fs_loader, eopatch):
     with fs_loader() as temp_fs:
         temp_fs.makedir(folder)
 
-        save_compressed_task = SaveTask(folder, filesystem=temp_fs, compress_level=9, overwrite_permission=1)
-        save_noncompressed_task = SaveTask(folder, filesystem=temp_fs, compress_level=0, overwrite_permission=1)
+        save_compressed_task = SaveTask(
+            folder, filesystem=temp_fs, compress_level=9, overwrite_permission="OVERWRITE_FEATURES"
+        )
+        save_noncompressed_task = SaveTask(
+            folder, filesystem=temp_fs, compress_level=0, overwrite_permission="OVERWRITE_FEATURES"
```
mlubej, line 295: Should there also be some usage of the lowercase permissions?

core/eolearn/core/constants.py
zigaLuksic: those are taken from the Enum implementation. The `name` is the name of the enum constant. I dont think we should be losing sleep over this one

core/eolearn/core/eodata_io.py
zigaLuksic: this is ignored because here WE are using the constant. Even if the user provide `ADD_ONLY` we check it against `OVERWRITE_PATCH` which triggers the warning.

core/eolearn/tests/test_eodata_io.py
zigaLuksic: not sure, i consider that a favor to users since it's not explicitly mentioned anywhere. Intended use is `OverwritePermission.ADD_ONLY`.

we could if you consider it relevant enough

mlubej (approved):

PR #669, opened 2023-05-23
Replace isort with ruff

zigaLuksic:
Ruff can also integrate with VSCode, but beware, both `ruff` and `isort` trigger if you have both enabled (and for me isort won). I disabled the isort extension for eo-learn and stuff works


pyproject.toml
```diff
@@ -58,12 +46,28 @@ select = [
...
     "NPY", # numpy
+    "I",   # isort
 ]
 target-version = "py38"
 
 [tool.ruff.mccabe]
 max-complexity = 15
 
+[tool.ruff.isort]
+# required-imports = ["from __future__ import annotations"]
```
mlubej, line 57: is this comment intentional?

PR #670, opened 2023-05-24
Use new-style type annotations

zigaLuksic:
I would suggest you go through commit by commit and skip the big two (the new-style union autofix and new style builtins autofix).

I tried to simplify and reorder the pyproject.toml. The ruff section got very long due to us using nearly every code it has :smile_cat:


mlubej (approved):
😻 😻 😻

PR #673, opened 2023-05-25
Fix remaining annotations to new-style

zigaLuksic:
Some files were missed by the FA flag

### DIFF
--- a/core/eolearn/core/constants.py
+++ b/core/eolearn/core/constants.py
@@ -6,9 +6,11 @@ For the full list of contributors, see the CREDITS file in the root directory of
 
 This source code is licensed under the MIT license, see the LICENSE file in the root directory of this source tree.
 """
+from __future__ import annotations
+
 import warnings
 from enum import Enum, EnumMeta
-from typing import Any, Optional, TypeVar
+from typing import Any, TypeVar
 
 from sentinelhub import BBox, MimeType
 from sentinelhub.exceptions import deprecated_function
@@ -170,7 +172,7 @@ class FeatureType(Enum, metaclass=EnumWithDeprecations):
         """True if FeatureType stores a dictionary of numpy.ndarrays. False otherwise."""
         return self.is_array()
 
-    def ndim(self) -> Optional[int]:
+    def ndim(self) -> int | None:
         """If given FeatureType stores a dictionary of numpy.ndarrays it returns dimensions of such arrays."""
         if self.is_array():
             return {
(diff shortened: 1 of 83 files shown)


mlubej (commented):
Why were they missed? What kind of manual action did you have to do to fix the remaining ones?

zigaLuksic:
the flag doesnt seem to be working 100% of the time
i searched for all occurences of `Union`, `Dict`, `Tuple` and `List` (the main changes)

mlubej:
> the flag doesnt seem to be working 100% of the time
> i searched for all occurences of `Union`, `Dict`, `Tuple` and `List` (the main changes)

Perhaps then it _would_ be better to just add `from __future__ import annotations` everywhere? especially since you mentioned that the type aliases don't work correctly without it in the docs.

I'm mentioning this because it seems you missed a few which seem to get corrected if I add the future annotation import under the `tool.ruff.isort` required-imports:
- https://github.com/sentinel-hub/eo-learn/blob/fix-remaining-annotations/core/eolearn/tests/test_utils/test_testing.py#L35
- https://github.com/sentinel-hub/eo-learn/blob/fix-remaining-annotations/core/eolearn/tests/test_utils/test_testing.py#L78
- https://github.com/sentinel-hub/eo-learn/blob/fix-remaining-annotations/docs/source/conf.py#LL381C44-L381C44
- https://github.com/sentinel-hub/eo-learn/blob/fix-remaining-annotations/docs/source/conf.py#L435-L441

Unless you intentionally left those out

zigaLuksic:
> > the flag doesnt seem to be working 100% of the time
> > i searched for all occurences of `Union`, `Dict`, `Tuple` and `List` (the main changes)
> 
> Perhaps then it _would_ be better to just add `from __future__ import annotations` everywhere? especially since you mentioned that the type aliases don't work correctly without it in the docs.
> 
> I'm mentioning this also because it seems you missed a few which seem to get corrected if I add the future annotation import under the `tool.ruff.isort` required-imports:
> 
>     * https://github.com/sentinel-hub/eo-learn/blob/fix-remaining-annotations/core/eolearn/tests/test_utils/test_testing.py#L35
> 
>     * https://github.com/sentinel-hub/eo-learn/blob/fix-remaining-annotations/core/eolearn/tests/test_utils/test_testing.py#L78
> 
>     * https://github.com/sentinel-hub/eo-learn/blob/fix-remaining-annotations/docs/source/conf.py#LL381C44-L381C44
> 
>     * https://github.com/sentinel-hub/eo-learn/blob/fix-remaining-annotations/docs/source/conf.py#L435-L441
> 
> 
> Unless you intentionally left those out

The `test_testing` was not intentional, the conf.py was (i'd add an ignore to it)
For eo-learn it might actually be reasonable to do this. In sh-py and eo-grow there are conflicts due to certain dataclasses, but in eo-learn this should work just fine

zigaLuksic:
adding the required import adds it to a further 58 files. I assume we want this to be autofixable?

mlubej:
> adding the required import adds it to a further 58 files. I assume we want this to be autofixable?

I would say so, yes. This ignores `__init__.py` files?

zigaLuksic:
> > adding the required import adds it to a further 58 files. I assume we want this to be autofixable?
> 
> I would say so, yes. This ignores `__init__.py` files?

I made it ignore init files, the conf files and notebooks.

mlubej (approved):

PR #681, opened 2023-06-08
Add zarr format support (no partial IO yet)

zigaLuksic:
This just establishes Zarr as a possible save format for raster data.

the `compress_level` parameter on all FeatureIOs is still annoying me, but once again i will delay dealing with it until we add partial IO (since the signature will likely change then anyway)


core/eolearn/core/eodata_io.py
```diff
@@ -579,6 +589,55 @@ def _write_to_file(cls, data: BBox, file: BinaryIO | gzip.GzipFile, _: str) -> N
...
+        cls._check_dependencies_imported(feature_path)
+
+        path = f"{feature_path}.{cls.get_file_extension()}"
+        store = cls._get_mapping(path, filesystem)
+        zarr.save(store, data)
+
+    @staticmethod
+    def _check_dependencies_imported(path: str) -> None:
+        if not all(dep in sys.modules for dep in ["zarr", "s3fs"]):
+            msg = f"Encountered use of Zarr for {path}, but missing dependencies. Please install `zarr` and `s3fs`."
```
mlubej, line 637:
```suggestion
            msg = f"Encountered use of Zarr for {path} with missing dependencies. Please install `zarr` and `s3fs`."
```

core/eolearn/core/eodata_io.py
```diff
@@ -156,41 +170,35 @@ def _save_single_feature(save_spec: tuple[type[FeatureIO[T]], T, str], *, filesy
...
     current_compress_level: int,
 ) -> None:
-    """Removes files that should have been overwritten but were not due to different compression levels."""
-
-    def has_different_compression(path: str | None) -> bool:
-        return path is not None and MimeType.GZIP.matches_extension(path) != (current_compress_level > 0)
+    """Removes files that should have been overwritten but were not due to different file extensions."""
+    feature_paths = (path for _, ftype_dict in preexisting_files.features.items() for _, path in ftype_dict.items())
+    meta_paths = (preexisting_files.bbox, preexisting_files.meta_info, preexisting_files.timestamps)
+    split_paths = (path.split(".", maxsplit=1) for path in filter(None, itertools.chain(meta_paths, feature_paths)))
```
mlubej, line 180: What if a `.` occurs in the path before the extension? sounds like it could be problematic, but not sure if anyone would go and persist with such a choice.

core/eolearn/core/eodata_io.py
```diff
@@ -156,41 +170,35 @@ def _save_single_feature(save_spec: tuple[type[FeatureIO[T]], T, str], *, filesy
 
 def remove_redundant_files(
```
mlubej, line 171: nice, much more elegant

mlubej (commented):
first batch, didn't check the tests yet

core/eolearn/core/eodata_io.py
zigaLuksic: i also had such concerns
I feel like I'll have to write a custom `split_ext` function that takes into account multi-extensions.

mlubej: It could be done with regex and [lookahead](https://www.geeksforgeeks.org/python-regex-lookahead/) `(?==)`

```python
import re
complex_path = r'a\b/c.d\f.g/e.np.gzip'
re_ext = re.compile(r'\.(?=[^/\\]*$)')
```

The above searches for a match where there is a literal dot, and no path characters (`/` or the escaped `\` for windows) after the dot, then performs a single split.

```
In [49]:  re_ext.split(s, maxsplit=1)
Out[49]: ['a\\b/c.d\\f.g/e', 'np.gzip']
```

zigaLuksic: does it handle dot-hidden files? :)
i prefer to avoid regex when possible since its hard to debug.
I'm cooking up a solution with pathlib, which is stronger than os.path or fs.path

zigaLuksic: done

mlubej (approved):
Tests look good as well, and the extension extraction makes sense to me.

As you say, the compression and partial IO will be handled in separate MRs for zarr

👍

PR #682, opened 2023-06-09
Use a conditional in CI to avoid --cov unless coverage is uploaded

zigaLuksic:
seems to shave off around 30s

### DIFF
--- a/.github/workflows/ci_action.yml
+++ b/.github/workflows/ci_action.yml
@@ -108,7 +108,11 @@ jobs:
           sentinelhub.config \
             --sh_client_id "${{ secrets.SH_CLIENT_ID }}" \
             --sh_client_secret "${{ secrets.SH_CLIENT_SECRET }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_action.yml
```diff
@@ -108,7 +108,11 @@ jobs:
           sentinelhub.config \
             --sh_client_id "${{ secrets.SH_CLIENT_ID }}" \
             --sh_client_secret "${{ secrets.SH_CLIENT_SECRET }}"
-          pytest -m "not geopedia" --cov --cov-report=term --cov-report=xml
+          if [ ${{ github.event_name }} == 'push' ]; then
```
mlubej, line 111: in what situations are now tests faster? most of the tests were push-related anyway, were they not?

zigaLuksic: Most tests are pull-request based, and those are the ones that we need ASAP. push triggers only on develop and master (so one per MR)

mlubej (approved):

PR #683, opened 2023-06-12
Partial EOPatch loading

zigaLuksic:
Adds partial loading from a temporal perspective.

We have yet to reach a consensus on what to do with vector data

### DIFF
--- a/core/eolearn/tests/test_eodata_io.py
+++ b/core/eolearn/tests/test_eodata_io.py
@@ -539,7 +539,7 @@ def test_zarr_and_numpy_combined_loading(eopatch):
 @mock_s3
 @pytest.mark.parametrize("fs_loader", FS_LOADERS)
 @pytest.mark.parametrize("use_zarr", [True, False])
-@pytest.mark.parametrize("temporal_selection", [])
+@pytest.mark.parametrize("temporal_selection", [None, slice(None, 3), slice(2, 4, 2), [3, 4]])
 def test_partial_loading(fs_loader: type[FS], eopatch: EOPatch, use_zarr: bool, temporal_selection):
     _skip_when_appropriate(fs_loader, use_zarr)
     with fs_loader() as temp_fs:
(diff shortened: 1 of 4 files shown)


core/eolearn/tests/test_eodata_io.py
```diff
@@ -533,3 +534,35 @@ def test_zarr_and_numpy_combined_loading(eopatch):
...
+        assert_array_equal(full_patch.data["data"][adjusted_selection, ...], partial_patch.data["data"])
+        assert_array_equal(full_patch.mask_timeless["mask"], partial_patch.mask_timeless["mask"])
+        assert_array_equal(np.array(full_patch.timestamps)[adjusted_selection], partial_patch.timestamps)
+
+
+@mock_s3
+@pytest.mark.parametrize("fs_loader", FS_LOADERS)
+@pytest.mark.parametrize("use_zarr", [True, False])
+@pytest.mark.parametrize("temporal_selection", [[3, 4, 10]])
+def test_partial_loading_fails(fs_loader: type[FS], eopatch: EOPatch, use_zarr: bool, temporal_selection):
```
mlubej, line 562: Just a comment, sometimes it's becoming a bit hard to keep in mind all combinations of the parameters and all the conditions for skipping tests, so it's a bit of a hassle to interpret these tests. Is this a temporary solution due to the issues with zarr in some aspects of FS?

mlubej (commented):
It's nice to see how you set everything up to easily implement this partial saving/loading 👏

core/eolearn/tests/test_eodata_io.py
zigaLuksic: another option is to construct `(fs_loader, use_zarr)` pairs as input, but then you need to unpack etc.
In general if you use a zarr backend, you should have the skip helper function.

Currently the helper function will:
- skip the zarr tests if there is no zarr installed
- skip the moto + zarr combo, because it doesn't work correctly ATM

I agree it's a hassle

core/eolearn/tests/test_eodata_merge.py
```diff
@@ -186,3 +186,16 @@ def test_lazy_loading(test_eopatch_path):
...
+
+def test_temporally_independent_merge(test_eopatch_path):
+    full_patch = EOPatch.load(test_eopatch_path)
+    part1 = EOPatch.load(test_eopatch_path, temporal_selection=slice(None, 10))
+    part2 = EOPatch.load(test_eopatch_path, temporal_selection=slice(10, None))
+
+    # remove ill-formed data, wrong temporal size
+    del full_patch[(FeatureType.DATA, "REFERENCE_SCENES")]
+    del part1[(FeatureType.DATA, "REFERENCE_SCENES")]
+    del part2[(FeatureType.DATA, "REFERENCE_SCENES")]
```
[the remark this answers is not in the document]
zigaLuksic: REFERENCE_SCENES has a temporal dimension of 5, while the EOPatch has 68 timestamps. All other features have the correct temporal dimension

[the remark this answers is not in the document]
zigaLuksic: (leaving this in breaks merging because it works under the assumption that the temporal dimension is correct)

core/eolearn/tests/test_eodata_merge.py
```diff
@@ -186,3 +186,16 @@ def test_lazy_loading(test_eopatch_path):
...
     assert isinstance(eop1.mask.get("CLM"), np.ndarray)
     assert isinstance(eop1.mask_timeless._get_unloaded("LULC"), FeatureIO)  # noqa: SLF001
+
+
+def test_temporally_independent_merge(test_eopatch_path):
+    full_patch = EOPatch.load(test_eopatch_path)
+    part1 = EOPatch.load(test_eopatch_path, temporal_selection=slice(None, 10))
+    part2 = EOPatch.load(test_eopatch_path, temporal_selection=slice(10, None))
+
+    # remove ill-formed data, wrong temporal size
```
mlubej, line 196: just to make it a bit more intuitive, since it wasn't clear if it's an issue of the test or from before

```suggestion
    # remove pre-existing ill-formed data of wrong temporal size
```

mlubej:
@zigaLuksic mentioned in private re. the parameter name `temporal_selection`. I guess I would prefer `timestamp_filter`, now that the topic was brought up.

PR #687, opened 2023-06-19
Switch "no timestamps" value to `None`

zigaLuksic:
This establishes a separation of Patches with no temporal dimension and those with a zero-sized temporal dimension.

Also forces users to think about the possibility of not having timestamps available.

The `get_timestamps` method was added because most of the time we want an error to be raised if timestamps are missing.

### DIFF
--- a/core/eolearn/core/eodata.py
+++ b/core/eolearn/core/eodata.py
@@ -285,7 +285,9 @@ class EOPatch:
 
     @timestamps.setter
     def timestamps(self, value: Iterable[dt.datetime] | None) -> None:
-        if isinstance(value, Iterable) and all(isinstance(time, (dt.date, str)) for time in value):
+        if value is None:
+            self._timestamps = None
+        elif isinstance(value, Iterable) and all(isinstance(time, (dt.date, str)) for time in value):
             self._timestamps = [parse_time(time, force_datetime=True) for time in value]
         else:
             raise TypeError(f"Cannot assign {value} as timestamps. Should be a sequence of datetime.datetime objects.")
@@ -381,7 +383,7 @@ class EOPatch:
         if feature_type == FeatureType.BBOX:
             raise ValueError("The BBox of an EOPatch should never be undefined.")
         if feature_type == FeatureType.TIMESTAMPS:
-            self[feature_type] = []
+            self[feature_type] = None
         else:
             self[feature_type] = {}
 
--- a/core/eolearn/core/eodata.py
+++ b/core/eolearn/core/eodata.py
@@ -222,7 +222,7 @@ class EOPatch:
     """
 
     # establish types of property value holders
-    _timestamps: list[dt.datetime]
+    _timestamps: list[dt.datetime] | None
     _bbox: BBox | None
     _meta_info: FeatureIOJson | _FeatureDictJson
 
@@ -265,10 +265,10 @@ class EOPatch:
         )
         self.meta_info: MutableMapping[str, Any] = _FeatureDictJson(meta_info or {}, FeatureType.META_INFO)
         self.bbox = bbox
-        self.timestamps = timestamps or []
+        self.timestamps = timestamps
 
     @property
-    def timestamp(self) -> list[dt.datetime]:
+    def timestamp(self) -> list[dt.datetime] | None:
         """A property for handling the deprecated timestamp attribute."""
         warn(TIMESTAMP_RENAME_WARNING, category=EODeprecationWarning, stacklevel=2)
         return self.timestamps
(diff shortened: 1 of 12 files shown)


core/eolearn/core/eodata.py
```diff
@@ -279,17 +279,25 @@ def timestamp(self, value: list[dt.datetime]) -> None:
...
         """A property for handling the `timestamps` attribute."""
         return self._timestamps
 
     @timestamps.setter
-    def timestamps(self, value: Iterable[dt.datetime]) -> None:
-        if isinstance(value, Iterable) and all(isinstance(time, (dt.date, str)) for time in value):
+    def timestamps(self, value: Iterable[dt.datetime] | None) -> None:
+        if value is None:
+            self._timestamps = None
+        elif isinstance(value, Iterable) and all(isinstance(time, (dt.date, str)) for time in value):
```
mlubej, line 290: The `isinstance` check here is not in line with the expected type of the setter, it would make more sense if this is `isinstance(time, dt.datetime)`, right? Otherwise the error message should be fixed

Seems like this has worked so far because `isinstance(time, dt.date)` is `True` even if `time` is of the `dt.datetime` type.

core/eolearn/core/eodata_merge.py
```diff
@@ -121,15 +121,15 @@ def _return_if_equal_operation(arrays: np.ndarray) -> bool:
 
 def _merge_timestamps(
     eopatches: Sequence[EOPatch], reduce_timestamps: bool
-) -> tuple[list[dt.datetime], list[np.ndarray]]:
+) -> tuple[list[dt.datetime] | None, list[np.ndarray]]:
     """Merges together timestamps from EOPatches. It also prepares a list of masks, one for each EOPatch, how
     timestamps should be ordered and joined together.
     """
-    timestamps_per_eopatch = [eopatch.timestamps for eopatch in eopatches]
+    timestamps_per_eopatch = [eopatch.timestamps or [] for eopatch in eopatches]
```
mlubej, line 128: So, for the purpose of merging, if `eopatch.timestamps` is `None` it's temporarily converted to `[]`.

However, the final merged eopatch could still in principle have temporal features with empty temporal slices, in that case the `[]` timestamp should be eligible, right?

core/eolearn/core/eodata.py
```diff
@@ -279,17 +279,25 @@ def timestamp(self, value: list[dt.datetime]) -> None:
...
             self._timestamps = [parse_time(time, force_datetime=True) for time in value]
         else:
             raise TypeError(f"Cannot assign {value} as timestamps. Should be a sequence of datetime.datetime objects.")
 
+    def get_timestamps(self, message_on_failure: str = "The EOPatch does not contain timestamps.") -> list[dt.datetime]:
```
mlubej, line 295: I understand the need for this function is to not change all the types and interactions of the timestamps object with the rest of the code in the repo. 

However, I see some possible issues:
- people would use `get_timestamps`, thinking it's the same as `eop.timestamps`
- people would use `eop.timestamps` and not knowing about this function, they would implement their own versions on how to handle `None` cases

I have no idea if there is any way how to minimize the chances of the above happening, at least more than what you already documented. 

So I guess this is more a comment than anything else, just trying to see if you had similar concerns, but probably there is nothing else to do here but to let developers find this out on their own, or correct them during PRs.

mlubej, line 295:
```suggestion
    def get_timestamps(self, message_on_failure: str = "This EOPatch does not contain timestamps.") -> list[dt.datetime]:
```

core/eolearn/core/eodata.py
```diff
@@ -279,17 +279,25 @@ def timestamp(self, value: list[dt.datetime]) -> None:
...
+        """Returns the `timestamps` attribute if timestamps if the EOPatch is temporally defined. Fails otherwise."""
```
mlubej, line 296:
```suggestion
        """Returns the `timestamps` attribute if the EOPatch is temporally defined. Fails otherwise."""
```

core/eolearn/core/eodata.py
zigaLuksic: `dt.date` is technically a superclass of `dt.datetime` (although it is actually not, it's weird)
The error message tells the users the desired input, but here we do some convenience parsing for them (we're all lazy and use `"2021-05-01"` instead of actual datetime objects).

The type of the setter actually doesn't even matter here :see_no_evil: mypy always assumes the type is the same as that in the getter. python....

core/eolearn/core/eodata.py
zigaLuksic: My first thought was to not provide this at all. But then you really do need to do this in a lot of places, so I decided to add this with a way to provide a custom message anyway. I believe the documentation is sufficient to address your first bullet point. As for the second one, no clue how to do that :/

core/eolearn/core/eodata_merge.py
zigaLuksic: you are correct... what a stupid edge case. Will add more firepower here. Great catch

core/eolearn/core/eodata.py
mlubej: Yeah, let's see how it turns out, not much else we can do at the moment without changing a bunch of code.

Unless you think changing the code is the proper way to do it, then we could address this some other time and open a ticket for handling this

mlubej (approved):
No more comments, looks great, but let's wait till tomorrow just so that it sits in my head a bit more :D

core/eolearn/core/eodata.py
zigaLuksic: i dont really see any other way of doing it, but i'm open to suggestions. It has also been raised in the ticket since last week with 0 replies :)

mlubej: You're right, I guess without this users would implement their own None handling, which would then become a pattern converging to the same thing. 

I tried to think if better names could be found, but I couldn't think of any, so let's just go ahead with it as it is.

zigaLuksic: i was thinking about `get_timestamps_or_fail` to be more explicit in what the intended use is, but it is a bit verbose

PR #689, opened 2023-06-20
Use context manager to control backend when making reports

zigaLuksic:


mlubej (approved):
the PR shows a lot of changes, but if I understand correctly, the only difference is the definition of `backend_ctx` and indenting all the relevant lines inside the context. If yes, then it seems fine to me.

I also tested it out and it works as expected now.

zigaLuksic:
Yes, its just indents

PR #690, opened 2023-06-21
Refactor before timestamp save changes

zigaLuksic:
I tried to start with the timestamp saving changes, but I noticed that core tasks have not yet been updated for partial IO. I also refactored how loading is done, so that timestamp loading changes will be easier

### DIFF
--- a/core/eolearn/core/core_tasks.py
+++ b/core/eolearn/core/core_tasks.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 import copy
 from abc import ABCMeta
-from typing import Any, Callable, Iterable, Tuple, Union, cast
+from typing import Any, Callable, Iterable, Literal, Tuple, Union, cast
 
 import fs
 import numpy as np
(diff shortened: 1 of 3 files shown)


core/eolearn/core/core_tasks.py
```diff
@@ -90,7 +90,9 @@ def __init__(
         features: FeaturesSpecification = ...,
         overwrite_permission: OverwritePermission = OverwritePermission.ADD_ONLY,
         compress_level: int = 0,
+        *,
         use_zarr: bool = False,
+        temporal_selection: None | slice | list[int] = None,
```
mlubej, line 95: Not sure if I noticed this before or not, but now there seems to be some redundancy in parameters, right?

If `temporal_selection` is not `None`, then usage of `zarr` is/should be implied. The issue is the init save, I guess. 

Did you have any thought on this already?

zigaLuksic: yeah, you can save in zarr format and not specify anything. If you have a zarr based eo-grow project probably only a few pipelines will work with temporal selection

core/eolearn/core/eodata_io.py
```diff
@@ -248,26 +247,62 @@ def load_eopatch_content(
...
+
+
+def _load_features(
+    filesystem: FS,
+    features: FeaturesSpecification,
+    temporal_selection: None | slice | list[int],
+    err_msg: str,
+    file_information: FilesystemDataInfo,
+) -> dict[tuple[FeatureType, str], FeatureIO]:
+    requested_meta = {ftype for ftype, _ in FeatureParser(features).get_feature_specifications() if ftype.is_meta()}
```
mlubej, line 279: This is repeated from L254, would it make sense to pass it to the `_load_features`?

zigaLuksic: i was thinking if maybe i should make a note there.
this one will vanish when we adjust timestamps saving (since it will no longer rely on `features`)

PR #692, opened 2023-06-21
Automate timestamp saving

zigaLuksic:
The users have control over whether or not timestamps get saved, but it's set to `"auto"` by default which saves/loads timestamps if there are any temporal features being saved/loaded.

In the future any temporal-information-containing eopatches will REQUIRE timestamps. If users dont save/load them, it's their problem, should've left it on "auto".

### DIFF
--- a/core/eolearn/core/core_tasks.py
+++ b/core/eolearn/core/core_tasks.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 import copy
 from abc import ABCMeta
-from typing import Any, Callable, Iterable, Tuple, Union, cast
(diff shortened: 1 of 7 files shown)


zigaLuksic:
We can also switch from `bool | Literal["auto"]` to `Literal["auto", "always"]` to avoid any ill defined patches. Didnt get many responses on the issues

mlubej (approved):
Seems straightforward, I have no comments, aside from those pointed out internally, where I expressed a preference against https://github.com/sentinel-hub/eo-learn/pull/692#issuecomment-1600722165

I share your concern re. the issue of a lot of new parameters, but also still think we are going in the right direction, I just hope there won't be any large surprises

PR #694, opened 2023-06-23
Add warnings for temporally ill defined EOPatches

zigaLuksic:
- fixes an issue where one could 'sneak' ill formed data into the EOPatch at `__init__` (since feature just copied the contents)
- adds a warning when one adds a temporal feature to a patch without timestamps
- adds a warning when one adds a temporal feature to a patch and the dimensions clash with timestamps
- adds a warning when one sets/changes timestamps and the dimensions clash with existing features
- adjusts A LOT of tests to get rid of said warnings


core/eolearn/core/eodata.py
```diff
@@ -78,7 +79,11 @@ def __init__(self, feature_dict: Mapping[str, T | FeatureIO[T]], feature_type: F
         super().__init__()
 
         self.feature_type = feature_type
-        self._content = dict(feature_dict.items())
+
+        # we need trigger parsing and validation
+        self._content: dict[str, T | FeatureIO[T]] = {}
+        for key, value in feature_dict.items():
+            self[key] = value
```
mlubej, line 86: Not sure I understand this

core/eolearn/core/eodata.py
```diff
@@ -162,6 +191,25 @@ def _parse_feature_value(self, value: object, feature_name: str) -> np.ndarray:
 
         return value
 
+    def _update_temporal_dim(self, new_value: int | None) -> None:
+        self._temporal_dim = new_value
+        if self._temporal_dim is None:
+            if self._content:
```
mlubej, line 197: Here you are checking if the features exist, but I'm missing the point where you check if this is a FeatureDictNumpy of temporal features or not.

In other words, I would expect this to be `if self._content and self.feature_type.is_temporal()`

core/eolearn/core/eodata.py
```diff
@@ -162,6 +191,25 @@ def _parse_feature_value(self, value: object, feature_name: str) -> np.ndarray:
...
+                warnings.warn(
+                    f"EOPatch does not have timestamps, but has temporal features of type {self.feature_type}.",
+                    category=TemporalDimensionWarning,
+                    stacklevel=4,
+                )
+            return
+
+        for feature_name, value in self._content.items():
+            if not isinstance(value, FeatureIO) and value.shape[0] != self._temporal_dim:
```
mlubej, line 206: What else can it also be, if not `isinstance(value, FeatureIO)`?

core/eolearn/core/eodata.py
zigaLuksic: `self._content` is just a regular dictionary, no checks are done on it.
So instead of directly filling the `_content` we feed the values through `self[key]` so that our custom `__setitem__` triggers and parses/validates the values. Before you could sneak in anything at init, now you cannot (I added tests to make sure of that)

core/eolearn/core/eodata.py
zigaLuksic: i didn't do this because temporal dimensions are only updated for temporal feature dictionaries. I guess it wouldn't hurt to have another check here as well.

core/eolearn/core/eodata.py
zigaLuksic: it can be an already loaded value. Content has both `FeatureIO[T]` and `T` values (if something is lazy loaded the FeatureIO is replaced with the value itself)

core/eolearn/tests/test_utils/test_raster.py
```diff
@@ -23,6 +23,7 @@
 @pytest.mark.parametrize("nan_ratio", [0, 0.05, 0.1, 0.5, 0.9, 1])
 @pytest.mark.parametrize("dtype", [np.float64, np.float32, np.int16])
 @pytest.mark.parametrize("method", ["linear", "nearest"])
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
```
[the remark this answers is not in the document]
zigaLuksic: yeah, it is redundant, good catch

core/eolearn/tests/test_core_tasks.py
```diff
@@ -80,7 +80,7 @@ def test_copy(task: type[CopyTask], patch: EOPatch) -> None:
     patch_copy.data["bands"][0, 0, 0, 0] += 1
     assert (patch_copy != patch) if task == DeepCopyTask else (patch_copy == patch)
 
-    patch_copy.data["new"] = np.arange(1).reshape(1, 1, 1, 1)
+    patch_copy.data["new"] = np.ones((len(patch_copy.get_timestamps()), 1, 1, 1))
```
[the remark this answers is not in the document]
zigaLuksic: should work,i just did minimal adjustments to get rid of warnings

PR #695, opened 2023-06-23
Temporal selection improvements

zigaLuksic:
Improves partial temporal IO:

- loading can now be done via filter function
- saving can be done with "infer" which figures out where to insert the temporal slices


core/eolearn/core/core_tasks.py
```diff
@@ -109,7 +110,8 @@ def __init__(
             features are being saved.
         :param use_zarr: Saves numpy-array based features into Zarr files. Requires ZARR extra dependencies.
         :param temporal_selection: Writes all of the data to the chosen temporal indices of preexisting arrays. Can be
-            used for saving data in multiple steps for memory optimization.
+            used for saving data in multiple steps for memory optimization. When set to `"infer"` it will match the
+            timestamps of the EOPatch to the timestamps of the stored EOPatch to calculate indices.
```
mlubej, line 114: Can you explain how is this different from `None`? What use case does this tackle?

zigaLuksic: if you use "infer" it will load the pre-existing timestamps, and calculate which temporal indices your current EOPatch describes, and inject the data to appropriate slices.

If you use None it will not do temporally-partial saving and just save a regular array

devisperessutti (approved):
all good, had a discussion offline about options supported, we'll follow up in upcoming MRs

PR #698, opened 2023-06-28
Single package restructure

zigaLuksic:


zigaLuksic:
The [docs](https://eo-learn.readthedocs.io/en/single-package-restructure/index.html) get built and I also tested the docker build locally.

zigaLuksic:
Just to clarify why there are still old folders which only contain requirements and `setup.py`.
I plan on doing a last 'redirect' release of these subpackages. I have some ideas on what to do, but nothing is final, so I'm keeping these files until that is done. By the time 1.5.0 is released those will be removed.

docs/source/conf.py
```diff
@@ -43,11 +42,13 @@
...
-        if "version=" in line:
-            release = line.split("=")[1].strip('", \n').strip("'")
-            version = release.rsplit(".", 1)[0]
+with open(os.path.join(os.path.dirname(__file__), "../../eolearn/__init__.py")) as init_file:
+    for line in init_file:
+        if line.find("__version__") >= 0:
+            release = line.split("=")[1].strip()
+            release = release.strip('"').strip("'")
+
+version = release.rsplit(".", 1)[0]
```
mlubej, line 51: Just noticed this, why is the version string here shortened?

mlubej (commented):
Seems straightforward to me, nice changes. As mentioned, we still need to figure out how to handle the specific tasks and potentially differently group extras.

I'll also manually check the installation

docs/source/conf.py
zigaLuksic: no clue, i asked myself this as well. Couldn't really figure it out so I didnt touch it since it's not really relevant in this scope

PR #700, opened 2023-06-29
Infer can initialize zarr

zigaLuksic:
another :cookie: for my cuties

using `temporal_selection="infer"` when saving can now initialize Zarr arrays (provided timestamps are saved)

### DIFF
--- a/core/eolearn/core/eodata_io.py
+++ b/core/eolearn/core/eodata_io.py
@@ -99,6 +99,14 @@ class FilesystemDataInfo:
                 yield (ftype, fname), path
 
 
+@dataclass
+class InferredSelection:
+    """Temporal selection that was inferred from timestamps. Also carries info on how to initialize Zarr if needed."""
+
+    selection: None | slice | list[int]
+    full_size: int | None  # None means unknown
+
+
 def save_eopatch(
     eopatch: EOPatch,
     filesystem: FS,
@@ -165,9 +173,9 @@ def _infer_temporal_selection(
     filesystem: FS,
     file_information: FilesystemDataInfo,
     eopatch: EOPatch,
-) -> None | slice | list[int]:
+) -> InferredSelection:
     if temporal_selection != "infer":
-        return temporal_selection
+        return InferredSelection(temporal_selection, None)
 
     patch_timestamps = eopatch.get_timestamps("Cannot infer temporal selection. EOPatch to be saved has no timestamps.")
     if file_information.timestamps is None:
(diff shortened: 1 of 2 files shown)


core/eolearn/core/eodata_io.py
```diff
@@ -99,6 +99,14 @@ def iterate_features(self) -> Iterator[tuple[tuple[FeatureType, str], str]]:
                 yield (ftype, fname), path
 
 
+@dataclass
+class InferredSelection:
+    """Temporal selection that was inferred from timestamps. Also carries info on how to initialize Zarr if needed."""
```
batic, line 104: > Temporal selection that was inferred from timestamps.

-> 

Temporal selection that was inferred from existing(?) timestamps.

core/eolearn/core/eodata_io.py
```diff
@@ -766,25 +774,32 @@ def save(
...
+        selection, full_size = temporal_selection.selection, temporal_selection.full_size
+
         try:
             zarray = zarr.open_array(store, "r+")
         except ValueError as error:  # zarr does not expose the proper error...
-            raise IOError(
-                f"Unable to open Zarr array at {path!r}. Saving with `temporal_selection` requires preexisting zarr."
-            ) from error
+            if full_size is not None:
+                zarr_shape = (full_size, *data.shape[1:])
```
batic, line 794: `zarr_shape` and `chunk_size` are very similar; perhaps you could have a var `data_shape = *data.shape[1:]` and then use `(full_size, data_shape)` and `(1, data_shape)` respectively. 

Just an idea, no need to change.

core/eolearn/tests/test_eodata_io.py
```diff
@@ -660,6 +660,30 @@ def test_partial_temporal_saving_into_existing(eopatch: EOPatch, temporal_select
         assert_array_equal(loaded_patch.timestamps, eopatch.timestamps)
 
 
+@mock_s3
+@pytest.mark.parametrize("dtype", [float, np.float32, np.int8])
```
batic, line 664: Would it make sense to parametrize few cases of `temporal_selection` (or e.g. the number of timestamps to be saved?)

batic (commented):
Probably missing some bits as I didn't manage to follow zarr saga closely.

core/eolearn/tests/test_eodata_io.py
zigaLuksic: The `test_partial_temporal_saving_into_existing` covers multiple cases, here I just wanted to test that 'infer' really does initialize an array if it doesn't exist

core/eolearn/core/eodata_io.py
zigaLuksic: i was considering it, but i'd keep it as it is for now since it's not that much simpler

core/eolearn/core/eodata_io.py
zigaLuksic: Changed the name and the description. At some point this was obtained just when the selection is inferred, but now even predefined selections (such as `slice(1,2)`) are transformed into such objects, so `TemporalSelection` is a more fitting name

batic: Sounds much better, yes.

PR #701, opened 2023-06-29
Reduce number of tasks

zigaLuksic:
certain tasks were just slightly specialized base classes. No reason to have those.


mlubej (approved):
Breezy!

So the original tasks already supported everything, we just had extra tasks for convenience?

PR #702, opened 2023-07-03
Fix mypy complaining about cv2

zigaLuksic:
I tested it out and the attribute does exist


zigaLuksic:
merging since all the errors are related to the pydantic 2 feud with ray

PR #704, opened 2023-07-21
Fix mypy issue which causes failing build

mlubej:
Mypy complained about the provided `Booster` class when it was used in place of the custom `ClassifierType` Protocol. Since the `Booster` class in principle corresponds to the `ClassifierType` protocol, I just manually cast the former to the latter.


mlubej:
LightGBM V4 was released last week, causing some issues when using older models (even Boosters!) 

adding to this PR also the constraint for `lightgbm<4.0.0`

mlubej:
There are still failing tests under `core/eolearn/tests/test_eoexecutor.py`, but I can't reproduce them locally, even with a fresh environment.

PR #705, opened 2023-07-25
Adjust EOExecutor tests to pass until ray fixes logging

zigaLuksic:
I was also considering:
- restricting ray versions (introduces nasty restrictions that we must remember to update once ray fixes stuff)
- warning users that logging wont work for bad ray versions (but it really shouldn't be our responsibility to clean up after them)
- trying to patch loggers (lots of work, too much for a known bug that is worked on)

This option will make our tests pass, if ray 2.6.2 still doesn't fix the mistake we will be notified, and if we forget about this and it stays in the tests for another 3 years it does no harm.

### DIFF
--- a/core/eolearn/tests/test_eoexecutor.py
+++ b/core/eolearn/tests/test_eoexecutor.py
@@ -13,6 +13,7 @@ import multiprocessing
 import os
 import tempfile
 import time
+from contextlib import suppress
 from logging import FileHandler
 
 import pytest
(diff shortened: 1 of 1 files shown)


zigaLuksic:
tests are failing due to lgbm stuff, but it is the only test that fails

mlubej (approved):
I agree with the temporary proposed solution and that other alternatives are not so optimal.

PR #708, opened 2023-07-27
Simple Feature Parsers

zigaLuksic:
1. fixes `CopyTask` to correctly copy timestamps
2. `EOPatch.get_features` no longer returns bbox & timestamps
3. feature parsers no longer return bbox & timestamp features
4. `SimpleFilterTask` can still be used on timestamps by providing `feature="timestamps"`, looking for better ideas
5. Replaced types. I also noticed that a lot of tasks had `SingleFeatureSpecification` but they did not use feature renaming inputs, so i corrected those to `tuple[FeatureType, str]`. The whole renaming thing should be removed at one point...

Some test cases had to be removed as they are no longer valid (such as deleting timestamps via `RemoveFeatureTask`)

### DIFF
--- a/core/eolearn/core/types.py
+++ b/core/eolearn/core/types.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 import sys
 
 # pylint: disable=unused-import
-from typing import Dict, Iterable, Optional, Sequence, Tuple, Union
+from typing import Dict, Iterable, Sequence, Tuple, Union
 
 from .constants import FeatureType
 
@@ -28,17 +28,10 @@ else:
 
 # DEVELOPER NOTE: the #: comments are applied as docstrings
 
-#: Specification describing a single feature
-FeatureSpec: TypeAlias = Tuple[FeatureType, str]
-#: Specification describing a feature with its current and desired new name
-FeatureRenameSpec: TypeAlias = Tuple[FeatureType, str, str]
+SingleFeatureSpec: TypeAlias = Union[Tuple[FeatureType, str], Tuple[FeatureType, str, str]]
 
-SingleFeatureSpec: TypeAlias = Union[FeatureSpec, FeatureRenameSpec]
-
-SequenceFeatureSpec: TypeAlias = Sequence[
-    Union[SingleFeatureSpec, FeatureType, Tuple[FeatureType, Optional[EllipsisType]]]
-]
-DictFeatureSpec: TypeAlias = Dict[FeatureType, Union[None, EllipsisType, Iterable[Union[str, Tuple[str, str]]]]]
+SequenceFeatureSpec: TypeAlias = Sequence[Union[SingleFeatureSpec, FeatureType, Tuple[FeatureType, EllipsisType]]]
+DictFeatureSpec: TypeAlias = Dict[FeatureType, Union[EllipsisType, Iterable[Union[str, Tuple[str, str]]]]]
 MultiFeatureSpec: TypeAlias = Union[
     EllipsisType, FeatureType, Tuple[FeatureType, EllipsisType], SequenceFeatureSpec, DictFeatureSpec
 ]
--- a/core/eolearn/core/types.py
+++ b/core/eolearn/core/types.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 import sys
 
 # pylint: disable=unused-import
-from typing import Dict, Iterable, Literal, Optional, Sequence, Tuple, Union
+from typing import Dict, Iterable, Optional, Sequence, Tuple, Union
 
 from .constants import FeatureType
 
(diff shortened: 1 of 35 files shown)


core/eolearn/core/types.py
```diff
@@ -28,18 +28,10 @@
...
 # DEVELOPER NOTE: the #: comments are applied as docstrings
 
-#: Specification describing a single feature
-FeatureSpec: TypeAlias = Union[Tuple[Literal[FeatureType.BBOX, FeatureType.TIMESTAMPS], None], Tuple[FeatureType, str]]
-#: Specification describing a feature with its current and desired new name
-FeatureRenameSpec: TypeAlias = Union[
-    Tuple[Literal[FeatureType.BBOX, FeatureType.TIMESTAMPS], None, None], Tuple[FeatureType, str, str]
-]
-SingleFeatureSpec: TypeAlias = Union[FeatureSpec, FeatureRenameSpec]
+SingleFeatureSpec: TypeAlias = Union[Tuple[FeatureType, str], Tuple[FeatureType, str, str]]
```
mlubej, line 31: So, if I understand correctly, you decided to drop the simple `FeatureSpec` and `FeatureRenameSpec` because they are straightforward and don't need type aliases?

geometry/eolearn/geometry/transformations.py
```diff
@@ -221,7 +221,7 @@ def _get_raster_shape(self, eopatch: EOPatch) -> tuple[int, int]:
                 return self.raster_shape
 
             ftype, fname = self.parse_feature(self.raster_shape, allowed_feature_types=lambda fty: fty.is_array())
-            return eopatch.get_spatial_dimension(ftype, cast(str, fname))  # cast verified in parser
+            return eopatch.get_spatial_dimension(ftype, fname)  # cast verified in parser
```
mlubej, line 224: comment can be removed

io/eolearn/io/sentinelhub_process.py
```diff
@@ -442,7 +441,7 @@ def __init__(
             self.requested_bands = parse_data_collection_bands(data_collection, bands)
 
         self.requested_additional_bands = []
-        self.additional_data: list[FeatureRenameSpec] | None = None
+        self.additional_data: list[tuple[FeatureType, str, str]] | None = None
         if additional_data is not None:
             self.additional_data = parse_renamed_features(additional_data)  # parser gives too general type
```
mlubej, line 446: is the comment here still relevant?

io/eolearn/io/sentinelhub_process.py
```diff
@@ -542,7 +541,9 @@ def _extract_additional_features(
         self, eopatch: EOPatch, images: Iterable[np.ndarray], shape: tuple[int, ...]
     ) -> None:
         """Extracts additional features from response into an EOPatch"""
-        additional_data = cast(List[FeatureRenameSpec], self.additional_data)  # verified by `if` in _extract_data
+        additional_data = cast(
+            List[Tuple[FeatureType, str, str]], self.additional_data
+        )  # verified by `if` in _extract_data
```
mlubej, line 546: You could probably drop the `cast` part if you replace with:

```suggestion
        if self.additional_data is None:
            return None
```

ml_tools/eolearn/ml_tools/tdigest.py
```diff
@@ -121,8 +121,8 @@ def _is_output_ftype(feature_type: FeatureType, mode: ModeTypes, pixelwise: bool
 
 
 def _looper(
-    in_feature: list[FeatureSpec], out_feature: list[FeatureSpec], eopatch: EOPatch
-) -> Generator[tuple[FeatureSpec, FeatureSpec, np.ndarray], None, None]:
+    in_feature: list[tuple[FeatureType, str]], out_feature: list[tuple[FeatureType, str]], eopatch: EOPatch
+) -> Generator[tuple[tuple[FeatureType, str], tuple[FeatureType, str], np.ndarray], None, None]:
```
mlubej, line 125: since it repeats four times in a very limited space, perhaps it would make sense to define a local type alias

core/eolearn/core/types.py
zigaLuksic: yeah, we might consider having a `Feature` alias as we do in `eo-grow` just because it makes the signatures a bit nicer rather than `tuple[FeatureType, str]`

but the rename stuff can be dropped, almost never used anyway

ml_tools/eolearn/ml_tools/tdigest.py
zigaLuksic: we can replace it if we decide to keep a `Feature` alias, otherwise it can be as it is, it's an extra anyway

io/eolearn/io/sentinelhub_process.py
zigaLuksic: removed, code simplified

io/eolearn/io/sentinelhub_process.py
zigaLuksic: did something similar

geometry/eolearn/geometry/transformations.py
zigaLuksic: done

mlubej:
regarding

> SimpleFilterTask can still be used on timestamps by providing feature="timestamps", looking for better ideas

I don't have any better ideas. Perhaps a separate task in its own? but probably it doesn't make sense and is completely fine as it is.

mlubej (approved):
all good in the hood, the changes make sense and I like how it simplifies the types, I can't wait to see the implications on eo-grow

PR #715, opened 2023-08-17
Switch default resize backend

zigaLuksic:
As discussed, this removes Pillow from the list of requirements. It is still installed as part of sh-py, but we might get rid of it in the future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,7 @@ coregistration = [
 ] # using headless version of opencv, otherwise gitlab runners fail
 features = [
     "numba>=0.53.0",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -67,7 +67,7 @@ coregistration = [
 ] # using headless version of opencv, otherwise gitlab runners fail
 features = [
     "numba>=0.53.0",
-    "pillow>=9.1.0",
+    "opencv-python-headless",
```
mlubej, line 70: nice

mlubej (approved):

PR #716, opened 2023-08-18
Remove partial loading for numpy arrays

zigaLuksic:
We decided that since we do not support partial numpy saving, we shouldn't support partial numpy loading either.

### DIFF
--- a/eolearn/core/eodata_io.py
+++ b/eolearn/core/eodata_io.py
@@ -233,6 +233,15 @@ def _yield_savers(
         yield feature_saver
 
 
+def _get_feature_io_constructor(ftype: FeatureType, use_zarr: bool) -> type[FeatureIO]:
+    """Creates the correct FeatureIO, corresponding to the FeatureType."""
+    if ftype is FeatureType.META_INFO:
+        return FeatureIOJson
+    if ftype.is_vector():
+        return FeatureIOGeoDf
+    return FeatureIOZarr if use_zarr else FeatureIONumpy
+
+
 def _remove_redundant_files(
     filesystem: FS,
     new_files: list[str],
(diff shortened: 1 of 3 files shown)


eolearn/core/eodata_io.py
```diff
@@ -233,6 +233,15 @@ def _yield_savers(
         yield feature_saver
 
 
+def _get_feature_io_constructor(ftype: FeatureType, use_zarr: bool) -> type[FeatureIO]:
+    """Creates the correct FeatureIO, corresponding to the FeatureType."""
+    if ftype is FeatureType.META_INFO:
+        return FeatureIOJson
+    if ftype.is_vector():
+        return FeatureIOGeoDf
+    return FeatureIOZarr if use_zarr else FeatureIONumpy  # type: ignore[return-value] # not sure why
```
mlubej, line 242: This is probably unnecessary?

eolearn/core/eodata_io.py
```diff
@@ -361,11 +370,23 @@ def _load_features(
...
-    return constructor(path, filesystem)
+        if use_zarr:
+            return FeatureIOZarr(path, filesystem, temporal_selection)
+        if temporal_selection is None:
+            return FeatureIONumpy(path, filesystem)
+        raise IOError(
+            f"Cannot perform loading with temporal selection for numpy data at {path}. Resave feature with"
+            " `use_zarr=True` to enable loading with temporal selections."
+        )
+    return (FeatureIOZarr if use_zarr else FeatureIONumpy)(path, filesystem)
```
mlubej, line 389: - Do we support zarr for timeless features?

eolearn/core/eodata_io.py
zigaLuksic: mypy would tell us if it's not necessary :upside_down_face:  seems like some issue on their side when it comes to resolving inheritance hierarchies (wasnt an issue a while ago, not sure why it popped up?)

eolearn/core/eodata_io.py
zigaLuksic: yeah

eolearn/core/eodata_io.py
mlubej: My bad for not being clear. I thought that the whole function is not necessary, because you updated the code in `_get_feature_io` to be standalone, but I see now that it's used in `_yield_savers`

mlubej (approved):

eolearn/core/eodata_io.py
zigaLuksic: ah yes, i also deleted it at first and was then surprised when tests started failing :see_no_evil:

PR #720, opened 2023-08-24
Better exception grouping in reports

zigaLuksic:
Exceptions are now grouped by exception origin instead of error message (which often contains unique information and is thus unsuitable for grouping).

the summary also contains a list of all failed executions (per error)

(added screenshot of test case, the button toggles visibility)
image

### DIFF
--- a/eolearn/visualization/eoexecutor.py
+++ b/eolearn/visualization/eoexecutor.py
@@ -12,6 +12,7 @@ import base64
 import datetime as dt
 import importlib
 import inspect
+import itertools as it
 import os
 import warnings
 from collections import defaultdict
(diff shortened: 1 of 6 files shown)


eolearn/visualization/eoexecutor.py
```diff
@@ -223,3 +236,21 @@ def _format_datetime(value: dt.datetime) -> str:
     def _format_timedelta(value1: dt.datetime, value2: dt.datetime) -> str:
         """Method for formatting time delta into report"""
         return str(value2 - value1)
+
+
+@dataclass()
+class _ErrorSummary:
```
mlubej, line 242: Is it a standard to put dataclasses to the bottom of the files?

eolearn/visualization/eoexecutor.py
```diff
@@ -97,40 +100,49 @@ def _create_dependency_graph(self) -> str:
...
+        """
 
-        exception_stats: defaultdict[str, defaultdict[str, int]] = defaultdict(lambda: defaultdict(lambda: 0))
+        exception_stats: defaultdict[str, dict[str, _ErrorSummary]] = defaultdict(dict)
 
-        for workflow_results in self.eoexecutor.execution_results:
-            if not workflow_results.error_node_uid:
+        for execution, results, execution_idx in zip(
+            self.eoexecutor.execution_names, self.eoexecutor.execution_results, it.count()
+        ):
```
mlubej, line 113: You could also do 

```suggestion
        for execution_idx, (execution, results) in enumerate(zip(
            self.eoexecutor.execution_names, self.eoexecutor.execution_results)):
```

and loose the extra import

eolearn/visualization/eoexecutor.py
```diff
@@ -97,40 +100,49 @@ def _create_dependency_graph(self) -> str:
...
 
     def _to_ordered_stats(
-        self, exception_stats: defaultdict[str, defaultdict[str, int]]
-    ) -> list[tuple[str, str, list[tuple[str, int]]]]:
+        self, exception_stats: defaultdict[str, dict[str, _ErrorSummary]]
+    ) -> list[tuple[str, str, list[_ErrorSummary]]]:
         """Exception stats get ordered by nodes in their execution order in workflows. Exception stats that happen
         for the same node get ordered by number of occurrences in a decreasing order.
+
+        Returns tuples of form (name, uid, [exception_origin, example_message, num_occurences])
```
mlubej, line 136: Shouldn't this be:

```suggestion
        Returns tuples of form (name, uid, [_error_summary])
```

eolearn/visualization/eoexecutor.py
```diff
@@ -223,3 +236,21 @@ def _format_datetime(value: dt.datetime) -> str:
...
         """Method for formatting time delta into report"""
         return str(value2 - value1)
+
+
+@dataclass()
+class _ErrorSummary:
+    """Contains data for errors of a node."""
+
+    origin: str
+    example_message: str
```
mlubej, line 246: The name of this field isn't intuitive to me. I guess you named it like this because it represents one of the multiple same error messages? why not just `error_message`?

mlubej (commented):
Can you also provide a screenshot?

eolearn/visualization/eoexecutor.py
zigaLuksic: we dont have any standards for this. Since it's a private class I didnt really bother with placement and added it on the bottom.

eolearn/visualization/eoexecutor.py
zigaLuksic: at some point this could fit into one line, but now we can remove it i guess, gotta see how the lines break

eolearn/visualization/eoexecutor.py
zigaLuksic: the error messages are not necessarily the same

Imagine specifying the folder wrong when loading eopatches. You would get 45 error messages generated by `f"Cannot find BBOX at {path}."` and since the paths would be unique, so would the error messages. This is the reason why we're now grouping by origin and not message.

zigaLuksic:
> Can you also provide a screenshot?

edited description to include screenshot

mlubej (approved):

PR #722, opened 2023-08-25
Split cloud mask task

mlubej:
This is one part of multiple cloud-mask-split MRs. This is just to split the simplified cloud mask task from the complex legacy one.

In the next MR we will use the classifier code from `s2cloudless`

### DIFF
--- a/eolearn/mask/cloud_mask.py
+++ b/eolearn/mask/cloud_mask.py
@@ -41,18 +41,7 @@ class ClassifierType(Protocol):
 
 
 class CloudMaskTask(EOTask):
-    """Cloud masking with the s2cloudless model. Outputs a cloud mask, while the output of cloud probabilities
-    is optional.
-
-    Example usage:
-    .. code-block:: python
-        task = CloudMaskTask(
-            data_feature = (FeatureType.DATA, 'BANDS'),
-            valid_data_feature = (FeatureType.MASK, 'IS_DATA'),
-            output_mask_feature=(FeatureType.MASK, 'CLM_S2C'),
-            output_proba_feature=(FeatureType.DATA, 'CLP_S2C')
-        )
-    """
+    """Cloud masking with the s2cloudless model. Outputs a cloud mask and optionally the cloud probabilities."""
 
     MODELS_FOLDER = os.path.join(os.path.dirname(__file__), "models")
     CLASSIFIER_NAME = "pixel_s2_cloud_detector_lightGBM_v0.2.txt"
@@ -75,12 +64,11 @@ class CloudMaskTask(EOTask):
         :param output_proba_feature: The output feature containing cloud probabilities. By default this is not saved.
         :param data_indices: List of indices to use in case of custom input data. Defaults to 10 band indices used in
             s2cloudless '("B01", "B02", "B04", "B05", "B08", "B8A", "B09", "B10", "B11", "B12")'.
-        :param threshold: Cloud probability threshold for the classifier. Defaults to `0.4`.
-        :param average_over: Size of the pixel neighbourhood used in the averaging post-processing step.
-            A value of `0` skips this post-processing step. Default value mimics the default for
-            s2cloudless: `4`.
-        :param dilation_size: Size of the dilation post-processing step. A value of `0` or `None` skips this
-            post-processing step. Default value mimics the default for s2cloudless: `2`.
+        :param threshold: Cloud probability threshold for the classifier.
+        :param average_over: Size of the pixel neighbourhood used in the averaging post-processing step. Set to `None`
+            to skip this post-processing step.
+        :param dilation_size: Size of the dilation post-processing step. Set to `None` to skip this post-processing
+            step.
         """
         self.data_feature = self.parse_feature(data_feature)
         self.valid_data_feature = self.parse_feature(valid_data_feature)
--- a/eolearn/mask/cloud_mask.py
+++ b/eolearn/mask/cloud_mask.py
@@ -108,14 +108,6 @@ class CloudMaskTask(EOTask):
 
         self._classifier: ClassifierType | Booster | None = None
 
-    @staticmethod
-    def _parse_resolution_arg(resolution: None | float | tuple[float, float]) -> tuple[float, float] | None:
-        """Parses initialization resolution argument"""
-        if isinstance(resolution, (int, float)):
-            resolution = resolution, resolution
-
-        return resolution
-
     @property
     def classifier(self) -> ClassifierType | Booster:
         """An instance of a custom-provided cloud classifier. Loaded only the first time it is required."""
--- a/eolearn/mask/cloud_mask.py
+++ b/eolearn/mask/cloud_mask.py
@@ -18,7 +18,7 @@ import numpy as np
 from lightgbm import Booster
 from skimage.morphology import disk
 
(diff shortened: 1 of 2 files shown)


eolearn/mask/cloud_mask.py
```diff
@@ -41,6 +41,159 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:  # noqa: N803
...
+    is optional.
+
+    Example usage:
+    .. code-block:: python
+        task = CloudMaskTask(
+            data_feature = (FeatureType.DATA, 'BANDS'),
+            valid_data_feature = (FeatureType.MASK, 'IS_DATA'),
+            output_mask_feature=(FeatureType.MASK, 'CLM_S2C'),
+            output_proba_feature=(FeatureType.DATA, 'CLP_S2C')
+        )
```
zigaLuksic, line 54: i'd remove this

eolearn/mask/cloud_mask.py
```diff
@@ -41,6 +41,159 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:  # noqa: N803
...
+        average_over: int | None = 4,
+        dilation_size: int | None = 2,
+    ):
+        """
+        :param data_feature: A data feature which stores raw Sentinel-2 reflectance bands.
+        :param valid_data_feature: A mask feature which indicates whether data is valid.
+        :param output_mask_feature: The output feature containing cloud masks.
+        :param output_proba_feature: The output feature containing cloud probabilities. By default this is not saved.
+        :param data_indices: List of indices to use in case of custom input data. Defaults to 10 band indices used in
+            s2cloudless '("B01", "B02", "B04", "B05", "B08", "B8A", "B09", "B10", "B11", "B12")'.
```
zigaLuksic, line 77: this parameter is confusing to me. Is it indices of bands? Do they have to represent the mentioned bands? The calculation in line 89 also seems a bit convoluted

eolearn/mask/cloud_mask.py
```diff
@@ -41,6 +41,159 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:  # noqa: N803
...
+        :param threshold: Cloud probability threshold for the classifier. Defaults to `0.4`.
```
zigaLuksic, line 78: no reason to state defaults

eolearn/mask/cloud_mask.py
```diff
@@ -41,6 +41,159 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:  # noqa: N803
...
+        :param average_over: Size of the pixel neighbourhood used in the averaging post-processing step.
+            A value of `0` skips this post-processing step. Default value mimics the default for
+            s2cloudless: `4`.
+        :param dilation_size: Size of the dilation post-processing step. A value of `0` or `None` skips this
+            post-processing step. Default value mimics the default for s2cloudless: `2`.
```
zigaLuksic, line 83: I would argue that `None` should be the only 'skip' value

eolearn/mask/cloud_mask.py
```diff
@@ -41,6 +41,159 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:  # noqa: N803
...
+    @staticmethod
+    def _parse_resolution_arg(resolution: None | float | tuple[float, float]) -> tuple[float, float] | None:
+        """Parses initialization resolution argument"""
+        if isinstance(resolution, (int, float)):
+            resolution = resolution, resolution
+
+        return resolution
+
+    @property
+    def classifier(self) -> ClassifierType | Booster:
```
zigaLuksic, line 112: this cannot be a `ClassifierType` since it always loads a booster

eolearn/mask/cloud_mask.py
```diff
@@ -41,6 +41,159 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:  # noqa: N803
...
+        if self._classifier is None:
+            path = os.path.join(self.MODELS_FOLDER, self.CLASSIFIER_NAME)
+            self._classifier = Booster(model_file=path)
+
+        return self._classifier
+
+    @staticmethod
+    def _run_prediction(classifier: ClassifierType | Booster, features: np.ndarray) -> np.ndarray:
+        """Uses classifier object on given data"""
+        is_booster = isinstance(classifier, Booster)
```
zigaLuksic, line 123: similar to before, we're using a booster, but i guess this will be simplified with s2cloudless anyway right?

eolearn/mask/cloud_mask.py
```diff
@@ -41,6 +41,159 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:  # noqa: N803
...
+        predict_method = classifier.predict if is_booster else classifier.predict_proba
+        prediction: np.ndarray = execute_with_mp_lock(predict_method, features)
+
+        return prediction if is_booster else prediction[..., 1]
+
+    def _average(self, data: np.ndarray) -> np.ndarray:
+        return cv2.filter2D(data.astype(np.float64), -1, self.avg_kernel, borderType=cv2.BORDER_REFLECT)
+
+    def _dilate(self, data: np.ndarray) -> np.ndarray:
+        return (cv2.dilate(data.astype(np.uint8), self.dil_kernel) > 0).astype(np.uint8)
```
zigaLuksic, line 134: do you think we could move these definitions into `_average_all` and `_dilate_all` or would that be equally messy?

zigaLuksic (commented):
If some things are not that relevant due to the intended switch to s2cloudless just mark them as such

eolearn/mask/cloud_mask.py
mlubej: Realized I need to revert this anyway due to how it's handled in `s2cloudless`.

eolearn/mask/cloud_mask.py
mlubej: will be removed

eolearn/mask/cloud_mask.py
mlubej: will be removed

eolearn/mask/cloud_mask.py
mlubej: will be removed in the next MR

eolearn/mask/cloud_mask.py
mlubej: I think this is meant as "practically skips", because dilation of 0 does basically nothing. But I guess the process is still executed. Will just update the description that `None` should be provided to skip.

eolearn/mask/cloud_mask.py
mlubej: removed

eolearn/mask/cloud_mask.py
mlubej: removed

eolearn/mask/cloud_mask.py
mlubej: I reverted it now to what it was before. The reason was that it bothered me that the old indices were hard-coded, because the user can put in anything, but I was overthinking it.

eolearn/mask/cloud_mask.py
mlubej: I checked, the `> 0` condition is also in the code, but I think it's fine if we omit this from the description.

PR #723, opened 2023-08-25
Cloud mask task should use s2cloudless

mlubej:
Changes:
- new requirement for `eo-learn-mask` (s2cloudless)
- simplified `CloudMaskTask` after using `S2PixelCloudDetector` from `s2cloudless`

I expected the test data to change and the tests to fail, since the eo-learn version was using a different version of the s2cloudless model (or seemed to be). Doesn't seem to be any change in the tests, so this one is a freebie.

### DIFF
--- a/eolearn/mask/cloud_mask.py
+++ b/eolearn/mask/cloud_mask.py
@@ -16,6 +16,7 @@ from typing import Protocol, cast
 import cv2
 import numpy as np
 from lightgbm import Booster
+from s2cloudless import S2PixelCloudDetector
 from skimage.morphology import disk
 
 from sentinelhub import BBox, bbox_to_resolution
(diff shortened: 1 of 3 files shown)


eolearn/mask/cloud_mask.py
```diff
@@ -89,81 +87,29 @@ def __init__(
...
-        output_proba = []
-        _, height, width, n_bands = bands.shape
-
-        for img in bands:
-            features = img.reshape(height * width, n_bands)
-            proba = self._run_prediction(self.classifier, features)
-            output_proba.append(proba.reshape(height, width, 1))
-
-        return np.array(output_proba)
+        self.classifier = S2PixelCloudDetector(
```
zigaLuksic, line 90: just wondering if this should still be lazy loaded (pickling etc.)

mlubej: I believe it should be handled in `S2PixelCloudDetector`, because in the `init` phase (see [L60](https://github.com/sentinel-hub/sentinel2-cloud-detector/blob/master/s2cloudless/cloud_detector.py#L60)) the classifier is still not loaded, and is only loaded when the `classifier` method is called, which happens during the `execute` phase of the new task.

zigaLuksic: we really are smart sometimes :thinking:

zigaLuksic (approved):

PR #726, opened 2023-08-28
Update examples

zigaLuksic:
There is actually a lot less work than I feared there would be


examples/core/CoreOverview.ipynb
```diff
@@ -793,11 +843,18 @@
...
+       "model_id": "a0e29f064af545a79ee277e719c27391",
+       "version_major": 2,
+       "version_minor": 0
+      },
+      "text/plain": [
+       "  0%|          | 0/5 [00:00<?, ?it/s]"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
```
zigaLuksic, line 857:
```suggestion
     "name": "stderr",
     "output_type": "stream",
     "text": [
      "100%|██████████| 5/5 [00:00<00:00, 510.50it/s]\n"
     ]
```

examples/visualization/EOPatchVisualization.ipynb
```diff
@@ -371,7 +342,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.4"
+   "version": "3.8.10"
```
zigaLuksic, line 345:
```suggestion
   "version": "3.10.4"
```

zigaLuksic (commented):
just reverting some things, vscode can be a PITA for this sometimes

mlubej (approved):
Made minor edits, otherwise looked only at the changes, so I'm gonna trust you with the unchanged content, unless you think the examples would benefit from a second comb-through

zigaLuksic:
> Made minor edits, otherwise looked only at the changes, so I'm gonna trust you with the unchanged content, unless you think the examples would benefit from a second comb-through

Thanks. No need for second passes IMO, i ran the notebooks and adjusted fields that produced warnings. If we missed something the users will tell us.

PR #731, opened 2023-08-29
Compression changes to eo-learn

mlubej:
Changes:
- `compression_level` parameter removed
- compress by default for everything except bbox and timestamps
- test updates

### DIFF
--- a/tests/core/test_eodata_io.py
+++ b/tests/core/test_eodata_io.py
@@ -416,6 +416,31 @@ def test_nonexistent_location(fs_loader, use_zarr: bool):
 
 @mock_s3
 @pytest.mark.parametrize("fs_loader", FS_LOADERS)
+def test_cleanup_different_compression(fs_loader, eopatch):
+    folder = "foo-folder"
+    patch_folder = "patch-folder"
+    with fs_loader() as temp_fs:
+        save_task = SaveTask(folder, filesystem=temp_fs, overwrite_permission="OVERWRITE_FEATURES")
+        bbox_path = fs.path.join(folder, patch_folder, "bbox.geojson")
+        timestamps_path = fs.path.join(folder, patch_folder, "timestamps.json")
+
+        # need to manually save uncompressed feature
+        ftype, fname = (FeatureType.MASK_TIMELESS, "mask")
+        ftype_path = fs.path.join(folder, patch_folder, ftype.value)
+        temp_fs.makedirs(ftype_path)
+        feature_io = FeatureIONumpy(os.path.join(ftype_path, fname + ".npy"), filesystem=temp_fs)
+        feature_io.save(eopatch[(ftype, fname)], temp_fs, os.path.join(ftype_path, fname), compress_level=0)
+
+        # re-save compressed and check cleanup
+        save_task(eopatch, eopatch_folder=patch_folder)
+        assert temp_fs.exists(bbox_path)
+        assert temp_fs.exists(timestamps_path)
+        assert temp_fs.exists(os.path.join(ftype_path, fname) + ".npy.gz")
+        assert not temp_fs.exists(os.path.join(ftype_path, fname) + ".npy")
+
+
+@mock_s3
+@pytest.mark.parametrize("fs_loader", FS_LOADERS)
 @pytest.mark.parametrize("use_zarr", [True, False])
 @pytest.mark.parametrize("folder_name", ["/", "foo", "foo/bar"])
 @pytest.mark.usefixtures("_silence_warnings")
--- a/tests/core/test_eodata_io.py
+++ b/tests/core/test_eodata_io.py
@@ -7,7 +7,6 @@ This source code is licensed under the MIT license, see the LICENSE file in the 
 from __future__ import annotations
 
 import datetime
-import json
 import os
 import sys
 import warnings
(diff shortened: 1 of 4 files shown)


mlubej:
@zigaLuksic turns out that backwards compatibility already worked out of the box, since the featureIO objects retained the ability to handled loading of compressed/uncompressed objects

I added back the compression_level parameter as discussed, and added the EODeprecation warning on the save task and the save method. I removed the parameter from the docs, though.

Lastly, I added tests for checking the compression deprecation warning, and augmented a test for checking backwards compatibility loading and proper cleanup if different compressions are present (from before)

zigaLuksic:
> @zigaLuksic turns out that backwards compatibility already worked out of the box, since the featureIO objects retained the ability to handled loading of compressed/uncompressed objects

Slight correction, the `FeatureIO` objects retained the GZip capabilities so that we have backwards compatibility ;)
After a while, when this is removed, we'll be able simplify FeatureIO classes

mlubej:
Also, I wasn't sure if I should keep the compression capabilities in the `save_eopatch` function (from eolearn/core/eodata_io.py). This one should be internal, so people shouldn't be using it, and the tests might've looked a bit better, but dunno, someone still might use it.

zigaLuksic:
> Also, I wasn't sure if I should keep the compression capabilities in the `save_eopatch` function (from eolearn/core/eodata_io.py). This one should be internal, so people shouldn't be using it, and the tests might've looked a bit better, but dunno, someone still might use it.

I see no reason to keep it, since it would control only part of the compression behaviour (for numpy and geodata) which we dont really want to be changed (except for tests)

tests/core/test_eodata_io.py
```diff
@@ -408,65 +420,23 @@ def test_cleanup_different_compression(fs_loader, eopatch):
...
-            folder, filesystem=temp_fs, compress_level=9, overwrite_permission="OVERWRITE_FEATURES", use_zarr=True
-        )
-        save_noncompressed_task = SaveTask(
-            folder, filesystem=temp_fs, compress_level=0, overwrite_permission="OVERWRITE_FEATURES", use_zarr=True
-        )
-        bbox_path = fs.path.join(folder, patch_folder, "bbox.geojson")
-        compressed_bbox_path = bbox_path + ".gz"
-        mask_timeless_path = fs.path.join(folder, patch_folder, "mask_timeless", "mask.zarr")
-        wrong_compressed_mask_timeless_path = mask_timeless_path + ".gz"
+        # need to manually save uncompressed feature
```
zigaLuksic, line 427: i'd also add this for bbox and timestamps

tests/core/test_eodata_io.py
```diff
@@ -408,65 +420,23 @@ def test_cleanup_different_compression(fs_loader, eopatch):
...
+        ftype, fname = (FeatureType.MASK_TIMELESS, "mask")
+        ftype_path = fs.path.join(folder, patch_folder, ftype.value)
+        temp_fs.makedirs(ftype_path)
+        feature_io = FeatureIONumpy(os.path.join(ftype_path, fname + ".npy"), filesystem=temp_fs)
+        feature_io.save(eopatch[(ftype, fname)], temp_fs, os.path.join(ftype_path, fname), compress_level=0)
```
zigaLuksic, line 432: save methods are class methods, no need to instantiate the object `feature_io` (they are attached to FeatureIO due to organizational reasons)

tests/core/test_eodata_io.py
```diff
@@ -520,17 +490,16 @@ def test_lazy_loading_plus_overwrite_patch(fs_loader, folder_name, eopatch, use_
         (FeatureIOTimestamps, [datetime.datetime(2017, 1, 1, 10, 4, 7), datetime.datetime(2017, 1, 4, 10, 14, 5)]),
     ],
 )
-@pytest.mark.parametrize("compress_level", [0, 1])
```
zigaLuksic, line 523: why are you removing this parametrization here?

tests/core/test_eodata_io.py
mlubej: added

tests/core/test_eodata_io.py
mlubej: thanks, updated

mlubej: btw, should I add here something that checks if `EOPatch.load` works with uncompressed features for backwards compatibility? 

just something like 

```python
EOPatch.load(fs.path.join(folder, patch_folder), filesystem=temp_fs)
```

tests/core/test_eodata_io.py
mlubej: my bad, reverted

tests/core/test_eodata_io.py
```diff
@@ -328,6 +327,19 @@ def test_overwrite_failure(fs_loader, use_zarr: bool):
             )
 
 
+@mock_s3
+@pytest.mark.parametrize("fs_loader", FS_LOADERS)
+@pytest.mark.parametrize("compress_level", [0, 1])
+def test_compression_deprecation(eopatch, fs_loader, compress_level: int | None):
+    folder = "foo-folder"
+
```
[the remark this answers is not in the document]
mlubej: done

tests/core/test_eodata_io.py
zigaLuksic: yes please <3

mlubej: done

zigaLuksic (approved):

PR #734, opened 2023-08-30
Use cv2 in eolearn.mask

mlubej:
Note: 
- In cloud masking legacy task we used `cv2_disk` from `s2cloudless`, which is a bit different than what we used in morphology

### DIFF
--- a/eolearn/mask/cloud_mask.py
+++ b/eolearn/mask/cloud_mask.py
@@ -17,7 +17,7 @@ import cv2
 import numpy as np
 from lightgbm import Booster
 from s2cloudless import S2PixelCloudDetector
-from skimage.morphology import disk
+from s2cloudless.utils import cv2_disk
 
 from sentinelhub import BBox, bbox_to_resolution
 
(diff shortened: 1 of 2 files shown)


zigaLuksic:
If different ways of producing the kernel are used, then it needs to be documented whether it's the radius or the diameter.

I am starting to question why we're having these kernel operations in these tasks if we have a morphology task anyway :thinking: 
But i'm not sure about uniformity

mlubej:
> If different ways of producing the kernel are used, then it needs to be documented whether it's the radius or the diameter.

it's not the radius/diameter difference, it's just the different implementation of the pixelated circle. From the user interface POV it's the same, it's just that the circle shape is different, because we added a home-cooked cv2_disk to s2cloudless and field-delineation

zigaLuksic:
Nevermind, i just saw that this was for the legacy cloudmask task. I wouldn't even change that one. It's getting exported to the repo, no reason to adjust it's requirements IMO. I'll also let you merge these changes, i dont care really.

zigaLuksic (approved):

mlubej:
mah, you're right, I reverted the legacy part, just so it stays as legacy and everything is cleaner.

zigaLuksic (approved):

PR #735, opened 2023-08-30
Prepare final subpackage release

zigaLuksic:
This should prepare all `eo-learn-mask` style subpackages for the final release which will:
- install the same deps as before
- install a `deprecated_eolearn_mask` package, that is detected by `eo-learn` so that it can warn the user
- not add any code to anything


core/README.md
```diff
@@ -1,5 +1,5 @@
 # eolearn.core subpackage
 
-The main subpackage which implements basic building blocks (EOPatch, EOTask and EOWorkflow) and commonly used functionalities.
+The deprecated way to install `eolearn.core`. See [migration guide](https://github.com/sentinel-hub/eo-learn/issues/733) on how to correctly install the `eo-learn` package.
```
mlubej, line 3: Should we include the version here, like "how to correctly install the `eo-learn>=1.5` package.

mlubej (commented):
The initial comment is misleading since you only mention `eo-learn-mask`

core/README.md
zigaLuksic: Unsure, seems a bit redundant :thinking:

zigaLuksic: we are after all trying to get people to use 1.5.0 and onward

mlubej (approved):

PR #736, opened 2023-08-31
Get rid of BaseSnowMaskTask

zigaLuksic:


zigaLuksic:
If we were to keep both tasks the common method could be extracted as a function, getting rid of another 5 lines

a clear example that base classes are rarely the way to go

PR #741, opened 2023-09-05
Remove tasks that were moved to eo-learn-examples

zigaLuksic:


eolearn/features/bands_extraction.py
mlubej, line 1: rename file, perhaps to `ndi.py`

eolearn/io/geometry_io.py
mlubej, line 1: The only remaining task now is the `VectorImportTask`, so perhaps there is no more need to have `_BaseVectorImportTask`

mlubej, line 1: Additionally, I would remove the `SentinelHubSen2corTask` from `eolearn/io/sentinelhub_process.py`

eolearn/mask/extra/cloud_mask.py
mlubej, line 1: the `ClassifierType` can be removed

eolearn/ml_tools/__init__.py
mlubej, line 1: There are also some deprecated tools in the `ml_tools/utils.py`, should we remove them? Potentially also other deprecated things? or perhaps it's better to do this in a dedicated PR

mlubej:
I also noticed that the `example_data/REFERENCE_SCENES.npy` isn't used anywhere now, since it was moved to `eo-learn-examples/extra-tasks/radiometric_normalization/_test_data.npy`

eolearn/io/geometry_io.py
zigaLuksic: did we move it? if so, i will

mlubej: we didn't move it, that was my comment in https://github.com/sentinel-hub/eo-learn-examples/pull/34#issue-1880530818

> There is one more task which I have not ported, the SentinelHubSen2corTask. I think it pulls in too much code for a very simple thing, which is now easier achievable with the new code, so IMHO the SentinelHubSen2corTask has become redundant and I would just deprecated/remove it in eo-learn.

mlubej: feel free to do it separately, perhaps the review will be clearer

eolearn/mask/extra/cloud_mask.py
zigaLuksic: done

mlubej (approved):

PR #746, opened 2023-09-06
Make sure extra submodules are included when collecting list of tasks

zigaLuksic:
Need to test it out on read-the-docs as well (sometimes it takes a while before you can enable a branch)


zigaLuksic:
tested on RTD and works like a charm

PR #756, opened 2023-10-02
Add temporal filtration method

zigaLuksic:
I am open to adjust the input types.

If this gets merged I also think that we should add a corresponding task (that would then be suggested as a replacement for `FilterTimeSeriesTask`)

### DIFF
--- a/eolearn/core/eodata.py
+++ b/eolearn/core/eodata.py
@@ -726,6 +726,7 @@ class EOPatch:
             self, *eopatches, features=features, time_dependent_op=time_dependent_op, timeless_op=timeless_op
         )
 
+    @deprecated_function(EODeprecationWarning, "Please use the method `temporal_subset` instead.")
     def consolidate_timestamps(self, timestamps: list[dt.datetime]) -> set[dt.datetime]:
         """Removes all frames from the EOPatch with a date not found in the provided timestamps list.
 
@@ -765,7 +766,7 @@ class EOPatch:
         new_patch = EOPatch(bbox=self.bbox, timestamps=new_timestamps)
 
         for ftype, fname in self.get_features():
-            if ftype.is_timeless():
+            if ftype.is_timeless() or ftype.is_meta():
                 new_patch[ftype, fname] = self[ftype, fname]
             elif ftype.is_vector():
                 gdf: gpd.GeoDataFrame = self[ftype, fname]
--- a/eolearn/core/eodata.py
+++ b/eolearn/core/eodata.py
@@ -753,9 +753,12 @@ class EOPatch:
     def temporal_subset(
         self, timestamps: Iterable[dt.datetime] | Iterable[int] | Callable[[dt.datetime], bool]
     ) -> EOPatch:
-        """TODO
+        """Returns an EOPatch that only contains data for the temporal subset defined by `timestamps`.
 
-        :param timestamps: TODO
+        For array based data appropriate temporal slices are extracted. For vector data a filtration is performed.
+
+        :param timestamps: Parameter that defines the temporal subset. Can be a collection of timestamps, a
+            collection of timestamp indices, or a callable that returns whether a timestamp should be kept.
         """
         timestamp_indices = self._parse_temporal_subset_input(timestamps)
         new_timestamps = [ts for i, ts in enumerate(self.get_timestamps()) if i in set(timestamp_indices)]
@@ -775,6 +778,7 @@ class EOPatch:
     def _parse_temporal_subset_input(
         self, timestamps: Iterable[dt.datetime] | Iterable[int] | Callable[[dt.datetime], bool]
     ) -> list[int]:
+        """Parses input into a list of timestamp indices. Also adds implicit support for strings via `parse_time`."""
         if callable(timestamps):
             return [i for i, ts in enumerate(self.get_timestamps()) if timestamps(ts)]
         ts_or_idx = list(timestamps)
--- a/eolearn/core/eodata.py
+++ b/eolearn/core/eodata.py
@@ -750,6 +750,39 @@ class EOPatch:
 
         return remove_from_patch
 
+    def temporal_subset(
+        self, timestamps: Iterable[dt.datetime] | Iterable[int] | Callable[[dt.datetime], bool]
+    ) -> EOPatch:
+        """TODO
+
+        :param timestamps: TODO
+        """
+        timestamp_indices = self._parse_temporal_subset_input(timestamps)
+        new_timestamps = [ts for i, ts in enumerate(self.get_timestamps()) if i in set(timestamp_indices)]
+        new_patch = EOPatch(bbox=self.bbox, timestamps=new_timestamps)
+
+        for ftype, fname in self.get_features():
+            if ftype.is_timeless():
+                new_patch[ftype, fname] = self[ftype, fname]
+            elif ftype.is_vector():
+                gdf: gpd.GeoDataFrame = self[ftype, fname]
+                new_patch[ftype, fname] = gdf[gdf[TIMESTAMP_COLUMN].isin(new_timestamps)]
+            else:
+                new_patch[ftype, fname] = self[ftype, fname][timestamp_indices, ...]
+
+        return new_patch
+
+    def _parse_temporal_subset_input(
+        self, timestamps: Iterable[dt.datetime] | Iterable[int] | Callable[[dt.datetime], bool]
+    ) -> list[int]:
+        if callable(timestamps):
+            return [i for i, ts in enumerate(self.get_timestamps()) if timestamps(ts)]
+        ts_or_idx = list(timestamps)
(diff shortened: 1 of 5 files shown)


eolearn/core/eodata.py
```diff
@@ -750,6 +751,43 @@ def consolidate_timestamps(self, timestamps: list[dt.datetime]) -> set[dt.dateti
 
         return remove_from_patch
 
+    def temporal_subset(
+        self, timestamps: Iterable[dt.datetime] | Iterable[int] | Callable[[dt.datetime], bool]
+    ) -> EOPatch:
+        """Returns an EOPatch that only contains data for the temporal subset defined by `timestamps`.
+
+        For array based data appropriate temporal slices are extracted. For vector data a filtration is performed.
```
mlubej, line 759:
```suggestion
        For array-based data appropriate temporal slices are extracted. For vector data a filtration is performed.
```

eolearn/core/eodata.py
```diff
@@ -750,6 +751,43 @@ def consolidate_timestamps(self, timestamps: list[dt.datetime]) -> set[dt.dateti
 
         return remove_from_patch
 
+    def temporal_subset(
+        self, timestamps: Iterable[dt.datetime] | Iterable[int] | Callable[[dt.datetime], bool]
+    ) -> EOPatch:
+        """Returns an EOPatch that only contains data for the temporal subset defined by `timestamps`.
```
mlubej, line 757:
```suggestion
        """Returns an EOPatch that only contains data for the temporal subset corresponding to `timestamps`.
```

eolearn/core/eodata.py
```diff
@@ -750,6 +751,43 @@ def consolidate_timestamps(self, timestamps: list[dt.datetime]) -> set[dt.dateti
...
+                new_patch[ftype, fname] = gdf[gdf[TIMESTAMP_COLUMN].isin(new_timestamps)]
+            else:
+                new_patch[ftype, fname] = self[ftype, fname][timestamp_indices, ...]
+
+        return new_patch
+
+    def _parse_temporal_subset_input(
+        self, timestamps: Iterable[dt.datetime] | Iterable[int] | Callable[[dt.datetime], bool]
+    ) -> list[int]:
+        """Parses input into a list of timestamp indices. Also adds implicit support for strings via `parse_time`."""
```
mlubej, line 782: why not include the string option in the input types, then?

eolearn/core/eodata.py
```diff
@@ -750,6 +751,43 @@ def consolidate_timestamps(self, timestamps: list[dt.datetime]) -> set[dt.dateti
...
+        if callable(timestamps):
+            return [i for i, ts in enumerate(self.get_timestamps()) if timestamps(ts)]
+        ts_or_idx = list(timestamps)
+        if all(isinstance(ts, int) for ts in ts_or_idx):
+            return ts_or_idx  # type: ignore[return-value]
+        parsed_timestamps = {parse_time(ts, force_datetime=True) for ts in ts_or_idx}  # type: ignore[call-overload]
```
mlubej, line 788: should there be a check that timestamps and indices shouldn't be mixed?

eolearn/core/eodata.py
```diff
@@ -750,6 +751,43 @@ def consolidate_timestamps(self, timestamps: list[dt.datetime]) -> set[dt.dateti
...
+    ) -> EOPatch:
+        """Returns an EOPatch that only contains data for the temporal subset defined by `timestamps`.
+
+        For array based data appropriate temporal slices are extracted. For vector data a filtration is performed.
+
+        :param timestamps: Parameter that defines the temporal subset. Can be a collection of timestamps, a
+            collection of timestamp indices, or a callable that returns whether a timestamp should be kept.
+        """
+        timestamp_indices = self._parse_temporal_subset_input(timestamps)
+        new_timestamps = [ts for i, ts in enumerate(self.get_timestamps()) if i in set(timestamp_indices)]
```
mlubej, line 765: is the set here necessary? I don't expect duplicates, but if it makes inference faster, then it's fine

eolearn/core/eodata.py
```diff
@@ -750,6 +751,43 @@ def consolidate_timestamps(self, timestamps: list[dt.datetime]) -> set[dt.dateti
...
+        new_patch = EOPatch(bbox=self.bbox, timestamps=new_timestamps)
+
+        for ftype, fname in self.get_features():
+            if ftype.is_timeless() or ftype.is_meta():
+                new_patch[ftype, fname] = self[ftype, fname]
+            elif ftype.is_vector():
+                gdf: gpd.GeoDataFrame = self[ftype, fname]
+                new_patch[ftype, fname] = gdf[gdf[TIMESTAMP_COLUMN].isin(new_timestamps)]
+            else:
+                new_patch[ftype, fname] = self[ftype, fname][timestamp_indices, ...]
```
mlubej, line 775: the ellipsis are not necessary here, right?

tests/core/test_eodata_merge.py
```diff
@@ -206,8 +206,7 @@ def test_lazy_loading(test_eopatch_path):
 
 def test_temporally_independent_merge(test_eopatch_path):
     full_patch = EOPatch.load(test_eopatch_path)
-    part1, part2 = full_patch.copy(deep=True), full_patch.copy(deep=True)
-    part1.consolidate_timestamps(full_patch.get_timestamps()[:10])
-    part2.consolidate_timestamps(full_patch.get_timestamps()[10:])
+    part1 = full_patch.copy(deep=True).temporal_subset(range(10))
+    part2 = full_patch.copy(deep=True).temporal_subset(range(10, len(full_patch.timestamps)))
```
mlubej, line 210: usually you warn us to use the `get_timestamps` method, even if it's obvious that the timestamps will be present. Can you elaborate on your decision process?

tests/core/test_eodata.py
```diff
@@ -444,20 +445,48 @@ def test_timestamp_consolidation() -> None:
...
+        timestamps=[
+            dt.datetime(2017, 1, 5),
+            dt.datetime(2017, 4, 8),
+            dt.datetime(2017, 9, 17),
+            dt.datetime(2018, 1, 5),
+            dt.datetime(2018, 12, 1),
+        ],
+    )
+    vector_data = GeoDataFrame(
+        {"TIMESTAMP": eop.get_timestamps()}, geometry=[eop.bbox.geometry.buffer(i) for i in range(5)], crs=32633
```
mlubej, line 473: probably better if you import the `TIMESTAMP_COLUMN` variable

eolearn/core/eodata.py
zigaLuksic: i allow such things for research purposes. I want eo-grow to smack you if you try it.

eolearn/core/eodata.py
zigaLuksic: if it's all ints it's assumed to be indices, otherwise we assume it's timestamps. If you mix it's not our problem.
In general i would only check things that might slip the type-checker. Because this is a method that I imagine might be used by researchers in notebooks i also included the parser in here.

eolearn/core/eodata.py
zigaLuksic: i wanted to see what the impact is. For a whole year worth of indices, if you use a set instead of a list it's about one magnitude faster (still under 1ms in both cases). However here you must also BUILD the set. Turns out constructing the set is slower than the suboptimal list lookup, which sounded weird to me.

But then i remembered that this is Python and Python is unpredictable in best of times.
image

The set is recalculated every time, and it should be. But one quickly forgets that with list comprehensions.

tests/core/test_eodata_merge.py
zigaLuksic: it's tests and i'm lazy :upside_down_face:

eolearn/core/eodata.py
zigaLuksic: i removed the set transformation since even for a full year the whole thing is under ms when using lists

eolearn/core/eodata.py
zigaLuksic: yeah, they are :D

mlubej: I don't get it, index slicing works fine without the ellipsis, since it's applied only on the first axis. Or am I missing something?

mlubej: Ah, I probably misunderstood, I see you removed tem

zigaLuksic: sorry, in my mind it was `yeah, they are unnecessary`

eolearn/core/eodata.py
mlubej: just a random question, if you know by heart. Would the set also be recalculated every time if this was a standard for loop?

zigaLuksic: yes, because one can do some stupid things like
```
for x in blabla:
    if x in fetch_constantly_updating_thing():
        ...
```
So it was a problem of my brain assuming stuff

mlubej (approved):

PR #759, opened 2023-10-13
Small refactor of parallelization utils

zigaLuksic:
Did some snooping, i think this module could be re-done to be less scattered and complex.

For now i only did some simplifications.

### DIFF
--- a/eolearn/core/utils/parallelize.py
+++ b/eolearn/core/utils/parallelize.py
@@ -49,9 +49,7 @@ def _decide_processing_type(workers: int | None, multiprocess: bool) -> _Process
     """
     if workers == 1:
         return _ProcessingType.SINGLE_PROCESS
-    if multiprocess:
-        return _ProcessingType.MULTIPROCESSING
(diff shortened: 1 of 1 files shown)


eolearn/core/utils/parallelize.py
```diff
@@ -74,10 +72,7 @@ def parallelize(
     :return: A list of function results.
     """
     if not params:
-        raise ValueError(
-            "At least 1 list of parameters should be given. Otherwise it is not clear how many times the"
-            "function has to be executed."
-        )
+        return []
```
mlubej, line 75: Just to be clear, you removed the warning and just return the empty list, meaning that nothing is done in the end? Or is the error raised at some later stage?

I guess the warning really isn't necessary, if you don't provide params, there's nothing to be done, period.

zigaLuksic: yeah, i dont know why there was an error there in the first place. It is clear what the result should be

mlubej (approved):

PR #760, opened 2023-10-16
Add option to treat temporal warnings as errors in eoexecutions

zigaLuksic:
Do you think this might be useful?

### DIFF
--- a/eolearn/core/eoexecution.py
+++ b/eolearn/core/eoexecution.py
@@ -55,7 +55,7 @@ class _ProcessingData:
     filter_logs_by_thread: bool
     logs_filter: Filter | None
     logs_handler_factory: _HandlerFactoryType
-    temporal_dim_warning_is_error: bool
+    raise_on_temporal_mismatch: bool
 
 
 @dataclass(frozen=True)
@@ -87,7 +87,7 @@ class EOExecutor:
         filesystem: FS | None = None,
         logs_filter: Filter | None = None,
         logs_handler_factory: _HandlerFactoryType = FileHandler,
-        temporal_dimension_warning_is_error: bool = False,
+        raise_on_temporal_mismatch: bool = False,
     ):
         """
         :param workflow: A prepared instance of EOWorkflow class
@@ -110,7 +110,7 @@ class EOExecutor:
               object.
 
             The 2nd option is chosen only if `filesystem` parameter exists in the signature.
-        :param temporal_dimension_warning_is_error: Whether to treat `TemporalDimensionWarning` as an exception.
+        :param raise_on_temporal_mismatch: Whether to treat `TemporalDimensionWarning` as an exception.
         """
         self.workflow = workflow
         self.execution_kwargs = self._parse_and_validate_execution_kwargs(execution_kwargs)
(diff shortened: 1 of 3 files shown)


eolearn/core/eoexecution.py
```diff
@@ -55,6 +55,7 @@ class _ProcessingData:
     filter_logs_by_thread: bool
     logs_filter: Filter | None
     logs_handler_factory: _HandlerFactoryType
+    temporal_dim_warning_is_error: bool
```
mlubej, line 58: I'm having doubts about the name, but not sure about alternatives. Maybe:
- `raise_on_temporal_mismatch`
- `temporal_mismatch_is_error`

or something along those lines

eolearn/core/eoexecution.py
```diff
@@ -86,6 +87,7 @@ def __init__(
         filesystem: FS | None = None,
         logs_filter: Filter | None = None,
         logs_handler_factory: _HandlerFactoryType = FileHandler,
+        temporal_dimension_warning_is_error: bool = False,
```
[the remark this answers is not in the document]
zigaLuksic: yes, the motivation for this MR is to add a flag to eo-grow for this.

eolearn/core/eoexecution.py
zigaLuksic: yeah,  no matter what you try you end up with something long :see_no_evil: I was thinking of shortening it to `raise_on_tdim_missmatch` but `tdim` is kinda  non-standard

mlubej: Yes, generally I meant more for the parameter in the function than the private one. Perhaps I'm more inclined to use the one with `raise` because I remember seeing it in libraries a few times as a parameter as well.

but in the end it's a specific thing, which requires a specific name, so if you don't like the suggestions, it's fine and feel free to leave it as it is.

zigaLuksic:
> BTW, perhaps we should be adding to the changelog as we go along, not before each release.

I was thinking about it, but enforcing it seems like a drag compared to 5mins of staring at the history to come up with a condensed history.

I am willing to go with `raise_on_temporal_dimension_mismatch` though it is long

mlubej:
> `raise_on_temporal_dimension_mismatch`

Do you feel that the `dimension` keyword is necessary? IMO it's clear enough to me, so I would drop it, but if you don't want to drop it, then I'm OK with the current name as well.

zigaLuksic:
> > `raise_on_temporal_dimension_mismatch`
> 
> Do you feel that the `dimension` keyword is necessary? IMO it's clear enough to me, so I would drop it, but if you don't want to drop it, then I'm OK with the current name as well.

I am inclined to keep it because a temporal missmatch is also if the timestamps don't match, but we only check dimensions. But then again the docs do clarify it further :thinking:  so maybe we can drop it

tests/core/test_eoexecutor.py
```diff
@@ -251,3 +265,21 @@ def test_without_lock(num_workers):
...
+            CreateEOPatchTask(bbox=BBox((0, 0, 1, 1), CRS.POP_WEB)),
+            InitializeFeatureTask([FeatureType.DATA, "data"], (2, 5, 5, 1)),
+        )
+    )
+
+    executor = EOExecutor(workflow, [{}, {}])
+    for result in executor.run(workers=2, multiprocess=multiprocess):
+        assert result.error_node_uid is None
+
+    executor = EOExecutor(workflow, [{}, {}], temporal_dimension_warning_is_error=True)
```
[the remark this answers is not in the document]
zigaLuksic: no it's not :P

mlubej (approved):

PR #762, opened 2023-10-17
Adjust EOPatch repr method

zigaLuksic:
Turns out dictionaries were never covered. Also any object other than `tuple` or `list` would just be printed out in full. And if the element inside a list was huge, why not, print that fully as well...

So what i did is make sure that:
- Every element with a representation that is too large gets shortened to `str(type(x))`
- Dictionaries follow a similar set of rules as tuples and lists (print first element and show length)
- When doing the above, the first element is recursively formatted with the same function
- I switched from `[x, ...], length=100` to `[x, ...]<length=100>` otherwise recursive nesting was unreadable since you got `[[x, ...], length=100], length=100`, but `[[x, ...]<length=100>, ...]<length=100>` is a bit better.

Example of how it works now:
image

### DIFF
--- a/eolearn/core/eodata.py
+++ b/eolearn/core/eodata.py
@@ -467,16 +467,14 @@ class EOPatch:
             crs = CRS(value.crs).ogc_string() if value.crs else value.crs
             return f"{EOPatch._repr_value_class(value)}(columns={list(value)}, length={len(value)}, crs={crs})"
 
-        if isinstance(value, (list, tuple, dict)) and value:
-            repr_str = str(value)
-            if len(repr_str) <= MAX_DATA_REPR_LEN:
-                return repr_str
+        repr_str = str(value)
+        if len(repr_str) <= MAX_DATA_REPR_LEN:
+            return repr_str
 
+        if isinstance(value, (list, tuple)) and value:
             l_bracket, r_bracket = ("[", "]") if isinstance(value, list) else ("(", ")")
-            if isinstance(value, (list, tuple)) and len(value) > 2:
-                repr_str = f"{l_bracket}{value[0]!r}, ..., {value[-1]!r}{r_bracket}"
 
-            if len(repr_str) > MAX_DATA_REPR_LEN and isinstance(value, (list, tuple)) and len(value) > 1:
+            if len(repr_str) > MAX_DATA_REPR_LEN and len(value) > 1:
                 repr_str = f"{l_bracket}{value[0]!r}, ...{r_bracket}"
 
             if len(repr_str) > MAX_DATA_REPR_LEN:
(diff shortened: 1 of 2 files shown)


zigaLuksic:
Previously it also had another option where it tried to print out the first and last element
image

This is now no longer the case. I guess it made sense for timestamps (so you could guesstimate the time range if you hoped the timestamps are ordered)
image

I think one should just open the timestamps if they want more info, but i'm interested in the opinion of researchers.

veseln:
> I think one should just open the timestamps if they want more info, but i'm interested in the opinion of researchers.

 Yeah, I agree. I don't think I ever need that feature ( atleast not knowingly), so the proposed change is okay. 

Other changes look okay to me.

eolearn/core/eodata.py
```diff
@@ -455,36 +455,36 @@ def __repr__(self) -> str:
...
+            lb, rb = ("[", "]") if isinstance(value, list) else ("(", ")") if isinstance(value, tuple) else ("{", "}")
 
-            l_bracket, r_bracket = ("[", "]") if isinstance(value, list) else ("(", ")")
-            if isinstance(value, (list, tuple)) and len(value) > 2:
-                repr_str = f"{l_bracket}{value[0]!r}, ..., {value[-1]!r}{r_bracket}"
+            if isinstance(value, dict):  # generate representation of first element or key, value pair
+                some_key = next(iter(value))
+                repr_of_el = f"{EOPatch._repr_value(some_key)}: {EOPatch._repr_value(value[some_key])}"
+            else:
+                repr_of_el = EOPatch._repr_value(value[0])
```
veseln, line 477: can it happen that value is empty?

mlubej:
> I think one should just open the timestamps if they want more info, but i'm interested in the opinion of researchers.

I guess it isn't so important that this would need a specific handling. and once the eop is loaded, the info is not far away

eolearn/core/eodata.py
```diff
@@ -455,36 +455,36 @@ def __repr__(self) -> str:
...
         if isinstance(value, (list, tuple, dict)) and value:
-            repr_str = str(value)
-            if len(repr_str) <= MAX_DATA_REPR_LEN:
-                return repr_str
+            lb, rb = ("[", "]") if isinstance(value, list) else ("(", ")") if isinstance(value, tuple) else ("{", "}")
 
-            l_bracket, r_bracket = ("[", "]") if isinstance(value, list) else ("(", ")")
-            if isinstance(value, (list, tuple)) and len(value) > 2:
-                repr_str = f"{l_bracket}{value[0]!r}, ..., {value[-1]!r}{r_bracket}"
+            if isinstance(value, dict):  # generate representation of first element or key, value pair
```
mlubej, line 473: just to make it less confusing

```suggestion
            if isinstance(value, dict):  # generate representation of first element or (key, value) pair
```

eolearn/core/eodata.py
zigaLuksic: if the value is empty then the lenght of the string will (should) be under 100 characters

zigaLuksic: ah, also, we have `if isinstance(value, (dict, list, tuple)) and value` so if it were empty the second part would fail (to ensure you sleep even better)

PR #764, opened 2023-10-23
Refactor eoexecutor

zigaLuksic:
Wanted to figure out where to push in the ray kwargs and decided to clean up a bit first.

### DIFF
--- a/eolearn/core/eoexecution.py
+++ b/eolearn/core/eoexecution.py
@@ -15,22 +15,25 @@ from __future__ import annotations
 import concurrent.futures
 import datetime as dt
 import inspect
+import itertools as it
 import logging
 import threading
 import warnings
 from dataclasses import dataclass
 from logging import FileHandler, Filter, Handler, Logger
-from typing import Any, Callable, Protocol, Sequence, Union
+from typing import Any, Callable, Iterable, Protocol, Union
 
 import fs
 from fs.base import FS
 
+from sentinelhub.exceptions import deprecated_function
+
 from .eonode import EONode
 from .eoworkflow import EOWorkflow, WorkflowResults
-from .exceptions import EORuntimeWarning, TemporalDimensionWarning
+from .exceptions import EODeprecationWarning, EORuntimeWarning, TemporalDimensionWarning
 from .utils.fs import get_base_filesystem_and_path, get_full_path, pickle_fs, unpickle_fs
 from .utils.logging import LogFileFilter
-from .utils.parallelize import _decide_processing_type, _ProcessingType, parallelize
+from .utils.parallelize import parallelize
 
 
 class _HandlerWithFsFactoryType(Protocol):
@@ -79,9 +82,9 @@ class EOExecutor:
     def __init__(
         self,
         workflow: EOWorkflow,
-        execution_kwargs: Sequence[dict[EONode, dict[str, object]]],
+        execution_kwargs: Iterable[dict[EONode, dict[str, object]]],
         *,
-        execution_names: list[str] | None = None,
+        execution_names: Iterable[str] | None = None,
         save_logs: bool = False,
         logs_folder: str = ".",
         filesystem: FS | None = None,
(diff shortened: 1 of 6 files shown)


eolearn/core/eoexecution.py
```diff
@@ -181,11 +180,7 @@ def run(self, workers: int | None = 1, multiprocess: bool = True, **tqdm_kwargs:
         if self.save_logs:
             self.filesystem.makedirs(self.report_folder, recreate=True)
 
-        log_paths: Sequence[str | None]
-        if self.save_logs:
-            log_paths = self.get_log_paths(full_path=False)
-        else:
-            log_paths = [None] * len(self.execution_kwargs)
+        log_paths = self.get_log_paths(full_path=False) if self.save_logs else it.repeat(None)
```
mlubej, line 183: would it make sense to provide the `times` parameter to `repeat`, since we know it?

eolearn/visualization/eoexecutor.py
```diff
@@ -69,8 +69,13 @@ def make_report(self, include_logs: bool = True) -> None:
             template = self._get_template()
 
             execution_log_filenames = [fs.path.basename(log_path) for log_path in self.eoexecutor.get_log_paths()]
-            if self.eoexecutor.save_logs:
-                execution_logs = self.eoexecutor.read_logs() if include_logs else None
+            if not include_logs:
+                execution_logs = None
+            elif self.eoexecutor.save_logs:
+                execution_logs = []
+                for log_path in self.eoexecutor.get_log_paths():
```
mlubej, line 76: There is now a for loop in place of the `ThreadPoolExecutor`. What are the implications of this w.r.t. the eo-grow pipeline executions?

tests/core/test_eoexecutor.py
```diff
@@ -143,7 +142,11 @@ def test_read_logs(test_args, execution_names, workflow, execution_kwargs, logs_
         )
         executor.run(workers=workers, multiprocess=multiprocess)
 
-        execution_logs = executor.read_logs()
+        execution_logs = []
+        for log_path in executor.get_log_paths():
+            with open(log_path) as f:
+                execution_logs.append(f.read())
+
```
mlubej, line 149: Seems like the `read_logs` pops up quite often, are you sure you want to deprecate it?

eolearn/core/eoexecution.py
```diff
@@ -128,27 +131,23 @@ def __init__(
...
+    def _parse_execution_names(execution_names: Iterable[str] | None, execution_kwargs: list) -> list[str]:
         """Parses a list of execution names."""
         if execution_names is None:
             return [str(num) for num in range(1, len(execution_kwargs) + 1)]
 
-        if not isinstance(execution_names, (list, tuple)) or len(execution_names) != len(execution_kwargs):
-            raise ValueError(
-                "Parameter 'execution_names' has to be a list of the same size as the list of execution arguments."
-            )
+        execution_names = list(execution_names)
```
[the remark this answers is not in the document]
zigaLuksic: here we transform to a list so that generator objects can be passed in here. we also want to have a very precise type after parsing.

eolearn/core/eoexecution.py
```diff
@@ -128,27 +131,23 @@ def __init__(
 
     @staticmethod
     def _parse_and_validate_execution_kwargs(
-        execution_kwargs: Sequence[dict[EONode, dict[str, object]]]
+        execution_kwargs: Iterable[dict[EONode, dict[str, object]]]
     ) -> list[dict[EONode, dict[str, object]]]:
         """Parses and validates execution arguments provided by user and raises an error if something is wrong."""
-        if not isinstance(execution_kwargs, (list, tuple)):
```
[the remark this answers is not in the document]
zigaLuksic: i think types are the correct approach here

eolearn/core/eoexecution.py
zigaLuksic: any reason for it?

eolearn/visualization/eoexecutor.py
zigaLuksic: if there are a lot of them you shouldn't include logs anyway, or you get huge reports.
i can re-add the threadpool to try and improve efficiency a bit

tests/core/test_eoexecutor.py
zigaLuksic: i see no reason why the executor would read logs for you. I even think that the current approach where log paths are slapped onto it is wrong. In the future we should make it stateless, and all of these things get pinned to a 'result' object.

the only place in the code where it pops up is generating the report, other uses are just in tests because it is convenient

eolearn/core/eoexecution.py
mlubej: infinities are scary! 

my thinking was that it's better to fail if potentially a larger number is requested, rather than passing silently

🤷  

but not so important, I guess

eolearn/visualization/eoexecutor.py
mlubej: It wasn't a complaint, just an observation, since I'm not so familiar with the processes so deep down the hole

zigaLuksic: it is a legitimate concern

zigaLuksic: readded treaded loading

eolearn/core/eoexecution.py
zigaLuksic: tried adding it to make you feel safer, but the line gets too long and more difficult to read :(

mlubej: I will accidentally click resolve and look away while you merge

PR #766, opened 2023-10-26
Ray executor can request resources for workers

zigaLuksic:
Kinda hard to test :/ ideas welcome

### DIFF
--- a/eolearn/core/eoexecution.py
+++ b/eolearn/core/eoexecution.py
@@ -210,13 +210,12 @@ class EOExecutor:
 
         return full_execution_results
 
-    @classmethod
     def _run_execution(
-        cls, processing_args: list[_ProcessingData], run_params: _ExecutionRunParams
+        self, processing_args: list[_ProcessingData], run_params: _ExecutionRunParams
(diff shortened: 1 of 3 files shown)


mlubej (approved):
basically what was already discussed in the pair session

zigaLuksic:
> I'm also not sure about testing this without actually properly running things on the cluster.

i guess we could try and request too much of a resource and then try and catch the warning?

mlubej:
> i guess we could try and request too much of a resource and then try and catch the warning?

Seems like if you ask for a non-existing resource, or for a larger resource chunk than it exists, the process just stalls

zigaLuksic:
> > i guess we could try and request too much of a resource and then try and catch the warning?
> 
> Seems like if you ask for a non-existing resource, or for a larger resource chunk than it exists, the process just stalls

yeah, but on the cluster it returns a warning. But i cant remember if it does that inside the python runtime as well, might just be a log message.

mlubej:
No further ideas about the tests...

mlubej (approved):

PR #768, opened 2023-11-03
Tiff tasks parameter `folder` renamed to `path`

zigaLuksic:
It's often not a folder, so it's very confusing...

Ellipsis as a default seemed like the best way to signal that this might still be required (using `None` would be odd).

- `TiffTask(feature, my_path)` works as before
- `TiffTask(feature)` will raise a `TypeError` that the parameter `path` is missing (similar to what would happen before)
- `TiffTask(feature, path=my_path)` works as intended
- `TiffTask(feature, folder=my_path)` also works as previously intended, but also warns the user

### DIFF
--- a/eolearn/io/raster_io.py
+++ b/eolearn/io/raster_io.py
@@ -32,7 +32,7 @@ from sentinelhub import CRS, BBox, SHConfig, parse_time_interval
 
 from eolearn.core import EOPatch
 from eolearn.core.core_tasks import IOTask
-from eolearn.core.exceptions import EORuntimeWarning
+from eolearn.core.exceptions import EODeprecationWarning, EORuntimeWarning
 from eolearn.core.types import Feature
 from eolearn.core.utils.fs import get_base_filesystem_and_path, get_full_path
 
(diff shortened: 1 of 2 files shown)


eolearn/io/raster_io.py
```diff
@@ -42,39 +42,43 @@ class BaseRasterIoTask(IOTask, metaclass=ABCMeta):
...
-        ftype, fname = self.parse_feature(feature)
-        if fname is None:
-            raise ValueError(f"Feature {feature} is not eligible for this task.")
-        self.feature = ftype, fname
+        if folder is not None:
+            warnings.warn("The parameter `folder` has been renamed to `path`.", EODeprecationWarning, stacklevel=3)
+            path = folder
+
+        if path is ...:  # type: ignore[comparison-overlap]
+            raise TypeError(f"{type(self).__name__} is missing a required positional argument 'path'.")
```
mlubej, line 71: If `path` has to be set, why not just leave it uninitialized and let the function complain on its own?

eolearn/io/raster_io.py
```diff
@@ -406,18 +409,19 @@ class ImportFromTiffTask(BaseRasterIoTask):
...
         filesystem: FS | None = None,
         image_dtype: np.dtype | type | None = None,
         no_data_value: float | None = None,
         config: SHConfig | None = None,
+        folder: str | None = None,
     ):
         """
         :param feature: EOPatch feature into which data will be imported
-        :param folder: A directory containing image files or a path of an image file
+        :param path: A directory containing image files or a path of an image file
```
mlubej, line 424: The same comment about the `filesystem` parameter should be given here, no?

```suggestion
        :param path: Path pointing to an image file or to a directory of image files to be imported. If `filesystem` is not defined, absolute paths should be provided.
```

eolearn/io/raster_io.py
```diff
@@ -42,39 +42,43 @@ class BaseRasterIoTask(IOTask, metaclass=ABCMeta):
...
         config: SHConfig | None = None,
+        folder: str | None = None,
     ):
         """
         :param feature: Feature which will be exported or imported
-        :param folder: A path to a main folder containing all image, potentially in its subfolders. If `filesystem`
-            parameter is defined, then `folder` should be a path relative to filesystem object. Otherwise, it should be
-            an absolute path.
+        :param path: A path to the image(s). If `filesystem` is defined, then `folder` should be given relative to the
+            filesystem object. Otherwise, it should be an absolute path.
```
mlubej, line 57: This is a bit confusing, since you mention `folder`, which seems like a reference to the parameter which has been deprecated.

```suggestion
        :param path: Path pointing to an image file or to a directory of image files. If `filesystem` is not defined, absolute paths should be provided.
```

eolearn/io/raster_io.py
```diff
@@ -156,12 +159,12 @@ def __init__(
...
         config: SHConfig | None = None,
+        folder: str | None = None,
     ):
         """
         :param feature: A feature to be exported.
-        :param folder: A path to a main folder containing all image, potentially in its subfolders. If `filesystem`
-            parameter is defined, then `folder` should be a path relative to filesystem object. Otherwise, it should be
-            an absolute path.
+        :param path: A path to the image(s). If `filesystem` is defined, then `folder` should be given relative to the
+            filesystem object. Otherwise, it should be an absolute path.
```
mlubej, line 167: Same here

```suggestion
        :param path: Path pointing to an image file or to a directory of image files to be exported. If `filesystem` is not defined, absolute paths should be provided.
```

eolearn/io/raster_io.py
zigaLuksic: because then `TiffTask(feature, folder=my_path)` would fail.

zigaLuksic:
thanks for the new docstrings, much clearer :bow:

eolearn/io/raster_io.py
mlubej: What if both are allowed to be `None`, since we allow the user to provide the path also in the `execute` part? This way we can just check if the `folder` parameter is used in order to warn the user, and there is no need for any ellipsis magic

zigaLuksic: i used ellipsis instead of `None` because `path` will be required after the renaming is finalized. No real magic in it, just a different default since `path=None` and `path=...` invoke different things in (my) mind.

We could make path part of the execution arguments, but that means that things fail much later :/

mlubej (approved):

PR #769, opened 2023-11-06
Enable github to trigger internal gitlab docker image build

mlubej:
Some context:

- removed gitlab mirror + CI job from the old actions file
- created a new actions file which is triggered in all old cases + on releases
  - there are multiple release types, more info [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release)
- `$GITHUB_REF` env variable is sent to gitlab as well, which starts with `refs/tags/` for releases. this info is used so that docker build is not triggered on push/PR/scheduled events

We can test this by creating a draft release, and then we can also delete it if need be. When everything is tested, only the published release type should trigger the docker build in gitlab

### DIFF
--- a/.github/workflows/ci_action.yml
+++ b/.github/workflows/ci_action.yml
@@ -111,19 +111,3 @@ jobs:
           files: coverage.xml
           fail_ci_if_error: true
           verbose: false
-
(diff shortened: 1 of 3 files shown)


zigaLuksic (approved):
looks good :)

PR #770, opened 2023-11-06
Fix bug in gitlab CI file

mlubej:


mlubej:
Pipeline now works, but we have a different issue, since the mirror+CI action doesn't support passing over the environment variables in the API call.  I opened an issue: https://github.com/SvanBoxel/gitlab-mirror-and-ci-action/issues/26

but until something is done, this won't work. We could do our own implementation, but probably it would be faster if I open a PR to the action repo

zigaLuksic (approved):

PR #801, opened 2025-02-25
Fix failing tests

mlubej:
Changes:
- increased min `numba` version because it fails to install
- limited `zarr` version after v3 release
- fixed `pylint` issues by ignoring
- fixed `mypy` issues

### DIFF
--- a/tests/geometry/test_transformations.py
+++ b/tests/geometry/test_transformations.py
@@ -239,7 +239,7 @@ VECTOR_TO_RASTER_TEST_CASES = (
             no_data_value=0,
             raster_dtype=bool,
         ),
-        img_exp_statistics={"exp_shape": (100, 150, 1), "exp_dtype": bool, "exp_min": False, "exp_max": True},
+        img_exp_statistics={"exp_shape": (100, 150, 1), "exp_dtype": bool, "exp_min": np.False_, "exp_max": np.True_},
     ),
 )
 
--- a/eolearn/core/core_tasks.py
+++ b/eolearn/core/core_tasks.py
@@ -13,7 +13,7 @@ import copy
 import datetime as dt
 import warnings
 from abc import ABCMeta
-from typing import Any, Callable, Iterable, Literal, cast
+from typing import Any, Callable, Iterable, Literal, Tuple, cast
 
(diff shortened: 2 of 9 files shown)


tests/geometry/test_transformations.py
```diff
@@ -239,7 +239,7 @@ class VectorToRasterTestCase:
             no_data_value=0,
             raster_dtype=bool,
         ),
-        img_exp_statistics={"exp_shape": (100, 150, 1), "exp_dtype": bool, "exp_min": False, "exp_max": True},
+        img_exp_statistics={"exp_shape": (100, 150, 1), "exp_dtype": bool, "exp_min": np.False_, "exp_max": np.True_},
```
mlubej, line 242: I hate this change, but the alternative is doing some casting in sh-py ... Or we limit it to numpy < 2? but this would negatively affect field boundaries..

zigaLuksic: I don't think we want to add such a harsh restriction. The reintroduced numpy booleans are causing some headaches...

zigaLuksic (approved):

tests/geometry/test_transformations.py
mlubej: Do you suggest to update `sh-py`, then? It would be cleaner, but we would need to release sh-py afterwards

zigaLuksic: what would you change there?

mlubej: The issue is that this tests uses the `assert_statistics_match` function, which lives in `sentinelhub/testing_utils.py`, so when the statistics are computed, they have the `numpy` dtypes

zigaLuksic: but is that not correct? does the image not use numpy dtypes?

mlubej: Discussed in private to leave it as it is, the main issue in fact stems from the `pytest.approx` comparison, where the comparison doesn't yet work for the new numpy boolean types, meaning we need to have the exact types. 

image

image

Perhaps this will be updated in the future. If that happens, we can simply revert the change.

